PR #164, opened 2026-07-25
feat: sim Route53 real DNS queryable with dig

### LINKED ISSUE #160: Inspect simulated Route53 records with dig and in the browser when serving locally
## Problem

Simulated Route53 records are only reachable through SDK commands. There's no way to look at what a simulated hosted zone actually contains, either from a terminal or a browser, which makes local experimentation and debugging harder than it needs to be — you have to write code to inspect state that a real Route53 user would just `dig`.

Route53 is the one simulated service whose whole purpose is name resolution, yet names are the one thing you can't currently resolve with normal tooling.

## Current behaviour

- `src/service/route53/command/` supports `create-hosted-zone`, `get-hosted-zone`, `list-hosted-zones-by-name` and `change-resource-record-sets`. There is no `ListResourceRecordSets`.
- `SimRoute53HostedZoneRecords` exposes `create()`, `upsert()`, `delete()`, `get(name, type)` and `count`. **There is no way to enumerate the records in a hosted zone.**
- `SimRoute53Resolver` resolves an HTTP hostname to a `SimAwsServiceTarget` by chasing CNAMEs (bounded by a visited set and a max depth). It is HTTP-routing shaped, not DNS shaped — it answers "which sim service handles this host", not "what are this name's records".
- `SimAwsHttp.handleRequest` routes purely by hostname through `route53().resolveHttpHost()`, returning 501 for anything unrecognised.
- `serveSimAws()` binds TCP only, on `sim-aws.localhost`, defaulting to port `0` (ephemeral).

## Desired behaviour

A served sim AWS environment answers DNS queries for its simulated hosted zones on UDP, and serves a human-readable summary of those zones over HTTP — both on `dns.sim-aws.localhost` at the same port number.

```bash
dig @dns.sim-aws.localhost -p 12345 www.example.test
```

...works at the same time as:

```bash
curl http://my-site.s3-website.eu-west-2.sim-aws.localhost:12345/
```

DNS is UDP and needs its own socket — it can't share the HTTP listener. But UDP and TCP port namespaces are separate, so the DNS server can bind UDP on the same port number the HTTP server took on TCP. That gives users one port number to remember and one hostname to type, with two protocols behind it:

- **UDP** → DNS answers, for `dig` and any other DNS client
- **TCP/HTTP** → a browser-viewable summary of every hosted zone and record

The payoff is that `dig` and `curl` tell the same story: deploy a CloudFormation template, `dig` shows the record that was created, and `curl` on that same name reaches the simulated S3 website behind it. When a queried name resolves to a sim service target, the A answer should be the loopback address the HTTP server is listening on, so the two views stay consistent.

## Proposed scope

Four phases, each independently useful and mergeable:

1. **`ListResourceRecordSets` + record enumeration.** Add enumeration to `SimRoute53HostedZoneRecords` behind a real Route53 API that's currently missing. Useful on its own regardless of the rest of this issue, and a hard prerequisite for both views.
2. **HTTP zone summary at `dns.sim-aws.localhost`.** Cheap, lands early, no new protocol work.
3. **DNS message codec.** Pure encode/decode, no sockets — exhaustively testable as `.iso.test.ts`.
4. **UDP socket + DNS answer semantics.** Bind on the shared port number, wire the codec to a new DNS answer resolver.

## Acceptance criteria

- [ ] `ListResourceRecordSets` returns a hosted zone's records in an AWS-shaped output, and is accepted from a real `@aws-sdk/client-route-53` command instance in tests.
- [ ] `http://dns.sim-aws.localhost:<port>/` renders the hosted zones and records of the served `SimAws` instance.
- [ ] `dig @dns.sim-aws.localhost -p <port> <name>` returns an authoritative answer for a name in a simulated hosted zone.
- [ ] Querying a name that resolves to a sim service target returns an A record for the loopback address the HTTP server is bound to, and `curl` against that same name reaches the same simulated service.
- [ ] A name in an existing zone with no record of the queried type returns NOERROR with no answers (NODATA); a name in no zone returns NXDOMAIN.
- [ ] The DNS port is exposed on the server object and does not require the caller to assume it matches the HTTP port.
- [ ] No new runtime dependency is added to `package.json`.

- **Port collision.** With `port: 0`, the HTTP server binds TCP first and gets an ephemeral port; the DNS server then tries the same number on UDP. That number could already be taken on UDP by an unrelated process — not unlikely with parallel vitest workers. Fall back to an independently chosen UDP port and expose `dnsPort` explicitly rather than letting callers assume the numbers always match.
- **Ports below 1024 need root**, so no attempt should be made to bind port 53.
- **Hostname resolution.** `dig @dns.sim-aws.localhost` depends on `*.sim-aws.localhost` resolving to loopback. The existing `sim-aws-local-server.loc.test.ts` already relies on this for real `fetch` calls against `my-site.s3-website.eu-west-2.sim-aws.localhost`, so it holds in dev and CI today. The hostname is cosmetic — the socket binds to loopback either way — but it reads better than `@127.0.0.1`.
- **`SimAwsServiceTarget.service` is currently the union `"s3" | "cloudFront"`.** The HTTP summary view needs either a widened union or handling ahead of controller dispatch in `SimAwsHttp`. Worth deciding deliberately rather than by accident.
- **Docs need to keep two roles distinct**, because this will otherwise confuse people: `dns.sim-aws.localhost` is *where the server is*, not *what it answers for*. It answers for whatever hosted zones exist — `example.test`, or whatever a template created.

## Out of scope

- Binding port 53, or anything requiring elevated privileges.
- Configuring the host OS resolver. macOS `/etc/resolver/<tld>` pointing at `127.0.0.1#<port>` makes simulated names resolve system-wide, but that's a user-side documentation footnote, not something Yulin should do.
- DNS-over-HTTPS/TLS.
- Recursion, forwarding, or answering for anything other than simulated hosted zones.
- Routing policies, weighted records, health checks, or DNSSEC — none of which sim Route53 models today.

## Implementation notes

**No new dependency.** `node:dgram` is built in. Yulin's runtime dependencies are currently just `@faker-js/faker`, `@kensio/part-factory` and `lodash-es`; a DNS package would be a disproportionate addition for what's needed here.

**The wire format work is bounded**, because this only ever acts as an authoritative server for the record types `SimRoute53RecordType` already models (`A`, `AAAA`, `CNAME`, `TXT`, `NS`, `SOA`):

- Question names are not compressed in practice, so parsing needs no compression-pointer handling.
- Name compression is optional when encoding — skipping it entirely removes the hardest part of DNS, and resolvers are required to accept uncompressed names.
- EDNS0 OPT records can be ignored with `ARCOUNT=0`, which is just pre-EDNS server behaviour. Worth verifying `dig` output is clean in phase 3 rather than assuming.
- UDP only to start. TCP DNS is the same codec behind a 2-byte length prefix over `node:net` if it's ever wanted.

**DNS answer semantics belong in a new class, not in `SimRoute53Resolver`.** The existing resolver chases CNAMEs to a `SimAwsServiceTarget` for HTTP routing; a DNS responder needs exact name+type matching, CNAME chasing within authoritative data, NXDOMAIN vs NODATA, and an SOA in the authority section for negative answers. Different question, different class. `SimRoute53Registry` is the right source of zones for both, since it already aggregates hosted zones across account scopes for exactly this kind of global name resolution.

**Decided:** the DNS server starts as part of `serveSimAws()` rather than being opted into separately. Serving a simulated AWS environment brings DNS up alongside HTTP, so there is nothing extra to discover and the demo is a single call. The cost is a UDP socket bound for every caller of `serveSimAws()`, including those who only want HTTP.

hughgrigg:
Resolves https://github.com/KensioSoftware/yulin/issues/160

## Summary by CodeRabbit

* **New Features**
  * Added UDP-based DNS support for simulated Route53 records.
  * Simulated servers now provide DNS alongside HTTP, including CNAME, alias, A, TXT, and negative responses.
  * Added automatic DNS port selection when the preferred port is unavailable.
  * Added Node.js and `dig` examples for querying simulated records.

* **Documentation**
  * Expanded Route53 DNS documentation with querying instructions, response behavior, and limitations.

### DIFF
--- a/docs/services/route53/README.md
+++ b/docs/services/route53/README.md
@@ -22,6 +22,7 @@ Sim Route53 currently supports:
 - Alias records, with `AliasTarget.DNSName` stored as the record value
 - Local hostname resolution through `*.sim-aws.localhost`
 - A browser-viewable hosted zone and record summary at `dns.sim-aws.localhost`
+- Real DNS answers over UDP, so records can be queried with `dig` or any DNS client
 - CloudFormation resources:
   - `AWS::Route53::HostedZone`
   - `AWS::Route53::RecordSet`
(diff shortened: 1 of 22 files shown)


hughgrigg:
@coderabbitai full review

PR #183, opened 2026-07-26
feat: simulated IAM SigV4 auth

### LINKED ISSUE #177: Verify SigV4 request signatures in simulated IAM
Depends on #176 for the clock that credential expiry is checked against.

## Problem

Simulated IAM can already decide whether a principal may perform an action on a resource, but nothing can work out which principal sent an HTTP request. Every endpoint served on localhost is therefore effectively unauthenticated. SigV4 is the missing authentication step: it is how a real AWS request carries its identity, and without it the authorization machinery that already exists cannot be reached from outside the process.

## Desired behaviour

Simulated IAM verifies an `AWS4-HMAC-SHA256` `Authorization` header against a registered access key and resolves it to the same `SimIamCredentialIdentity` that `resolveCredentials` already returns for in-process callers. A request signed with credentials from `CreateAccessKey` or STS `AssumeRole` verifies; a tampered request does not.

Verification is largely clock independent, because the signing date is an input to the signing key and therefore comes from the request itself. Two checks do need a clock, and they are deliberately treated differently:

- Signature age is not enforced. `X-Amz-Date` must be present, well formed, and consistent with the credential scope date, but is never compared to a clock. Nothing in a user's own system depends on SigV4 clock skew, and enforcing it would break any client that stamps real time and cannot know simulated time. This is a documented divergence from real AWS.
- Credential expiry is enforced against simulated time, because an assumed role session expiring part way through a test is a meaningful behaviour to assert.

## Example

Tests sign with the real AWS signer and assert the simulator accepts the result, giving conformance against the canonical implementation rather than hand written canonical strings:

```typescript
const signed = await new SignatureV4({
  service: "lambda",
  region: "us-east-1",
  credentials,
  sha256: Sha256,
}).sign(request);

expect(simAws.iam().verifySignedRequest(signed)).toEqual({
  principal: { kind: "arn", arn: userArn },
});
```

## Proposed scope

Authorization header parsing, credential scope, canonical request, string to sign, signing key derivation, signature comparison, and AWS shaped errors (`SignatureDoesNotMatch`, `IncompleteSignature`, `InvalidClientTokenId`, `ExpiredToken`). `node:crypto` only, no new runtime dependency and no `@aws-sdk/*` import in `src/`.

Access keys are registered per account on `SimIam`, but a signed request carries only an access key ID. That needs a global access key ID to account index on `SimAws`, the same pattern as `SimLambdaUrlRegistry.accountIdForFunctionUrl`. The secret access key should not leave IAM: the registry derives a signing key rather than handing the secret out.

## Acceptance criteria

- [ ] a request signed by `@smithy/signature-v4` (added as a devDependency) verifies and resolves to the signing principal
- [ ] altering the method, path, query, a signed header or the body fails verification
- [ ] an unknown or inactive access key ID gives `InvalidClientTokenId`
- [ ] temporary credentials verify via `X-Amz-Security-Token`, and an expired session is rejected against simulated time
- [ ] a signature stamped hours away from the current time still verifies
- [ ] canonicalization covers path double encoding, query ordering by encoded key then value, header trimming and whitespace collapsing, multi value headers, and `UNSIGNED-PAYLOAD`

## Out of scope

Presigned query string auth (`X-Amz-Algorithm` in the query), S3 `aws-chunked` streaming signatures, and SigV4A. Query string auth should be a sibling entry point in the design rather than something needing a rewrite later.

hughgrigg:
Resolves https://github.com/KensioSoftware/yulin/issues/177 

## Summary by CodeRabbit

* **New Features**
  * Added AWS Signature Version 4 request verification.
  * Requests can now be validated across accounts using IAM access keys and temporary session credentials.
  * Added support for canonicalizing request paths, queries, headers, and payloads.
  * Added AWS-compatible errors for invalid, expired, incomplete, or mismatched signatures.
* **Bug Fixes**
  * Improved validation of inactive, unknown, expired, and incorrectly presented credentials.
* **Tests**
  * Added comprehensive coverage for valid signatures, malformed requests, credential failures, and request tampering.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
   "license": "Apache-2.0",
   "packageManager": "pnpm@11.17.0+sha512.cca3cea332ad254bb84145f966d19f4879615210346fc92c79a047f23a0d7b3cca3c3792f0076ba1f1831d277efbcf0a9119b31a9a60eca7fb3d6231f331ef72",
   "devDependencies": {
+    "@aws-crypto/sha256-js": "^5.2.0",
     "@aws-sdk/client-acm": "^3.1086.0",
     "@aws-sdk/client-cloudformation": "^3.1086.0",
     "@aws-sdk/client-cloudfront": "^3.1086.0",
@@ -108,6 +109,7 @@
     "@aws-sdk/client-sts": "^3.1086.0",
     "@eslint/js": "^10.0.1",
     "@kensio/smartass": "^1.34.0",
+    "@smithy/signature-v4": "^5.6.10",
     "@smithy/smithy-client": "^4.14.8",
     "@smithy/types": "^4.16.1",
     "@types/eslint-plugin-security": "^3.0.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18,6 +18,9 @@ importers:
         specifier: ^4.18.1
         version: 4.18.1
     devDependencies:
+      '@aws-crypto/sha256-js':
+        specifier: ^5.2.0
+        version: 5.2.0
       '@aws-sdk/client-acm':
         specifier: ^3.1086.0
         version: 3.1093.0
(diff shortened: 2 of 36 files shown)


hughgrigg:
@coderabbitai full review

PR #211, opened 2026-07-28
feat: create simulated KMS keys from AWS::KMS::Key and AWS::KMS::Alias

### LINKED ISSUE #210: Create simulated KMS keys from AWS::KMS::Key and AWS::KMS::Alias in CloudFormation and CDK templates
## Problem

Simulated KMS keys can only be created through the SDK (`CreateKey`, `CreateAlias`). Deploying a CloudFormation or CDK template that declares `AWS::KMS::Key` or `AWS::KMS::Alias` fails, so a test that wants an encrypted resource has to create the key out of band and hand the ID to the template, which is not how the real stack is written.

## Current behaviour

`src/service/kms/` has no `cfn/` directory, and `sim-cfn-service-resolver.ts` has no `KMS` case, so any `AWS::KMS::*` resource throws `Unsupported sim CloudFormation Resource service KMS` during deployment. This is already recorded as a limitation in `docs/services/kms/README.md`.

## Desired behaviour

`AWS::KMS::Key` creates a simulated key in the stack's account and region, and `AWS::KMS::Alias` points an alias at it. Value adapters give the AWS behaviour: `Ref` on a key returns its key ID, `Fn::GetAtt` returns `Arn` and `KeyId`, and `Ref` on an alias returns the alias name (`alias/app-key`).

## Example

```yaml
AppKey:
  Type: AWS::KMS::Key
  Properties:
    Description: Application data key
    KeyPolicy: { ... }
AppKeyAlias:
  Type: AWS::KMS::Alias
  Properties:
    AliasName: alias/app-key
    TargetKeyId: !Ref AppKey
```

After `waitForDeployComplete()`, `DescribeKey` with `alias/app-key` resolves the key, and `Encrypt`/`Decrypt` work against it.

## Acceptance criteria

- [ ] A template declaring both resource types deploys, and the key and alias are visible through `DescribeKey`, `ListKeys` and `ListAliases`.
- [ ] `KeyPolicy` is applied as the key policy, and omitting it gets the default root-delegation policy, matching `CreateKey`.
- [ ] `Ref` and `Fn::GetAtt` return the values above.
- [ ] Properties the simulator does not model refuse deployment with a clear error rather than being ignored: `EnableKeyRotation`, `RotationPeriodInDays`, `MultiRegion`, non-`AWS_KMS` `Origin`, and asymmetric or HMAC `KeySpec`/`KeyUsage`.
- [ ] `Tags` is either simulated or refused, not silently dropped.

## Edge cases

`Enabled: false` needs a disabled key at creation, which `CreateKey` does not currently accept as input. `PendingWindowInDays` only matters on delete, and simulated CloudFormation has no stack deletion yet.

## Out of scope

Stack updates that retarget an alias, since `UpdateAlias` and `DeleteAlias` are not simulated. Making other services actually encrypt with a referenced key: Secrets Manager stores `KmsKeyId` without using it, and that stays true here.

hughgrigg:
Simulated CloudFormation creates a key from AWS::KMS::Key and points an AWS::KMS::Alias at it, through the ordinary CreateKey and CreateAlias commands. Ref on a key gives its key ID, Fn::GetAtt gives Arn and KeyId, and Ref on an alias gives the alias name.

Properties that ask for behaviour the simulator does not model refuse the deployment rather than being dropped: EnableKeyRotation, RotationPeriodInDays, MultiRegion, Tags, and asymmetric or HMAC key types.

Resolves https://github.com/KensioSoftware/yulin/issues/210

## Summary by CodeRabbit

* **New Features**
  * Added simulated CloudFormation support for KMS keys and aliases.
  * KMS keys can be created, configured, enabled or disabled, and referenced by ID or ARN.
  * KMS aliases can target deployed keys and be used for encryption and decryption.
  * Added CloudFormation integration for additional simulated services, including ACM, CloudFront, IAM, Lambda, Route 53, S3, and Secrets Manager.

* **Documentation**
  * Expanded CloudFormation and KMS documentation with usage examples, supported behavior, intrinsic references, and limitations.

### DIFF
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -20,6 +20,7 @@ Sim CloudFormation currently supports:
 - Deploying parsed template objects with `deployTemplate(...)`
 - Deploying synthesized JSON template files with `deployTemplateFile(...)`
 - Template `Parameters` with supplied values and defaults
+- Template `Outputs`, resolved after resource creation and read from `stack.outputs`
 - Common intrinsic functions:
   - `Ref`
   - `Fn::GetAtt`
@@ -987,6 +988,7 @@ Current supported resource areas include:
 Common template features currently supported include:
 
 - `Parameters`
+- `Outputs`, including `Description` and `Export`
 - `Ref`
 - `Fn::GetAtt`
 - `Fn::Join`
@@ -1000,7 +1002,7 @@ Notable limitations:
 - Only supported resource types create simulated service resources.
 - Unsupported resource properties may be ignored or rejected depending on the resource simulator.
 - Stack updates and deletes are not currently documented as supported operations.
-- Outputs, mappings, conditions, and many advanced CloudFormation features are not currently
+- Mappings, conditions, and many advanced CloudFormation features are not currently
   documented as supported operations.
 
 For best results, keep test templates focused on the resources and behaviours your application
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -33,6 +33,7 @@ Sim CloudFormation currently supports:
   - `AWS::CloudFront::Distribution`
   - `AWS::Lambda::Function`
   - `AWS::SecretsManager::Secret`
+  - `AWS::KMS::Key` and `AWS::KMS::Alias`
   - CloudFront Functions used by CDK-generated templates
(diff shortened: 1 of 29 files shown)


hughgrigg:
@coderabbitai review

PR #216, opened 2026-07-28
feat: read configuration from a simulated SSM Parameter Store

### LINKED ISSUE #212: Read configuration from a simulated SSM Parameter Store in tests and Lambda handlers
## Problem

A Lambda handler that reads its configuration from SSM Parameter Store on cold start cannot be exercised against Yulin. The usual shape is a `GetParameterCommand` for something like `/myapp/prod/db-host`, and today the only option is mocking the SDK client, which proves nothing about whether the parameter exists, whether the execution role is allowed to read it, or whether the IAM policy's resource ARN is actually right.

## Current behaviour

There is no `src/service/ssm/`. Intercepting an `SSMClient` throws `SimSdkUnknownServiceError` from `resolve-sim-sdk-command-router.ts`.

## Desired behaviour

A `SimSsm` facade reachable as `simAws.ssm()`, scoped to an account and region, handling AWS SDK commands and authorizing each one through simulated IAM.

Handler code needs no special treatment: `sim-sdk-lambda-vm-module-provider.ts` already intercepts any `@aws-sdk/*` package required by function code and dispatches with the execution role as the caller, so a router entry is enough to make it work inside a simulated Lambda.

## Example

```typescript
await simAws.ssm().putParameter(
  new PutParameterCommand({
    Name: "/myapp/prod/db-host",
    Type: "String",
    Value: "db.internal",
  }),
);

// Handler code reading it, authorized as the function's execution role.
const { Parameter } = await new SSMClient({}).send(
  new GetParameterCommand({ Name: "/myapp/prod/db-host" }),
);
```

## Proposed scope

`PutParameter`, `GetParameter`, `GetParameters`, `GetParametersByPath`, `DeleteParameter`, `DeleteParameters`, `DescribeParameters`. `String` and `StringList` types only.

## Acceptance criteria

- [ ] Parameter ARNs drop the name's leading slash (`/myapp/prod/db-host` becomes `arn:aws:ssm:eu-west-2:111111111111:parameter/myapp/prod/db-host`), so a policy written against the wrong form fails here rather than in a deployment
- [ ] `PutParameter` starts at `Version` 1 and increments on `Overwrite: true`; without `Overwrite` an existing name fails with `ParameterAlreadyExists`
- [ ] `GetParameter` resolves `name:version` and `name:label` selectors, and fails with `ParameterNotFound` for a missing name
- [ ] `GetParameters` returns found parameters in `Parameters` and unknown names in `InvalidParameters` rather than throwing, which is what makes a typo silently return nothing on real AWS
- [ ] `StringList` values come back as one comma separated `Value` string, not an array
- [ ] `GetParametersByPath` honours `Recursive`, and `DescribeParameters` authorizes against `*` because AWS gives it no resource-level permissions
- [ ] `Type: SecureString` is refused for now, since nothing would encrypt it
- [ ] `SSMClient` is intercepted by `SimSdk`, including from inside a simulated Lambda handler
- [ ] A `docs/services/ssm/` page, with limitations stated

Name validation is worth getting right because it is a common deploy-time failure: a hierarchical name must start with `/`, cannot exceed 15 levels, and cannot begin with `aws` or `ssm`, which are reserved.

The standard tier's 4KB value limit is the other one people hit, usually by putting a whole JSON config blob in one parameter.

## Out of scope

`AWS::SSM::Parameter` and `{{resolve:ssm:...}}`, which have their own issues. `SecureString` and KMS encryption, which has its own issue. Parameter policies and the advanced tier, tags, `LabelParameterVersion`, `GetParameterHistory`, and the Parameters and Secrets Lambda Extension HTTP endpoint. Everything in SSM that is not Parameter Store: Run Command, Session Manager, Patch Manager, State Manager, Automation, inventory and maintenance windows.

## Implementation notes

`src/service/secretsmanager/` is the closest existing template, and the two services solve the same problem in similar shapes: a store owning name resolution, a version model, a shared writer for every write path, and an authorizer that treats the list-style operation differently. Registration is in `resolve-sim-sdk-command-router.ts` and the `SimAws` service factory and account/region scope.

hughgrigg:
Adds a SimSsm facade for Parameter Store, scoped by account and region, handling PutParameter, GetParameter, GetParameters, GetParametersByPath, DeleteParameter, DeleteParameters and DescribeParameters through simulated IAM. Parameter ARNs drop the name's leading slash, so a policy written against the wrong form fails here rather than in a deployment.

SecureString is refused, since nothing would encrypt it, and parameter labels are refused with an explanation rather than looked up, since LabelParameterVersion is not implemented.

Resolves https://github.com/KensioSoftware/yulin/issues/212

## Summary by CodeRabbit

* **New Features**
  * Added simulated AWS Systems Manager (SSM) Parameter Store support, including parameter CRUD, versioning, and describe/list-by-path operations.
  * Supports `String` and `StringList` parameters with account/region scoping and IAM-based authorization, including Lambda-role access.
  * Added SDK command routing and a public `./ssm` package export.

* **Documentation**
  * Added SSM service docs, including the new simulator behavior/limitations overview and multiple usage examples.
  * Updated service documentation indexes to include SSM. 

* **Tests**
  * Added extensive SSM simulator tests covering validation, paging, batching, and authorization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ npm i -D @kensio/yulin
 - [Route53](./docs/services/route53 "Simulated Route53 docs")
 - [S3](./docs/services/s3 "Simulated S3 docs")
 - [Secrets Manager](./docs/services/secretsmanager "Simulated Secrets Manager docs")
+- [SSM Parameter Store](./docs/services/ssm "Simulated SSM Parameter Store docs")
 - [STS](./docs/services/sts "Simulated STS docs")
 
 ## Feature specific docs
--- a/docs/README.md
+++ b/docs/README.md
@@ -14,6 +14,7 @@ behaviour and includes example code that can be copied into tests or local devel
 - [Route53](./services/route53/ "Simulated Route53 usage docs")
 - [S3](./services/s3/ "Simulated S3 usage docs")
 - [Secrets Manager](./services/secretsmanager/ "Simulated Secrets Manager usage docs")
+- [SSM Parameter Store](./services/ssm/ "Simulated SSM Parameter Store usage docs")
 - [STS](./services/sts/ "Simulated STS usage docs")
 
 ## Feature documentation
--- a/docs/services/ssm/README.md
+++ b/docs/services/ssm/README.md
@@ -276,7 +276,7 @@ request.
 
 ## String lists
 
-A `StringList` value is one comma separated string, on read as well as on write. It is not returned
+A `StringList` value is one comma-separated string, on read as well as on write. It is not returned
 as an array.
 
 ```typescript sim-ssm-string-list
--- a/docs/services/ssm/README.md
+++ b/docs/services/ssm/README.md
@@ -1,0 +1,504 @@
+# Simulated SSM Parameter Store
+
+Yulin includes a simulated AWS Systems Manager Parameter Store for tests and local development.
+Parameters are stored in memory, versioned on every write, and every operation is authorized by
+simulated IAM.
+
+Only Parameter Store is simulated. Nothing else in Systems Manager is.
+
+SSM-specific types are imported from the `@kensio/yulin/ssm` subpath.
+
+## Available functionality
+
+Sim SSM currently supports:
+
+- `PutParameterCommand`, creating a parameter or overwriting one
+- `GetParameterCommand`, by name, by `name:version` or by `name:label`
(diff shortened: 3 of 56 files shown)


hughgrigg:
@coderabbitai review

PR #235, opened 2026-07-29
feat: sign real JWTs for simulated Cognito users

### LINKED ISSUE #220: Get real signed JWTs for simulated Cognito users, verifiable against a real JWKS
## Problem

Handler code that verifies a Cognito access or id token, usually with `aws-jwt-verify` or `jose`, cannot be exercised against Yulin. The workaround is signing a token with a hand-rolled key and stubbing the verifier, which says nothing about whether the issuer is right, whether the claim the code reads exists on that token type, or whether the token has expired.

## Desired behaviour

`AdminInitiateAuth` returns an `AuthenticationResult` whose `IdToken` and `AccessToken` are real RS256 JWTs, signed by a key the user pool also publishes as a JWKS. A verifier configured for the pool verifies them unchanged, with no network involved.

Depends on https://github.com/KensioSoftware/yulin/issues/218 and https://github.com/KensioSoftware/yulin/issues/219.

## Example

```typescript
const { AuthenticationResult } = await simAws.cognitoIdentityProvider().adminInitiateAuth(
  new AdminInitiateAuthCommand({
    UserPoolId: userPoolId,
    ClientId: clientId,
    AuthFlow: "ADMIN_USER_PASSWORD_AUTH",
    AuthParameters: { USERNAME: "alice", PASSWORD: "Sup3rSecret!" },
  }),
);

const verifier = CognitoJwtVerifier.create({ userPoolId, tokenUse: "access", clientId });
verifier.cacheJwks(simAws.cognitoIdentityProvider().userPool(userPoolId).jwks());

const payload = await verifier.verify(AuthenticationResult.AccessToken);
```

## Proposed scope

`AdminInitiateAuth` and `AdminRespondToAuthChallenge` for the `ADMIN_USER_PASSWORD_AUTH` flow and the `NEW_PASSWORD_REQUIRED` challenge, token signing, and JWKS exposure through the user pool.

## Acceptance criteria

- [ ] `ADMIN_USER_PASSWORD_AUTH` fails with `InvalidParameterException` when the app client's `ExplicitAuthFlows` does not include `ALLOW_ADMIN_USER_PASSWORD_AUTH`
- [ ] A user in `FORCE_CHANGE_PASSWORD` gets `ChallengeName: "NEW_PASSWORD_REQUIRED"` and a `Session` rather than tokens, and `AdminRespondToAuthChallenge` completes it and returns tokens
- [ ] A wrong password, and a disabled user, fail with `NotAuthorizedException`
- [ ] The claim split matches real Cognito: the id token carries `aud`, `token_use: "id"` and `cognito:username`, the access token carries `client_id` and no `aud`, `token_use: "access"`, `username` and `scope: "aws.cognito.signin.user.admin"`, and both carry `cognito:groups` and the same `sub`
- [ ] `iss` is `https://cognito-idp.eu-west-2.amazonaws.com/<userPoolId>`, and the JWT header carries `alg: "RS256"` and a `kid` present in the pool's JWKS
- [ ] `iat`, `exp` and `auth_time` come from the simulated clock and default to 60 minutes validity, so advancing simulated time expires a token
- [ ] An app client created with `GenerateSecret: true` requires a correct `SECRET_HASH` in `AuthParameters`
- [ ] The signing key is generated with `node:crypto` on first use and memoized for the process, so no private key material is committed to the repo
- [ ] Docs cover verifying a simulated token with `aws-jwt-verify`

Real Cognito publishes two keys per pool and rotates between them. Publishing one is a divergence worth documenting, since code that assumes a single JWKS entry is still wrong against real AWS.

The `RefreshToken` returned here should be an opaque string as on real Cognito, not a JWT, even though no flow consumes it yet.

## Out of scope

`InitiateAuth` and the client-side flows, which have their own issue. Serving the JWKS over HTTP, which has its own issue. SRP, MFA challenges, custom auth challenges, device tracking, custom scopes from resource servers, and pre-token-generation Lambda triggers.

## Implementation notes

`src/service/kms/key/sim-kms-key-material.ts` is the nearest precedent for generating real key material with `node:crypto`. Token timestamps must come from the injectable clock rather than `new Date()`, so `simAws.clock().advance(...)` expires a token.

Committing a private key would set off secret scanners for no gain, and caching one on disk adds cross-process state and cleanup that has not been shown to be needed. Generating per process is the starting point; a disk cache is worth revisiting only if 2048-bit RSA generation turns out to be slow enough to matter.

If the auth flow and the token model together look too big at implementation time, the signing key and JWKS split off cleanly, since they are testable through `userPool.jwks()` before any auth command exists.

hughgrigg:
AdminInitiateAuth runs the ADMIN_USER_PASSWORD_AUTH flow and answers with RS256 tokens signed by a key the pool publishes as a JWKS, so aws-jwt-verify verifies them unchanged with nothing on the network. A user in FORCE_CHANGE_PASSWORD gets the NEW_PASSWORD_REQUIRED challenge, and AdminRespondToAuthChallenge completes it.

Token timestamps come from the simulated clock, so signing in in the past produces a token a verifier rejects as expired. A user's password is now kept rather than discarded, since sign-in reads it.

Resolves https://github.com/KensioSoftware/yulin/issues/220

## Summary by CodeRabbit

* **New Features**
  * Added simulated Cognito admin sign-in with `ADMIN_USER_PASSWORD_AUTH`.
  * Added signed RS256 access and ID tokens with locally available JWKS verification.
  * Added `NEW_PASSWORD_REQUIRED` challenge handling and password completion.
  * Added password validation, user status transitions, token claims, expiry, and session handling.
  * Added IAM authorization coverage for authentication requests.

* **Documentation**
  * Expanded Cognito documentation with sign-in, token verification, challenge, expiry, and limitation guidance.
  * Added examples demonstrating sign-in, token verification, expired tokens, and password challenges.

* **Tests**
  * Added comprehensive coverage for authentication flows, JWT claims, verification, sessions, routing, and authorization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -308,6 +308,3 @@ dmypy.json
 tmp/
 *.tmp
 *.tmp.*
-
-# Video demo project (see .claude/video/HOUSE-STYLE.md)
-/yulin-demo/
--- a/docs/services/cognito/README.md
+++ b/docs/services/cognito/README.md
@@ -22,16 +22,21 @@ Sim Cognito currently supports:
 - `CreateGroupCommand`, `GetGroupCommand`, `UpdateGroupCommand`, `DeleteGroupCommand`,
   `ListGroupsCommand`, `AdminAddUserToGroupCommand`, `AdminRemoveUserFromGroupCommand`,
   `AdminListGroupsForUserCommand` and `ListUsersInGroupCommand`
+- `AdminInitiateAuthCommand` and `AdminRespondToAuthChallengeCommand`, for the
+  `ADMIN_USER_PASSWORD_AUTH` flow and the `NEW_PASSWORD_REQUIRED` challenge
+- Real RS256 JWTs, signed by a key the pool publishes as a JWKS, so a verifier configured for the
+  pool verifies them unchanged
 - Pool ids in the real `<region>_<nine characters>` form, and pool ARNs built from them
 - The real default password policy, applied to the passwords users are given
 - The real user status lifecycle, so an admin-created user stays in `FORCE_CHANGE_PASSWORD` until it
   has a permanent password
-- Group membership, and the precedence order the `cognito:groups` claim will use
+- Group membership, and the precedence order the `cognito:groups` claim uses
 - App client authentication flows, token lifetimes and generated client secrets
 - Authorization of every operation by simulated IAM, against the real IAM action and ARN
 - Calls made from inside a simulated Lambda handler, authorized as the function's execution role
 
-Nothing signs in yet. Tokens and the authentication flows themselves are a separate piece of work.
+The client-side flows are a separate piece of work: `InitiateAuth`, SRP and the hosted UI are not
+implemented.
 
 ## Creating a pool and an app client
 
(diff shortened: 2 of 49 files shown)


hughgrigg:
@coderabbitai review

PR #246, opened 2026-07-30
docs: move the feature lists off the top of the docs pages

hughgrigg:
Each docs page now starts with its intro and the first thing a reader wants to do. The "Available functionality" list is reference material for someone checking whether a thing is simulated, so it sits at the end of the page next to "Limitations".

The simulated time list is dropped rather than moved, since every bullet restated the heading below it. The CloudFormation page's top list and its "Supported resources and limitations" section are merged into one functionality list and one limitations list at the end.

Also list Cognito and SSM as supporting SDK interception, which they do.

## Summary by CodeRabbit

* **Documentation**
  * Expanded the SDK interception guide to include Cognito and SSM.
  * Reorganized service documentation for clearer separation of usage guidance, available functionality, and limitations.
  * Added standalone CloudFormation usage guidance, examples, supported capabilities, and clarified limitations.
  * Added Lambda guidance for SDK interception and package imports.
  * Removed the detailed simulated-time API reference from the time documentation.

### DIFF
--- a/docs/sdk/README.md
+++ b/docs/sdk/README.md
@@ -172,8 +172,8 @@ throw a diagnostic error.
 
 ## Supported services and Commands
 
-All simulated services support SDK interception: ACM, CloudFormation, CloudFront, DynamoDB, IAM, KMS,
-Lambda, Route53, S3, Secrets Manager and STS. Each service's own docs under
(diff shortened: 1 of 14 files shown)


hughgrigg:
@coderabbitai review

PR #249, opened 2026-07-30
feat: redrive failed messages to a simulated SQS dead-letter queue

### LINKED ISSUE #244: Redrive failed messages to a simulated SQS dead-letter queue
Depends on https://github.com/KensioSoftware/yulin/issues/242.

## Problem

That a consumer's poison message ends up on the dead-letter queue, rather than being redelivered forever, is one of the more valuable things to assert about a queue. It is also one of the easiest things to get wrong in a template, because a `RedrivePolicy` with the wrong `deadLetterTargetArn` looks fine until something actually fails. Neither can be tested.

## Desired behaviour

`RedrivePolicy` is accepted as a queue attribute on `CreateQueue` and `SetQueueAttributes`. Once a message has been received `maxReceiveCount` times without being deleted, the next time its visibility timeout lapses it moves to the queue named by `deadLetterTargetArn` instead of becoming receivable again. It arrives there with its body, message attributes and `MessageId` unchanged, so a test can identify it.

Advancing the simulation clock is what drives the move, in keeping with how visibility works in the core SQS issue.

## Example

```typescript
await simAws.sqs().setQueueAttributes(
  new SetQueueAttributesCommand({
    QueueUrl,
    Attributes: {
      RedrivePolicy: JSON.stringify({ deadLetterTargetArn: dlqArn, maxReceiveCount: 3 }),
    },
  }),
);
```

## Acceptance criteria

- [ ] `RedrivePolicy` is validated as JSON with `deadLetterTargetArn` and `maxReceiveCount`, and rejected otherwise
- [ ] `maxReceiveCount` outside 1 to 1000 is refused
- [ ] `deadLetterTargetArn` must name an existing queue in the same account and region, as real SQS requires; a target that does not exist fails at `SetQueueAttributes` rather than silently losing messages later
- [ ] A message received `maxReceiveCount` times without deletion moves to the dead-letter queue on the next visibility lapse, and is receivable there with the same `MessageId` and body
- [ ] A message deleted before reaching the count never reaches the dead-letter queue
- [ ] `ApproximateNumberOfMessages` on both queues reflects the move
- [ ] The `docs/services/sqs/` page documents redrive, including anything not simulated

Worth confirming against real SQS before implementing: whether `ApproximateReceiveCount` resets when a message lands on the dead-letter queue, and what `SentTimestamp` reports there. Both are visible to a handler reading the event, so a wrong guess is a divergence a consumer can see.

## Out of scope

`RedriveAllowPolicy` and the `StartMessageMoveTask` family of operations for draining a dead-letter queue back to its source. `ListDeadLetterSourceQueues`. The `RedrivePolicy` property on `AWS::SQS::Queue`, which can follow once the CloudFormation resource type has landed. FIFO dead-letter queues.

hughgrigg:
RedrivePolicy is accepted on CreateQueue and SetQueueAttributes. A message received maxReceiveCount times without being deleted moves to the queue named by deadLetterTargetArn on the next lapse of its visibility timeout, keeping its id, body, message attributes and sent timestamp.

The target has to be an existing queue in the same account and region, checked when the policy is set rather than when a message needs it.

Resolves https://github.com/KensioSoftware/yulin/issues/244

## Summary by CodeRabbit

* **New Features**
  * Added dead-letter queue support for simulated Amazon SQS standard queues.
  * Failed messages can be automatically moved after a configurable number of receive attempts.
  * Dead-letter deliveries preserve message content and metadata while reporting the source queue.
  * Added validation for redrive policies and dead-letter queue targets.

* **Documentation**
  * Added setup guidance and a complete dead-letter queue usage example.
  * Documented supported behavior, limitations, and unsupported policy features.

### DIFF
--- a/docs/services/sqs/README.md
+++ b/docs/services/sqs/README.md
@@ -185,6 +185,102 @@ A handle whose visibility timeout has lapsed with nobody else having received th
 still the most recent one, so deleting with it works. A handle the queue never issued fails with
 `ReceiptHandleIsInvalid`, and a repeated delete of a message already gone succeeds.
 
+## Dead-letter queues
+
+A `RedrivePolicy` says where a message goes once a consumer has had enough attempts at it. Once a
+message has been received `maxReceiveCount` times without being deleted, the next lapse of its
+visibility timeout moves it to the queue named by `deadLetterTargetArn` rather than making it
+receivable again. Advancing the clock is what drives the move, as it drives the timeout itself.
+
+```typescript sim-sqs-dead-letter-queue
+/**
+ * A message a consumer keeps failing on, ending up on the dead-letter queue.
+ */
+
+import {
+  CreateQueueCommand,
+  GetQueueAttributesCommand,
+  ReceiveMessageCommand,
+  SendMessageCommand,
+} from "@aws-sdk/client-sqs";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const sqs = simAws.sqs();
+
+// The dead-letter queue has to exist before anything can point at it.
+const { QueueUrl: DeadLetterQueueUrl } = await sqs.createQueue(
+  new CreateQueueCommand({ QueueName: "orders-dlq" }),
+);
+
+const deadLetter = await sqs.getQueueAttributes(
+  new GetQueueAttributesCommand({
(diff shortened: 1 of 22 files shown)


hughgrigg:
@coderabbitai review

PR #277, opened 2026-07-31
feat: write simulated DynamoDB items with the primary key rules PutIt…

### LINKED ISSUE #253: Store DynamoDB attribute values without losing number precision
## Problem

Simulated DynamoDB stores every number as a JavaScript `number`. DynamoDB numbers are decimals carrying up to 38 significant digits, and a double carries about 15. Anything held as `N` that is an identifier, a monetary amount or a large counter comes back changed, and a test that round trips it asserts a value the real service would never have returned.

The write path is not observable either. `SimDynamoDbTable.putItem` schedules the map write on the background scheduler and resolves before it happens, so PutItem is not read-your-writes, and the resulting `items` map is not read anywhere in the tree.

## Current behaviour

`DynamoDBItemAttribute.fromAttributeValue` parses `N` with `Number(value.N)` (`dynamodb-item-attribute.ts:39`) and serialises it back with `String(value)` (`:84`), so the round trip loses digits:

```typescript
const output = await simAws.dynamoDb().putItem(
  new PutItemCommand({
    TableName: "FoobarTable",
    Item: { id: { S: "a" }, counter: { N: "9007199254740993" } },
  }),
);
// output.Attributes.counter.N is "9007199254740992"
```

`NS` goes through `Number` the same way. Sets are built with `new Set(...)` over the raw values, so a binary set deduplicates by object identity rather than by bytes and keeps two `Uint8Array` values holding identical bytes. Serialisation also inherits JavaScript number formatting, including the switch to exponent notation above 1e21.

Separately, `put-item.handler.ts:81` returns `Attributes` unconditionally, containing the item just written. Real DynamoDB returns no `Attributes` unless `ReturnValues` is `ALL_OLD`, and then returns the previous item. `put-item.iso.test.ts` asserts the current behaviour throughout, so those assertions change with this work.

## Desired behaviour

A new `item/` domain model owns the attribute value type across all ten descriptors. `N` is held as a normalised decimal string: 38 significant digits, range 1E-130 to 9.9999999999999999999999999999999999999E+125 and its negative mirror, with the same normalisation applied inside `NS`. Whether that is a hand rolled decimal string or a stable existing decimal library is open, so long as the digits given come back exactly. `B` is raw bytes with base64 on the wire. Sets are homogeneous, never empty, and compare by value. Empty `S` and `B` values are allowed for non-key attributes and refused for key attributes. Nesting is capped at 32 levels, and the 400 KB item size rule counts attribute names as well as values.

The table gains an item store keyed by the marshalled primary key, written synchronously so PutItem is read-your-writes. PutItem requires the complete primary key with key value types matching the declared `AttributeDefinitions`, replaces the whole item, and accepts `ReturnValues` of `NONE` or `ALL_OLD` only.

If the attribute value model and the PutItem rework together run past a reviewable size, the model can land first and PutItem follow.

## Acceptance criteria

- [ ] A 38 digit `N` value round trips unchanged, and a value outside the documented range is a `ValidationException`
- [ ] An empty set and a mixed-type set are refused, and a binary set deduplicates by bytes rather than by reference
- [ ] An empty string is accepted as a non-key attribute and refused as a key attribute
- [ ] An item over 400 KB counting attribute names, or nested past 32 levels, is a `ValidationException`
- [ ] PutItem with no `ReturnValues` returns no `Attributes`, and `ALL_OLD` returns the previous item, or nothing when there was none
- [ ] The item is in the table store by the time the PutItem call resolves, with no wait on the background scheduler
- [ ] `ConditionExpression`, `ExpressionAttributeNames`, `ExpressionAttributeValues`, `ReturnConsumedCapacity`, `ReturnItemCollectionMetrics`, `Expected` and `ConditionalOperator` are each refused by name

## Out of scope

Reading items back, which is the next issue. Condition expressions. Consumed capacity and item collection metrics. Secondary index maintenance. The ordering comparators sort keys will need. DynamoDB Streams and PartiQL are not on this roadmap and belong under Limitations in the docs.

Builds on #251.

hughgrigg:
…em applies

PutItem checks the whole primary key against what the table declared, replaces the item under it, and answers with the item it replaced only when asked for ALL_OLD. The write lands before the call returns, so a table is read-your-writes. Condition and reporting inputs are refused by name.

Resolves https://github.com/KensioSoftware/yulin/issues/253

## Summary by CodeRabbit

* **New Features**
  * Added DynamoDB `PutItem` support with immediate replacement writes.
  * Added support for returning previous item attributes with `ALL_OLD`.
  * Added comprehensive handling for strings, numbers, binary values, sets, lists, maps, booleans, and nulls.
  * Preserved precision for large numeric values and normalized numeric representations.
* **Bug Fixes**
  * Improved validation for key attributes, nested values, duplicate set members, and item-size limits.
* **Documentation**
  * Expanded DynamoDB behavior, limitations, and usage examples.


hughgrigg:
@coderabbitai review

hughgrigg:
@coderabbitai review

hughgrigg:
@coderabbitai review

PR #279, opened 2026-07-31
feat: create simulated DynamoDB tables from AWS::DynamoDB::Table reso…

### LINKED ISSUE #255: Create simulated DynamoDB tables from AWS::DynamoDB::Table CloudFormation resources
## Problem

Simulated CloudFormation cannot deploy a DynamoDB table. A stack holding a table plus a Lambda that writes to it is an ordinary shape, and none of it can be exercised here, because the table resource never becomes a simulated table and the rest of the stack has nothing to point at. SQS, SSM, Secrets Manager and KMS all have a `cfn/` directory already. DynamoDB is the gap.

## Current behaviour

`resolveSimCloudFormationServiceResourceFactory` has no `DynamoDB` case, so an `AWS::DynamoDB::Table` resource reaches the default and throws `Unsupported sim CloudFormation Resource service DynamoDB`. That message is asserted at `sim-cfn-service-resolver.iso.test.ts:117`. There is no `cfn/` directory under `src/service/dynamodb/` and no DynamoDB entry in `simCfnResourceValueAdapter`, so a `Ref` to a table would fall through to `SimCfnDefaultResourceValueAdapter` and answer with the logical ID rather than the table name.

## Desired behaviour

`AWS::DynamoDB::Table` deploys a simulated table through the ordinary `CreateTable` path, so a table a template deployed is the same thing an SDK caller would have got, with the same validation rather than a parallel construction route with its own rules. The layout follows SQS and SSM: a resource factory, a table creator, a property parser, and a property rules file holding what is not simulated.

Supported properties are `KeySchema`, `AttributeDefinitions`, `BillingMode`, `ProvisionedThroughput`, `TableClass`, `DeletionProtectionEnabled` and `TableName`. When `TableName` is absent, the name is derived from the stack name and the logical ID, following `SimCfnSqsQueueName`, so it stays the same between deployments of one template and differs between two stacks deploying that template.

`Ref` returns the table name, and `Fn::GetAtt` `Arn` returns the table ARN. `Fn::GetAtt` `StreamArn` is refused by name, since streams are not simulated and an invented ARN would read as a working stream.

A property this simulation does not model records the resource as skipped, with a reason naming the property, rather than failing the whole template. That builds on the best-effort CloudFormation model being introduced separately, where a stack deploys what it can and reports what it left out.

## Example

```typescript
  stackName: "orders-stack",
  template: {
    Resources: {
      OrdersTable: {
        Type: "AWS::DynamoDB::Table",
        Properties: {
          KeySchema: [{ AttributeName: "id", KeyType: "HASH" }],
          AttributeDefinitions: [{ AttributeName: "id", AttributeType: "S" }],
        },
      },
    },
    Outputs: {
      OrdersTableName: { Value: { Ref: "OrdersTable" } },
      OrdersTableArn: { Value: { "Fn::GetAtt": ["OrdersTable", "Arn"] } },
    },
  },
});
```

With no `TableName` given, `Ref` resolves to the generated name, which is what a Lambda environment variable would carry.

## Acceptance criteria

- [ ] A template with an `AWS::DynamoDB::Table` resource deploys a table that `DescribeTable` finds, carrying the `KeySchema` and `AttributeDefinitions` the template gave
- [ ] A table with no `TableName` gets a name derived from the stack name and logical ID, and one template deployed under two stack names gives two differently named tables
- [ ] Property validation matches `CreateTable`, because the resource is created by calling it
- [ ] `Ref` resolves to the table name, `Fn::GetAtt Arn` to the table ARN, and `Fn::GetAtt StreamArn` is refused with a message naming the attribute
- [ ] An unmodelled property records the resource as skipped with a reason naming that property, and the rest of the stack still deploys
- [ ] A `.loc.test.ts` synthesises a CDK stack with a table, deploys it, then writes and reads an item through the deployed table
- [ ] `AWS::DynamoDB::Table` is added to the resource type list in `docs/services/cloudformation/README.md`, and the DynamoDB docs page gains a CloudFormation section with an extractable example

## Out of scope

`AWS::DynamoDB::GlobalTable` entirely. Secondary index properties, `Tags` and `TimeToLiveSpecification`, each of which arrives with the feature itself. Update and replacement semantics. Stack deletion: simulated CloudFormation has only `CreateStack` and `DescribeStacks`, so there is no `DeleteStack` for a table deletion to hang off yet.

Builds on #251 and #252.

hughgrigg:
…urces

AWS::DynamoDB::Table deploys through the ordinary CreateTable path, so a table a template deployed is the same thing an SDK caller would have got. Ref gives the table name and Fn::GetAtt Arn the table ARN. StreamArn is refused by name, since streams are not simulated. A property this simulation does not model skips the resource with a reason naming it, and the rest of the stack still deploys.

The generated name for an unnamed resource moves out of SimCfnSqsQueueName into a shared SimCfnGeneratedResourceName, parameterised by the length limit each service applies.

Resolves https://github.com/KensioSoftware/yulin/issues/255

### DIFF
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -972,6 +972,7 @@ The resource types it creates are:
 - `AWS::CloudFront::Distribution` and `AWS::CloudFront::Function`
 - `AWS::Cognito::UserPool`, `AWS::Cognito::UserPoolClient` and
   `AWS::Cognito::UserPoolGroup`
+- `AWS::DynamoDB::Table`
 - `AWS::IAM::Role`, `AWS::IAM::ManagedPolicy` and `AWS::IAM::Policy`
 - `AWS::KMS::Key` and `AWS::KMS::Alias`
 - `AWS::Lambda::Function`, `AWS::Lambda::Url` and `AWS::Lambda::Permission`
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -486,6 +486,95 @@ await simAws.backgroundTasksComplete();
 
 Creating a name that is already taken in the same scope fails with `ResourceInUseException`.
 
+## Deploying a table from CloudFormation
+
+Simulated CloudFormation creates a table from an `AWS::DynamoDB::Table` resource, in the stack's
+account and region. The table is created through `CreateTable`, so a template-created table is the
+same thing an SDK caller would get: the same name validation, the same key schema and attribute
+definition rules, the same ARN.
+
+`Ref` on the resource gives the table name, as it does on real AWS, so it can be handed straight to
+`PutItem`. `Fn::GetAtt … Arn` gives the table ARN, which is what an IAM policy names it by.
+
+```typescript sim-dynamodb-cloudformation-table
+/**
+ * Deploying a table from a CloudFormation template and writing to it.
+ */
+
+import { PutItemCommand } from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+
+const stack = await simAws.cloudFormation().deployTemplate({
+  stackName: "orders-stack",
+  template: {
+    Resources: {
+      OrdersTable: {
+        Type: "AWS::DynamoDB::Table",
+        Properties: {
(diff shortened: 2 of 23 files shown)


hughgrigg:
@coderabbitai rate limit

PR #280, opened 2026-07-31
feat: project attributes from simulated DynamoDB reads

### LINKED ISSUE #256: Project attributes from simulated DynamoDB reads with projection expressions
## Problem

Simulated DynamoDB handles no expressions at all. Nothing under `src/` mentions `ProjectionExpression`, `ConditionExpression` or `ExpressionAttributeNames`, so every expression parameter is currently absent rather than half done.

Most of the DynamoDB API is expressions: projections, conditions, filters, key conditions and updates. They share three parts, which are a tokeniser, document paths, and placeholder substitution. Writing those three times per feature is how the parsers drift apart and start disagreeing about what `a.b[0]` means. This issue builds them once under `expression/`, and ships them with the smallest real consumer so the code is callable rather than sitting unused: `ProjectionExpression` on `GetItem`.

## Desired behaviour

A tokeniser and a document path parser handling dot dereferencing and `[n]` indexing, with the 32 level nesting cap DynamoDB documents. A negative index, a fractional index and a path past the cap are each a `ValidationException` naming the path.

Placeholder substitution for `ExpressionAttributeNames` and `ExpressionAttributeValues` that fails closed in both directions. A `#name` or `:value` the request does not define is a `ValidationException`, and so is a defined entry no expression uses. Real DynamoDB rejects both, and the unused case is the one people hit after editing an expression and leaving the old placeholder behind.

`ProjectionExpression` on `GetItem` takes comma-separated document paths and returns only those, with the nested shape preserved: projecting `a.b` returns a map `a` containing only `b`, and projecting `list[0]` returns a one element list. A projected path the stored item does not have is silently omitted, not an error and not a `NULL`.

## Example

```typescript
const output = await simAws.dynamoDb().getItem(
  new GetItemCommand({
    TableName: "FoobarTable",
    Key: { id: { S: "order-1" } },
    ProjectionExpression: "#s, address.city, lines[0]",
    ExpressionAttributeNames: { "#s": "status" },
  }),
);
```

`output.Item` holds `status`, an `address` map containing only `city`, and a one element `lines` list. Dropping `lines[0]` from the expression while leaving an unused entry in `ExpressionAttributeNames` is a `ValidationException`.

## Acceptance criteria

- [ ] Document paths parse dot dereferencing and `[n]` indexing, with a `ValidationException` for a negative index, a fractional index and a path deeper than 32 levels
- [ ] A placeholder with no matching `ExpressionAttributeNames` or `ExpressionAttributeValues` entry is a `ValidationException` naming the placeholder
- [ ] A defined `ExpressionAttributeNames` or `ExpressionAttributeValues` entry that no expression uses is a `ValidationException`
- [ ] `GetItem` with a `ProjectionExpression` returns only the projected paths, preserving nested map and list shape
- [ ] A projected path the item does not have is left out of `Item` rather than raising an error
- [ ] The `ProjectionExpression` refusal on `GetItem` is lifted, and the DynamoDB docs page gains a projection section with an extractable example

## Out of scope

Condition, filter, key condition and update expressions, each of which reuses this parsing later. `ProjectionExpression` on `BatchGetItem`, `Query` and `Scan`, which lift their own refusals as those commands arrive. The legacy `AttributesToGet` parameter, which stays refused. `Select` and its validation matrix. The 4 KB expression length and 255 byte placeholder quotas, which are documented as unenforced.

## Implementation notes

The expression code is shared by most of the roadmap that follows, so it belongs in `src/service/dynamodb/expression/` rather than inside `command/get-item/`. Moving it later means moving it while four commands import it.

Builds on #254.

hughgrigg:
GetItem takes a ProjectionExpression, returning only the document paths it names with the nested map and list shape kept. A path the item does not have is left out rather than refused.

The tokenising, document path parsing and placeholder substitution behind it live under expression/, shared with the condition, filter and update expressions that follow, so the parsers cannot drift apart over what a.b[0] means. Placeholders fail closed in both directions, and two paths where one contains the other are refused as real DynamoDB refuses them.

Resolves https://github.com/KensioSoftware/yulin/issues/256

## Summary by CodeRabbit

* **New Features**
  * Added support for DynamoDB `GetItem` projection expressions.
  * Supports nested attributes, list indexes, document paths, and expression-name placeholders.
  * Returns only requested attributes, omitting unavailable values and compacting projected lists.
  * Distinguishes existing items with no projected fields from missing items.
  * Added validation for malformed expressions, invalid paths, unused placeholders, and nesting limits.

* **Documentation**
  * Added usage examples and comprehensive projection-expression guidance.

* **Tests**
  * Added coverage for projection behavior, validation, nested paths, aliases, list indexing, and edge cases.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -367,6 +367,89 @@ reports nothing removed. Its `ReturnValues` takes `NONE` and `ALL_OLD`, as `PutI
 
 Both take the table's name or its ARN, as the table commands do.
 
+## Projecting attributes
+
+`GetItem` takes a `ProjectionExpression`, which is a comma-separated list of document paths. Only
+those paths come back. A path is an attribute name, then any number of `.attribute` dereferences and
+`[n]` list indexes: `address.city`, `lines[0].sku`.
+
+An attribute name that is a DynamoDB reserved word, or that has a character an expression cannot
+carry, is written as a `#name` placeholder and defined in `ExpressionAttributeNames`.
+
+```typescript sim-dynamodb-projection-expression
+/**
+ * Reading part of an item with a projection expression.
+ */
+
+import {
+  CreateTableCommand,
+  GetItemCommand,
+  PutItemCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
+
+await dynamoDb.createTable(
+  new CreateTableCommand({
+    TableName: "FoobarTable",
+    KeySchema: [{ AttributeName: "orderId", KeyType: "HASH" }],
+    AttributeDefinitions: [{ AttributeName: "orderId", AttributeType: "S" }],
+    BillingMode: "PAY_PER_REQUEST",
+  }),
+);
+await simAws.backgroundTasksComplete();
+
+await dynamoDb.putItem(
+  new PutItemCommand({
+    TableName: "FoobarTable",
+    Item: {
+      orderId: { S: "order-1" },
+      status: { S: "shipped" },
+      address: { M: { city: { S: "Leeds" }, postcode: { S: "LS1 1AA" } } },
+      lines: { L: [{ S: "widget" }, { S: "gasket" }] },
+    },
+  }),
+);
+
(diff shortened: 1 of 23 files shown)


hughgrigg:
@coderabbitai full review

hughgrigg:
@coderabbitai review

PR #281, opened 2026-07-31
feat: guard simulated DynamoDB writes with condition expressions

### LINKED ISSUE #257: Guard simulated DynamoDB writes with condition expressions
## Problem

A conditional write is how DynamoDB application code does insert if absent and optimistic locking. Simulated DynamoDB has no `ConditionExpression` anywhere, so every write is an unconditional overwrite. A test cannot tell a correct conditional write from one whose condition was dropped, because both produce the same stored item.

This is also where DynamoDB's comparison and ordering rules go. They have no home in the codebase yet, and sort key conditions and filter expressions both need the same ones later, so they are built here as a separate collaborator rather than inline in the evaluator.

## Desired behaviour

A condition expression parser and evaluator on top of the expression parsing added with projection expressions.

The grammar covers the six comparators, `BETWEEN` with inclusive bounds, `IN` with up to 100 operands, `AND`, `OR`, `NOT`, parentheses, and the functions `attribute_exists`, `attribute_not_exists`, `attribute_type`, `begins_with`, `contains` and `size`. Precedence follows the documented order, so `NOT` binds tighter than `AND`, and `AND` tighter than `OR`.

Comparison rules: strings compare by UTF-8 byte order, numbers compare numerically as decimals rather than as doubles, and binary compares as unsigned bytes. A comparison between two different types evaluates false rather than throwing, which is what real DynamoDB does. `attribute_exists` is true for an attribute stored as `NULL`. `contains(a, a)` is an error.

`ConditionExpression` is accepted on `PutItem` and `DeleteItem`. A condition that evaluates false leaves the table unchanged and throws `ConditionalCheckFailedException`, with `.name` set to `ConditionalCheckFailedException` and the message `The conditional request failed.` so SDK error handling matches. `ReturnValuesOnConditionCheckFailure` set to `ALL_OLD` puts the existing item on the exception, and `NONE` leaves it off.

## Example

```typescript
  new PutItemCommand({
    TableName: "FoobarTable",
    Item: { id: { S: "order-1" }, version: { N: "2" } },
    ConditionExpression: "attribute_exists(id) AND version = :expected",
    ExpressionAttributeValues: { ":expected": { N: "1" } },
  }),
);
```

When the stored `version` is not 1, nothing is written and the call throws `ConditionalCheckFailedException`.

## Acceptance criteria

- [ ] The six comparators, `BETWEEN`, `IN`, `AND`, `OR`, `NOT` and parentheses parse and evaluate, with precedence covered by a test that fails if `AND` and `OR` are swapped
- [ ] `attribute_exists`, `attribute_not_exists`, `attribute_type`, `begins_with`, `contains` and `size` each evaluate against a stored item, with `attribute_exists` true for a stored `NULL`
- [ ] Strings compare by UTF-8 byte order, numbers numerically and binary by unsigned byte order, and a comparison between two different types evaluates false rather than throwing
- [ ] A false condition on `PutItem` or `DeleteItem` leaves the item as it was and throws `ConditionalCheckFailedException` with the AWS name and message
- [ ] `ReturnValuesOnConditionCheckFailure` `ALL_OLD` populates the exception's `Item`, and `NONE` leaves it absent
- [ ] The `ConditionExpression` refusals on `PutItem` and `DeleteItem` are both lifted by name, so neither is left behind
- [ ] The DynamoDB docs page covers insert if absent with `attribute_not_exists` and optimistic locking with a version attribute, with an extractable example

## Out of scope

Key condition expressions and filter expressions, which reuse this evaluator when `Query` and `Scan` land. Update expressions. `UpdateItem` and transactional condition checks, which wire the same evaluator in when those commands exist. The 4 KB expression length and 300 operator quotas, which are documented as unenforced.

## Implementation notes

The comparison rules are the piece with the longest reach, so they want to be their own collaborator rather than private helpers in the evaluator: sort key conditions compare the same way. If the change runs past a reviewable size, `size`, `contains` and `attribute_type` are the natural cut, refused by name and added in a follow-up.

Builds on #256 and #254.

hughgrigg:
PutItem and DeleteItem take a ConditionExpression, checked against whatever is stored under the key before anything changes. A condition that does not hold leaves the item as it was and throws ConditionalCheckFailedException with the name and message real DynamoDB uses.

The comparison rules live beside the value model rather than inside the evaluator, since sort key conditions and filter expressions compare the same way: strings by UTF-8 bytes, numbers by their digits rather than by what they round to, and binary as unsigned bytes. Two different types have no order, so a comparison between them is false rather than an error.

Resolves https://github.com/KensioSoftware/yulin/issues/257

## Summary by CodeRabbit

* **New Features**
  * Added conditional writes for DynamoDB `PutItem` and `DeleteItem`.
  * Supports comparison operators, `BETWEEN`, `IN`, logical operators, document paths, and condition functions.
  * Failed conditions leave data unchanged and can return the previous item with `ALL_OLD`.
  * Added DynamoDB-compatible validation and conditional failure errors.

* **Documentation**
  * Added usage guidance and examples, including duplicate prevention and optimistic locking.

* **Tests**
  * Expanded coverage for successful writes, failures, validation, and returned prior attributes.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -459,13 +459,18 @@ works as well as `AND`.
 The functions are `attribute_exists`, `attribute_not_exists`, `attribute_type`, `begins_with`,
 `contains` and `size`. Function names are read in lower case only, as they are on real AWS.
 `attribute_exists` is true for an attribute stored as `NULL`, since `NULL` is a value rather than an
-absent one. `size` is a number rather than a condition, so it goes beside a comparator: a string and
+absent one. The first operand of every one of them names a path in the item, so a supplied value
+there is refused rather than compared. `size` is a number rather than a condition, so it goes beside a comparator: a string and
 binary measure in bytes, and a set, a list or a map in how many things it holds.
 
 Strings compare by UTF-8 byte order, numbers compare by their digits rather than by what they round
-to, and binary compares as unsigned bytes. A comparison between two different types is false rather
-than an error, and so is one naming a path the item does not have. That is what real DynamoDB does,
-and it is what lets one condition guard items that do not all carry the same attributes.
+to, and binary compares as unsigned bytes.
+
+A comparison between two different types is never an error. Equality works across types, so a string
+and a number are not equal: `=` is false and `<>` is true. Ordering does not, so `<`, `<=`, `>` and
+`>=` are all false between them, as they are for a path the item does not have. That is what real
+DynamoDB does, and it is what lets one condition guard items that do not all carry the same
+attributes.
 
 `ExpressionAttributeNames` and `ExpressionAttributeValues` have to agree exactly with the expression,
 in both directions: a placeholder the request does not define is a `ValidationException`, and so is
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -367,6 +367,110 @@ reports nothing removed. Its `ReturnValues` takes `NONE` and `ALL_OLD`, as `PutI
 
 Both take the table's name or its ARN, as the table commands do.
 
+## Conditional writes
+
+`PutItem` and `DeleteItem` take a `ConditionExpression`, which is checked against whatever is stored
+under the key before anything changes. A condition that does not hold leaves the item exactly as it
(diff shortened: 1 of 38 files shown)


hughgrigg:
@coderabbitai review

PR #282, opened 2026-07-31
feat: update simulated DynamoDB items with SET and REMOVE update expr…

### LINKED ISSUE #258: Update simulated DynamoDB items with SET and REMOVE update expressions
## Problem

Simulated DynamoDB can replace a whole item with PutItem, but it cannot change part of one. Most application write paths touch a single attribute and go through UpdateItem, so a test either rewrites that call as a read, a mutation in the test and a full put, which proves nothing about the expression the application actually sends, or it leaves that path untested.

The SDK intercept router at `src/service/dynamodb/sdk/sim-dynamodb-sdk-command-router.ts` carries four routes: CreateTable, DescribeTable, ListTables and PutItem. An intercepted `UpdateItemCommand` has nowhere to go.

## Desired behaviour

UpdateItem parses an `UpdateExpression` using the tokeniser, document path parser and placeholder substitution built for condition expressions. Each clause keyword appears at most once, the clauses may come in either order, and actions within a clause are comma separated. SET takes `path = operand`, where an operand is a placeholder, another document path, or `if_not_exists(path, value)`. Update expressions have no literals, so every constant arrives through `ExpressionAttributeValues`, and an unresolved or defined-but-unused placeholder fails closed the way it already does for conditions. REMOVE deletes an attribute, and removing an attribute that is not there succeeds without changing the item.

The rule that decides whether this is simulated correctly: every action is evaluated against a snapshot of the item as it stood before the update, rather than being applied left to right.

UpdateItem upserts when the key matches no item, building the new item from the `Key` plus the SET actions. Assigning into a map whose parent attribute does not exist is a `ValidationException`. `ConditionExpression` works through the condition expression support already present, and `ReturnValues` covers NONE, ALL_OLD and ALL_NEW. The legacy `AttributeUpdates` parameter is refused by name alongside the other unsimulated inputs, through a collaborator in the shape of `src/service/sqs/command/queue/sim-sqs-unsimulated-queue-input.ts`.

## Example

Given the item `{ a: 1, b: 2, c: 3 }`, this expression:

```
REMOVE a SET b = a, c = b
```

leaves `{ b: 1, c: 2 }`. Both assignments read pre-update values, and `a` is still readable by the SET actions even though the REMOVE clause is written first. An implementation that applies actions left to right removes `a` before the assignment can read it and answers differently.

## Acceptance criteria

- [ ] SET and REMOVE are accepted in either order, each keyword at most once, with a repeated keyword giving a `ValidationException`
- [ ] Every action reads the pre-update snapshot, so `REMOVE a SET b = a, c = b` on `{ a: 1, b: 2, c: 3 }` leaves `{ b: 1, c: 2 }`
- [ ] `if_not_exists(path, value)` returns the stored value when the path is present and the operand value when it is not
- [ ] UpdateItem against a key with no item creates the item from the `Key` plus the SET actions
- [ ] `ReturnValues` NONE, ALL_OLD and ALL_NEW return their documented shapes, with no `Attributes` for ALL_OLD when there was no item
- [ ] `AttributeUpdates` is refused, naming the parameter, as are the ADD and DELETE keywords
- [ ] A docs section in `docs/services/dynamodb/` with an extractable example covers the snapshot rule

## Out of scope

Arithmetic in SET, `list_append`, list element paths, and the ADD and DELETE clauses, which follow in their own issue along with the UPDATED_OLD and UPDATED_NEW `ReturnValues` modes. Index maintenance on update. `AttributeUpdates`, which stays refused permanently.

Builds on #256 and #257.

hughgrigg:
…essions

UpdateItem reads an UpdateExpression made of SET and REMOVE clauses, in either order and each at most once. Every action is evaluated against the item as it stood before the request, so the actions of one expression do not see each other's work. It upserts from the Key when nothing is stored, takes a ConditionExpression sharing its placeholders, and answers NONE, ALL_OLD or ALL_NEW.

Resolves https://github.com/KensioSoftware/yulin/issues/258

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -367,10 +367,135 @@ reports nothing removed. Its `ReturnValues` takes `NONE` and `ALL_OLD`, as `PutI
 
 Both take the table's name or its ARN, as the table commands do.
 
+## Updating items
+
+`UpdateItem` changes part of an item, where `PutItem` replaces the whole thing. What to change is
+written as an `UpdateExpression` made of a `SET` clause, a `REMOVE` clause, or both in either order.
+Each keyword appears at most once, and the actions inside a clause are separated by commas.
+
+A `SET` action is `path = operand`, where an operand is a value from `ExpressionAttributeValues`,
+another document path, or `if_not_exists(path, operand)`. An update expression carries no literals,
+so every constant arrives through `ExpressionAttributeValues`. A `REMOVE` action is a document path
+on its own, and removing an attribute that is not there succeeds without changing the item.
+
+Every action reads the item as it stood before the request, rather than the item being built. So
+this expression, against `{ a: 1, b: 2, c: 3 }`:
+
+```text
+REMOVE a SET b = a, c = b
+```
+
+leaves `{ b: 1, c: 2 }`. Both assignments read the values from before the update, and `a` is still
+there to be read even though the `REMOVE` is written first.
+
+```typescript sim-dynamodb-update-item
+/**
+ * Changing part of an item, against the item as it stood before the update.
+ */
+
+import {
+  CreateTableCommand,
+  PutItemCommand,
+  UpdateItemCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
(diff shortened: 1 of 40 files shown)


hughgrigg:
@coderabbitai rate limit

PR #283, opened 2026-07-31
feat: increment numbers, append to lists and update sets in simulated…

### LINKED ISSUE #259: Increment numbers, append to lists and update sets in simulated DynamoDB
## Problem

`SET count = count + :one` and `SET history = list_append(history, :entry)` are two of the most common UpdateItem shapes in application code, and the SET and REMOVE support has neither an operator nor a list function, so both are rejected. ADD and DELETE are refused by keyword, and they are the only way to change a set attribute in place. Set attributes are effectively write-only: a test can put one, but nothing can add a member to it or take one away.

## Desired behaviour

SET gains a single binary plus or minus over two operands, each a document path or a placeholder. DynamoDB allows exactly one operator with no chaining and no parentheses, and the parser refuses more. Arithmetic against an attribute that is not present is a `ValidationException`, which is why `if_not_exists(count, :zero) + :one` is the usual counter form. The arithmetic runs on the decimal number representation the item model holds rather than on JavaScript numbers, so adding 1 to 9007199254740993 answers 9007199254740994 where a double based implementation answers 9007199254740992.

`list_append(list1, list2)` concatenates in argument order. A SET at a list index past the end appends rather than creating a sparse list, and REMOVE of a list element shifts the later elements down, so removing index 0 twice removes the first two elements. `ReturnValues` gains UPDATED_OLD and UPDATED_NEW, which return only the attributes the expression touched.

ADD writes a whitespace separated path and value with no operator between them. A number is added mathematically, with a missing attribute treated as 0 and a negative value subtracting. A set value is unioned into the stored set, and an error follows when the two set types differ. DELETE is set subtraction only: an empty set operand is refused, and the operand's set type must match the stored one. Both clauses are refused against any other attribute type, matching the AWS restriction to Number and set types, and both apply to top-level attributes only. The docs section repeats the AWS recommendation to prefer SET over ADD for numbers, since a retried ADD applies twice.

## Example

```typescript
await simAws.dynamoDb().updateItem(
  new UpdateItemCommand({
    TableName: "FoobarTable",
    Key: { id: { S: "foobar" } },
    UpdateExpression:
      "SET viewCount = if_not_exists(viewCount, :zero) + :one ADD tags :added",
    ExpressionAttributeValues: {
      ":zero": { N: "0" },
      ":one": { N: "1" },
      ":added": { SS: ["archived"] },
    },
    ReturnValues: "UPDATED_NEW",
  }),
);
```

## Acceptance criteria

- [ ] `SET a = a + :n` and `SET a = a - :n` evaluate as decimals, with a chained or parenthesised operator refused
- [ ] Arithmetic against an absent attribute is a `ValidationException`, and `if_not_exists` guards it
- [ ] `list_append` concatenates in argument order, a SET past the end of a list appends, and REMOVE of a list element shifts later elements down
- [ ] ADD on a number treats a missing attribute as 0 and a negative value as a subtraction, and ADD on a set unions with a mismatched set type refused
- [ ] DELETE subtracts set members, refusing an empty set operand and a set type that does not match
- [ ] ADD and DELETE against a String, Binary, List or Map attribute are refused, as are nested paths
- [ ] UPDATED_OLD and UPDATED_NEW return only the attributes the expression touched

## Out of scope

Atomic counters under concurrency, which an in-process simulator serialises trivially. Index maintenance on update. `AttributeUpdates`, which stays refused.

Builds on #258.

hughgrigg:
… DynamoDB

SET gains one plus or minus over two operands, worked out on the decimal digits an item holds rather than through a JavaScript number, and list_append. A SET past the end of a list appends and a REMOVE of a list element closes the list up, with every index read against the stored item. ADD and DELETE change a number or a set in place, on a top-level attribute. ReturnValues gains UPDATED_OLD and UPDATED_NEW, which report the parts the expression touched.

Resolves https://github.com/KensioSoftware/yulin/issues/259

## Summary by CodeRabbit

* **New Features**
  * Expanded DynamoDB `UpdateItem` support for `ADD`, `DELETE`, arithmetic, `list_append`, list indexing, and set operations.
  * Added support for `UPDATED_OLD` and `UPDATED_NEW` return values, including changed-attribute results.
  * Improved decimal precision for numeric updates.
* **Bug Fixes**
  * Improved validation and handling of overlapping update actions, list removals, and invalid operands.
* **Documentation**
  * Added comprehensive usage guidance and examples for counters, lists, and set tagging.


hughgrigg:
@coderabbitai review

PR #284, opened 2026-08-01
feat: read and write simulated DynamoDB items in batches

### LINKED ISSUE #260: Read and write simulated DynamoDB items in batches
## Problem

Application code that writes a page of items, or loads several items by key, uses BatchWriteItem and BatchGetItem. Neither is simulated, so a test has to be rewritten as a loop of single-item calls. That drops the rules which only exist on the batch path: the 25 request cap, the whole-batch rejection when the same item appears twice, and `ConsistentRead` being set per table rather than per call.

## Desired behaviour

BatchWriteItem takes `RequestItems` as a map of table name or ARN to at most 25 `WriteRequest` entries, each carrying exactly one `PutRequest` or `DeleteRequest`. A put is a full replace, exactly as PutItem is, and deleting an item that is not there succeeds. Six failures reject the whole request rather than one entry: an unknown table, key attributes that do not match the table's key schema, more than one operation on the same item, more than 25 requests, an item over 400 KB, and a total request over 16 MB. No `ConditionExpression` is accepted and no items come back. `UnprocessedItems` is present and empty rather than absent.

BatchGetItem takes a map of table name or ARN to a `KeysAndAttributes` holding `Keys`, `ConsistentRead` and `ProjectionExpression`, so both consistency and projection are settled per table. More than 100 keys in one call is a `ValidationException`, as is the same key appearing twice within one table. An item that does not exist is simply absent from `Responses`, with no placeholder entry, and the order of `Responses` is not guaranteed. `UnprocessedKeys` is present and empty. The projection support already added for GetItem evaluates the per-table `ProjectionExpression`, so a batch read projects the same way a single read does.

## Implementation notes

The SQS batch collaborator at `src/service/sqs/command/message/sim-sqs-batch-entries.ts` is not the model to copy here. It reports per-entry failures in a `Failed` array while the rest of the batch goes through. DynamoDB has no per-entry failure list at all, so one bad entry takes the whole call down with an exception and nothing is written.

`UnprocessedItems` and `UnprocessedKeys` are always empty, because the simulator models no throughput limits. The docs section states that under Limitations and still shows the retry loop, since the branch that real code needs is never exercised against the simulator.

## Acceptance criteria

- [ ] BatchWriteItem accepts a table name or an ARN, applies puts as full replaces and deletes idempotently, and returns an empty `UnprocessedItems` map rather than omitting it
- [ ] Two operations on the same item in one call reject the whole batch, leaving no writes applied
- [ ] More than 25 write requests, an item over 400 KB and a total request over 16 MB each reject the whole batch
- [ ] `ConditionExpression` on a write request is refused, naming the parameter
- [ ] BatchGetItem sets `ConsistentRead` and `ProjectionExpression` per table rather than per call
- [ ] More than 100 keys in one call, or a duplicate key within one table, is a `ValidationException`
- [ ] A missing item is absent from `Responses` with no placeholder, and `UnprocessedKeys` is present and empty

## Out of scope

Fault injection to force unprocessed items or keys, which is worth a follow-up but changes the API shape. `ConsumedCapacity` and `ItemCollectionMetrics`. Transactional reads and writes. `BatchExecuteStatement`, which is PartiQL and stays under Limitations.

Builds on #254 and #256.

hughgrigg:
BatchWriteItem puts and deletes items across tables in one call, and BatchGetItem reads them by primary key. Both take a table name or ARN in RequestItems and answer with an empty UnprocessedItems or UnprocessedKeys.

A batch write is refused whole rather than partly applied: an unknown table, a key that does not match the key schema, two operations on one item, more than 25 requests or an item over 400 KB each leave nothing written. A batch read settles ConsistentRead and ProjectionExpression per table, and leaves out a key that holds nothing.

The 16 MB request limit is not enforced, and is documented under Limitations with the reason.

Resolves https://github.com/KensioSoftware/yulin/issues/260

## Summary by CodeRabbit

* **New Features**
  * Added DynamoDB batch read and write support across multiple tables.
  * Batch reads support projections, consistency settings, missing items, and table ARN references.
  * Batch writes support putting and deleting items with atomic validation.
  * Added SDK command routing and practical batch operation examples.

* **Documentation**
  * Documented supported batch operations, limits, validation behavior, and current limitations.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -891,9 +891,12 @@ Six things take the whole batch down rather than one entry of it, leaving nothin
 - a table that is not there
 - key attributes that do not match the table's key schema
 - more than one operation on the same item of one table
+- one table named twice, once by its name and once by its ARN
 - more than 25 write requests, counted across every table the request names
 - an item over the 400 KB an item holds
-- a request over 16 MB
+
+Real DynamoDB also refuses a request over 16 MB. That one is not simulated, for the reason under
+Limitations.
 
 The same key in two different tables is two items rather than one, so a batch may write both.
 
@@ -981,14 +984,17 @@ caller that needs to tell its items apart reads the key attributes off them rath
 where they are in the list.
 
 More than 100 keys in one call, counted across every table the request names, is a
-`ValidationException`. So is the same key twice for one table.
+`ValidationException`. So is the same key twice for one table, and so is one table named twice, once
+by its name and once by its ARN.
 
 Both commands answer with the map of what they could not get to, `UnprocessedItems` for a write and
 `UnprocessedKeys` for a read. Both are always empty here, since nothing is throttled, but they are
 there rather than absent, so the retry loop real code is written around still terminates:
 
 ```typescript
-let unprocessed = { OrdersTable: [{ PutRequest: { Item: item } }] };
+let unprocessed = {
+  OrdersTable: [{ PutRequest: { Item: { orderId: { S: "order-1" } } } }],
+};
 
 while (Object.keys(unprocessed).length > 0) {
   const output = await dynamoDb.batchWriteItem(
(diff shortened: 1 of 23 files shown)


hughgrigg:
@coderabbitai full review

PR #285, opened 2026-08-01
feat: read and write simulated DynamoDB items in transactions

### LINKED ISSUE #261: Read and write simulated DynamoDB items in transactions
## Problem

Simulated DynamoDB has no transaction commands, and no read path at all: the items map on `SimDynamoDbTable` is written by `PutItem` and never read anywhere in the tree. Application code that uses `TransactWriteItems` to keep two items consistent, such as a ledger entry and the balance it moves, cannot be exercised against the simulator.

The behaviour worth simulating is the failure path. A transaction that succeeds looks the same as two separate writes, so it proves little. What a test needs to show is that a failed condition on the second action left the first one unwritten, and that the caller can tell which action failed. That means modelling the shape of `TransactionCanceledException`, not just its name.

## Desired behaviour

`TransactWriteItems` takes up to 100 ordered actions, each carrying exactly one of `Put`, `Update`, `Delete` and `ConditionCheck`, with a 4 MB aggregate and no two actions targeting the same item in the same table. The whole group is applied in one synchronous step, so atomicity follows from how it is applied rather than from unwinding a partial write.

When any condition fails, nothing is written and the call throws `TransactionCanceledException` carrying `CancellationReasons` positionally aligned with `TransactItems`. There is one entry per action, including the actions that would have succeeded, which carry the code `None`. The codes have no `Exception` suffix: `ConditionalCheckFailed`, `ItemCollectionSizeLimitExceeded`, `ValidationError` and `ProvisionedThroughputExceeded`. An action with `ReturnValuesOnConditionCheckFailure` set to `ALL_OLD` gets `Item` populated on its cancellation reason.

`ClientRequestToken` makes a retry idempotent. Replaying a token with the same payload inside the ten minute window succeeds without applying the writes again, and a changed payload inside that window gives `IdempotentParameterMismatchException`. The window is measured on the simulated clock, so `advanceBy(...)` is what moves a test past it.

`TransactGetItems` takes up to 100 `Get` actions and is always strongly consistent, with no `ConsistentRead` parameter to set. Its `Responses` array is positional and is never compacted: a retrieved item gives an entry with `Item`, and a missing item gives an entry with no `Item` key.

## Example

```typescript
const cancelled = await simAws
  .dynamoDb()
  .transactWriteItems(
    new TransactWriteItemsCommand({ TransactItems: [ledgerEntry, balanceUpdate] }),
  )
  .catch((thrown: TransactionCanceledException) => thrown);

expect(cancelled.CancellationReasons).toEqual([
  { Code: "None" },
  { Code: "ConditionalCheckFailed", Message: "The conditional request failed." },
]);
```

The first action is reported even though nothing was wrong with it, and neither item was written.

## Acceptance criteria

- [ ] A group of up to 100 actions applies in one step, and nothing is written when any action's condition fails
- [ ] An action carrying more than one of `Put`, `Update`, `Delete` and `ConditionCheck` is refused, as are two actions targeting the same item in the same table
- [ ] `TransactionCanceledException` carries one `CancellationReason` per action in request order, with `Code` `None` for the actions that would have succeeded
- [ ] `ReturnValuesOnConditionCheckFailure` `ALL_OLD` populates `CancellationReason.Item` with the item as it was
- [ ] Replaying a `ClientRequestToken` with the same payload inside ten simulated minutes does not apply the writes twice, and a changed payload gives `IdempotentParameterMismatchException`
- [ ] `TransactGetItems` returns a positional `Responses` array in which a missing item is an entry with no `Item` key

## Out of scope

`ConsumedCapacity`, including the doubled read and write accounting. Transactional reads and writes through a secondary index, which AWS does not support either. `ProjectionExpression` on the `Get` actions, which is refused until projection expressions land. `TransactionConflictException` and `TransactionInProgressException` are not modelled: an in-process simulator serialises every call, so neither is reachable, and the docs say so rather than the code pretending otherwise.

Builds on #257 and #259.

hughgrigg:
TransactWriteItems applies up to 100 Put, Update, Delete and ConditionCheck actions in one step, and cancels the whole request with positional CancellationReasons when any condition fails. ClientRequestToken makes a retry idempotent for ten simulated minutes. TransactGetItems reads up to 100 items with a positional Responses array in which a missing item is an entry with no Item.

Resolves https://github.com/KensioSoftware/yulin/issues/261

## Summary by CodeRabbit

* **New Features**
  * Added support for DynamoDB transactional writes, including atomic Put, Update, Delete, and condition-check operations.
  * Added transactional reads with ordered responses, projections, and support for missing items.
  * Added conditional cancellation reasons, transaction limits, validation, IAM authorization, and idempotent retries.
  * Added simulated DynamoDB transaction errors for cancellation and parameter mismatches.
  * Added examples demonstrating transactional reads and writes.

* **Documentation**
  * Expanded DynamoDB documentation with transaction capabilities, limitations, authorization, and usage guidance.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -1006,6 +1006,246 @@ while (Object.keys(unprocessed).length > 0) {
 }
 ```
 
+## Reading and writing items in transactions
+
+`TransactWriteItems` applies up to 100 actions in one step. Either all of them happen or none of
+them do, so two items that have to agree with each other can be written together.
+
+Each action carries exactly one of `Put`, `Update`, `Delete` and `ConditionCheck`, and names its own
+table. A `ConditionCheck` writes nothing: it is how a transaction says that an item it is not
+changing has to hold for the items it is changing to be written.
+
+What a test usually wants to show is the failure. A transaction that succeeds looks the same as two
+separate writes, so what is worth asserting is that a failed condition on the second action left the
+first one unwritten.
+
+```typescript sim-dynamodb-transact-write-items
+/**
+ * Writing a ledger entry and the balance it moves, or writing neither.
+ */
+
+import {
+  CreateTableCommand,
+  GetItemCommand,
+  PutItemCommand,
+  TransactWriteItemsCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
+
+await dynamoDb.createTable(
+  new CreateTableCommand({
+    TableName: "AccountsTable",
+    KeySchema: [{ AttributeName: "accountId", KeyType: "HASH" }],
+    AttributeDefinitions: [{ AttributeName: "accountId", AttributeType: "S" }],
+    BillingMode: "PAY_PER_REQUEST",
+  }),
+);
+await dynamoDb.createTable(
+  new CreateTableCommand({
+    TableName: "LedgerTable",
+    KeySchema: [{ AttributeName: "entryId", KeyType: "HASH" }],
+    AttributeDefinitions: [{ AttributeName: "entryId", AttributeType: "S" }],
+    BillingMode: "PAY_PER_REQUEST",
+  }),
+);
+await simAws.backgroundTasksComplete();
+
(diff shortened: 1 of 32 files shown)


hughgrigg:
@coderabbitai full review

PR #286, opened 2026-08-01
feat: tag simulated DynamoDB tables

### LINKED ISSUE #262: Tag simulated DynamoDB tables
## Problem

Simulated DynamoDB has no tag commands, and `CreateTable` does not look at `Tags` at all. A CDK app that calls `Tags.of(stack).add("Environment", "test")` puts a `Tags` property on every taggable resource in the template, tables included, so a tagged table is the normal case rather than an unusual one. Code that reads its own tags back, to find which environment a table belongs to or to check a cost allocation convention, has nothing to test against.

## Desired behaviour

`Tags` on `CreateTable`, plus `TagResource`, `UntagResource` and `ListTagsOfResource`. All three tag commands address the table by ARN through `ResourceArn`, and an ARN naming no table gives `ResourceNotFoundException`.

The constraints are DynamoDB's: at most 50 tags on a resource, keys up to 128 characters and values up to 256, and the documented character set of letters, digits, whitespace and `+ - = . _ : /`. Keys beginning `aws:` are reserved and refused. A key appears once, so tagging a key that is already present replaces its value rather than adding a second entry. Untagging a key that is not present is not an error. The exception DynamoDB raises when the 50 tag limit is exceeded should be checked against the real service before it is chosen, since the API reference makes both `LimitExceededException` and `ValidationException` plausible.

`TagResource` and `UntagResource` return an empty body, which makes `ListTagsOfResource` the only way a test can observe the result. It pages with `NextToken`. The API has no page size parameter, so the page size is the simulator's own choice and the docs should state what it is.

`Tags` also becomes a supported property on `AWS::DynamoDB::Table`, building on the best-effort CloudFormation model rather than the current whole-template refusal, so a CDK stack with tags applied deploys a tagged table.

One divergence to write down: AWS documents tagging as eventually consistent, so a `ListTagsOfResource` issued straight after a `TagResource` may not show the change. The simulator applies tag changes immediately. That belongs under Limitations.

## Example

```typescript
await simAws.dynamoDb().tagResource(
  new TagResourceCommand({
    ResourceArn: tableArn,
    Tags: [{ Key: "Environment", Value: "test" }],
  }),
);

const { Tags } = await simAws
  .dynamoDb()
  .listTagsOfResource(new ListTagsOfResourceCommand({ ResourceArn: tableArn }));
```

## Acceptance criteria

- [ ] `CreateTable` accepts `Tags`, and those tags come back from `ListTagsOfResource`
- [ ] Key length, value length, the allowed character set and the reserved `aws:` prefix are each refused with an error naming what was wrong
- [ ] A 51st tag on a resource is refused, and tagging a key that already exists replaces its value rather than adding a second entry
- [ ] All three tag commands give `ResourceNotFoundException` for an ARN naming no table
- [ ] `ListTagsOfResource` pages with `NextToken` and omits the token on the final page
- [ ] A CDK stack applying `Tags.of(...)` deploys a table carrying those tags
- [ ] IAM cases cover `dynamodb:TagResource`, `dynamodb:UntagResource` and `dynamodb:ListTagsOfResource`

## Out of scope

Tag based IAM condition keys such as `aws:RequestTag`. Tag propagation to secondary indexes, since no index exists yet. Tags on other DynamoDB resources such as backups and global table replicas.

Builds on #255.

hughgrigg:
CreateTable takes Tags, and TagResource, UntagResource and ListTagsOfResource work on a table by ARN, with DynamoDB's key, value, character set and 50 tag rules, and NextToken paging at 25 tags a page. AWS::DynamoDB::Table deploys its Tags, so a CDK stack applying Tags.of(...) gets a tagged table.

Resolves https://github.com/KensioSoftware/yulin/issues/262

## Summary by CodeRabbit

* **New Features**
  * Added DynamoDB table tagging support during table creation and through tag, untag, and list-tags operations.
  * Added CloudFormation support for deploying table tags.
  * Added validation for tag keys, values, reserved prefixes, and the 50-tag limit.
  * Added paginated tag listings with deterministic ordering and continuation tokens.

* **Documentation**
  * Updated DynamoDB documentation and examples to describe tagging behavior and limitations.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -228,6 +228,106 @@ console.log(description.Table?.TableStatus); // "ACTIVE"
 
 `DeleteTable` takes the table's name or its ARN, as `DescribeTable` does.
 
+## Tagging tables
+
+A table is tagged by `CreateTable`, or afterwards by `TagResource`. `UntagResource` takes tags off,
+and `ListTagsOfResource` reads them back. The three tag commands name their resource by ARN, in
+`ResourceArn`, where the table commands take a name or an ARN.
+
+```typescript sim-dynamodb-tag-table
+/**
+ * Tagging a table on creation and afterwards, and reading the tags back.
+ */
+
+import {
+  CreateTableCommand,
+  ListTagsOfResourceCommand,
+  TagResourceCommand,
+  UntagResourceCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
+
+const creation = await dynamoDb.createTable(
+  new CreateTableCommand({
+    TableName: "OrdersTable",
+    KeySchema: [{ AttributeName: "orderId", KeyType: "HASH" }],
+    AttributeDefinitions: [{ AttributeName: "orderId", AttributeType: "S" }],
+    BillingMode: "PAY_PER_REQUEST",
+    Tags: [{ Key: "Environment", Value: "test" }],
+  }),
+);
+await simAws.backgroundTasksComplete();
+
+const tableArn = creation.TableDescription?.TableArn ?? "";
+
+await dynamoDb.tagResource(
+  new TagResourceCommand({
+    ResourceArn: tableArn,
+    Tags: [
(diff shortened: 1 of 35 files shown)


hughgrigg:
@coderabbitai full review

PR #287, opened 2026-08-01
feat: expire simulated DynamoDB items with time to live

### LINKED ISSUE #263: Expire simulated DynamoDB items with time to live
## Problem

CDK sets `timeToLiveAttribute` on a table routinely, and a table that leans on time to live usually has a test asserting that an expired record went away. Simulated DynamoDB has neither `UpdateTimeToLive` nor `DescribeTimeToLive`, so that test cannot be written, and a template setting `TimeToLiveSpecification` has nowhere to put it.

The more interesting case is the system test. A session table, a Lambda function, a queue and a bucket all belong to one `SimAws`, so they all share one clock. Advancing that clock should expire the sessions the same way it moves everything else along, letting a test drive several moving parts forward together and assert on what the system did. That only works if expiry is something the clock causes, rather than something the test has to ask for separately.

See the [simulated time docs](https://yulinsim.dev/time/) for what the clock already does.

## Desired behaviour

`UpdateTimeToLive` and `DescribeTimeToLive` with the `TimeToLiveSpecification` shape and the four statuses `ENABLING`, `ENABLED`, `DISABLING` and `DISABLED`, moving through the background scheduler the way table status already does. A second `UpdateTimeToLive` on the same table within the hour is a `ValidationException`, matching the once per hour rule. That hour is measured on simulated time, so `simAws.clock().advanceBy({ hours: 1 })` is what makes the following call succeed.

The TTL attribute holds epoch seconds in a Number attribute. An item whose TTL attribute is absent, or holds a String or any other type, is never expired, and that is not an error.

### Expiry runs on the clock

Moving the simulated clock forward deletes items whose time to live has run out. There is nothing for the test to call. `simAws.clock().advanceBy({ days: 3 })` returns with the expired items already gone, alongside whatever else that advance caused elsewhere in the simulation.

The mechanism is `scheduleAt` on the background scheduler, which `SimClockControl` already dispatches in due order when time moves. `SimSecretsManagerSecretExpiry.scheduleRemoval` is the precedent: it schedules the removal and then re-checks the state at fire time, so a change in the meantime cancels it. The same re-check covers everything that should stop a TTL deletion:

- the item was deleted, or overwritten with a later TTL, or had its TTL attribute removed
- time to live was disabled on the table
- the attribute no longer holds a Number

Enabling time to live schedules expiry for items already in the table, since their TTL attributes were only inert while the feature was off.

### The deletion window

Real DynamoDB does not delete an item the moment its timestamp passes. It marks it expired, and deletes it typically within 48 hours. Until then reads keep returning the item, which is why the AWS docs tell you to filter expired items on read rather than trusting they have gone.

Simulate that. The item's removal is scheduled for 48 hours after its TTL timestamp, not for the timestamp itself, so there is a real window in which the clock has passed the TTL and a read still returns the item. That window is the useful part. A test can advance an hour past a session's expiry, see the session come back from `GetItem`, and find out that the code needs to cope with it. Then it can advance past the window and assert the item has gone.

48 hours goes in as a module constant, the way `minDays` in `SimSecretsManagerSecretDeletionSchedule` and `maximumVisibilityTimeoutSeconds` in SQS already do. No new configuration path. If a knob for it is ever wanted, the ACM precedent of a fluent mode method on the service (`requireDnsValidation()`, `autoIssueCertificates()`) is the shape to copy, and that can be a follow-up.

## Divergence from AWS

AWS promises a window, not an instant. Yulin has to pick one, and it picks the far end at 48 hours, because that maximises the period in which an expired item is still readable and that is the behaviour a user needs to defend against.

So a test can assert that an item survives its TTL timestamp, and that it has gone once the window has passed. It should not assert that an expired item is still present at some point inside the window, because real DynamoDB may well have deleted it by then. That belongs under Limitations.

## CloudFormation

`TimeToLiveSpecification` becomes a supported property on `AWS::DynamoDB::Table`, building on the best-effort CloudFormation model rather than the current whole-template refusal.

## Example

```typescript
const simAws = new SimAws({
  clock: new SimFixedClock(new Date("2026-08-01T09:00:00.000Z")),
});

await simAws.dynamoDb().updateTimeToLive(
  new UpdateTimeToLiveCommand({
    TableName: "Sessions",
    TimeToLiveSpecification: { Enabled: true, AttributeName: "expiresAt" },
  }),
);

// A session an hour out.
  new PutItemCommand({
    TableName: "Sessions",
    Item: {
      sessionId: { S: "abc" },
      expiresAt: { N: String(Math.floor(simAws.now().getTime() / 1000) + 3600) },
    },
  }),
);

await simAws.clock().advanceBy({ hours: 2 });

// Expired an hour ago, and still readable. This is what real DynamoDB does,
// and what application code has to allow for.
const stale = await simAws
  .dynamoDb()
  .getItem(new GetItemCommand({ TableName: "Sessions", Key: { sessionId: { S: "abc" } } }));
console.log(stale.Item !== undefined); // true

await simAws.clock().advanceBy({ days: 3 });

// Past the deletion window. Gone, with nothing else to call.
```

## Acceptance criteria

- [ ] `UpdateTimeToLive` moves the status `ENABLING` then `ENABLED`, and `DISABLING` then `DISABLED`, with `DescribeTimeToLive` reporting each
- [ ] A second `UpdateTimeToLive` within the hour is a `ValidationException`, and one that follows `advanceBy({ hours: 1 })` succeeds
- [ ] Advancing the clock past an item's TTL timestamp but not past the deletion window leaves the item readable
- [ ] Advancing past the deletion window deletes the item, with no further call from the test
- [ ] The deletion window is 48 hours measured from the TTL timestamp, not from the write or from the advance
- [ ] Items whose TTL attribute is absent, in the future, or not a Number survive any advance
- [ ] Overwriting an item with a later TTL before its window elapses keeps the item, and it goes 48 hours after the new timestamp
- [ ] Disabling time to live before an item's window elapses keeps that item
- [ ] Enabling time to live on a table that already holds items schedules their expiry
- [ ] An item written with a TTL already more than 48 hours in the past is still readable until the clock next moves
- [ ] A test with a `SimDynamoDb` and a `SimLambda` on one `SimAws` sees both respond to a single `advanceBy`, in due order
- [ ] A CDK table setting `timeToLiveAttribute` deploys, and `DescribeTimeToLive` reports that attribute
- [ ] Limitations in `docs/services/dynamodb/` says the window is a fixed 48 hours where AWS gives a range, so a test should not assert an expired item is still present partway through it
- [ ] `docs/time/` names time to live among what moving the clock causes, replacing the current "nothing schedules work on the clock yet" limitation

## Out of scope

Stream records for TTL deletions with `userIdentity` type `Service`. Removal from secondary indexes, since none exist yet.

Expiring on real time passing. Due work is dispatched by `simAws.clock()`, so an item whose window goes by while a running-mode clock tracks the host is not removed until something moves the clock. A simulated DynamoDB constructed standalone as `new SimDynamoDb()` has no clock control at all, so nothing there ever expires. Both follow from how time control works today rather than from anything here, and both belong under Limitations.

Making the 48 hours configurable.

Builds on #255.

hughgrigg:
UpdateTimeToLive and DescribeTimeToLive, with the status settling through ENABLING or DISABLING and the one update per table per hour rule measured on the simulated clock.

Items expire as the clock moves forward. The removal is scheduled 48 hours after the TTL timestamp rather than on it, so a test can advance past a session's expiry, find it still readable as it would be on AWS, and then advance past the window to see it go. Nothing else to call.

TimeToLiveSpecification becomes a supported AWS::DynamoDB::Table property.

Resolves https://github.com/KensioSoftware/yulin/issues/263

## Summary by CodeRabbit

- **New Features**
  - Added DynamoDB time-to-live support for enabling, disabling, and describing TTL settings.
  - Expired items now use simulated time and are deleted after the documented 48-hour window.
  - Added CloudFormation and SDK support for configuring and querying TTL.

- **Documentation**
  - Added usage examples, configuration guidance, lifecycle details, and documented limitations.

- **Tests**
  - Added coverage for TTL validation, status transitions, throttling, expiration behavior, and integrations.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -1350,7 +1350,8 @@ altogether. Here the answers stay lined up with the Gets that asked for them.
 
 `UpdateTimeToLive` names the attribute a table expires items by, and `DescribeTimeToLive` reports
 it. The attribute holds epoch seconds in a Number. An item without it, or holding a String or
-anything else, never expires, and that is not an error.
+anything else, never expires, and that is not an error. Nor does an item whose timestamp is more
+than five years in the past, which DynamoDB treats as a malformed value rather than as long overdue.
 
 Expiry runs on [the simulated clock](../../time/). Moving the clock forward is what deletes items
 whose time to live has run out, so one `advanceBy` expires a table's sessions alongside whatever
@@ -1446,8 +1447,12 @@ console.log(collected.Item === undefined); // true
 work has run, which is the sequence a table's own status goes through. Switching it off goes through
 `DISABLING` to `DISABLED`, and a `DISABLED` table reports no attribute name.
 
-DynamoDB takes one `UpdateTimeToLive` per table per hour. That hour is measured on the simulated
-clock, so a second call inside it is a `ValidationException` and
+An `UpdateTimeToLive` asking for the state the table is already in is a `ValidationException`, as it
+is on AWS, so code that has to be idempotent reads `DescribeTimeToLive` first. Changing the
+attribute an enabled table expires by means switching time to live off and then on again.
+
+DynamoDB also takes one `UpdateTimeToLive` per table per hour. That hour is measured on the
+simulated clock, so a second call inside it is a `ValidationException` and
 `simAws.clock().advanceBy({ hours: 1 })` is what lets the next one through.
 
 Switching time to live on reaches the items already on the table, since their attributes were only
@@ -1757,9 +1762,9 @@ nothing is written.
 - Time to live deletions publish no stream records. Real DynamoDB writes one with a `userIdentity`
   of type `Service`, which is how an application tells a TTL deletion from an application's own, and
   streams are not simulated here.
-- `UpdateTimeToLive` does not refuse switching time to live to the state it is already in. Real
-  DynamoDB answers that with a `ValidationException`. The one update per hour rule catches a repeat
-  inside the hour, so what is accepted here is a repeat an hour or more later.
+- The five year time to live eligibility rule counts 1825 days rather than five calendar years, so
+  an item whose timestamp sits within a couple of days of the boundary may be treated differently
+  here to how AWS treats it.
 - DynamoDB streams are not simulated. A `StreamSpecification` with `StreamEnabled` set is refused,
   so a table whose changes nothing is publishing cannot be created by accident. One that switches
   streams off describes the table this simulation already makes, so it is accepted.
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -1346,6 +1346,115 @@ console.log(output.Responses[1]); // {}
 That is the difference from `BatchGetItem`, which leaves a missing item out of its answer
 altogether. Here the answers stay lined up with the Gets that asked for them.
 
+## Expiring items with time to live
+
+`UpdateTimeToLive` names the attribute a table expires items by, and `DescribeTimeToLive` reports
+it. The attribute holds epoch seconds in a Number. An item without it, or holding a String or
+anything else, never expires, and that is not an error.
+
+Expiry runs on [the simulated clock](../../time/). Moving the clock forward is what deletes items
+whose time to live has run out, so one `advanceBy` expires a table's sessions alongside whatever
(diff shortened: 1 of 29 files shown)


hughgrigg:
@coderabbitai full review

PR #288, opened 2026-08-01
feat: use the DynamoDB document client against simulated DynamoDB

### LINKED ISSUE #265: Use the DynamoDB document client against simulated DynamoDB
## Problem

Most application code that talks to DynamoDB uses `@aws-sdk/lib-dynamodb`, so it writes `{ id: "a", count: 1 }` rather than `{ id: { S: "a" }, count: { N: "1" } }`. Yulin intercepts the base `DynamoDBClient` only, so a test of that code cannot use the simulator without rewriting the code under test to use attribute values.

## Current behaviour

Two independent things stop it, and fixing either one alone changes nothing.

`SimSdkCommandDispatcher.dispatch` identifies a Command by `command.constructor.name` and looks that name up in the service router. Document client Commands are named `PutCommand`, `GetCommand` and so on, while `SimDynamoDatabaseSdkCommandRouter` knows `CreateTableCommand`, `DescribeTableCommand`, `ListTablesCommand` and `PutItemCommand`, so a `PutCommand` throws `SimSdkUnsupportedCommandError`.

The second is where marshalling happens. `@smithy/core`'s `Client.send` is what calls `command.resolveMiddleware(this.middlewareStack, this.config, options)` and runs the resulting handler, and `installSendPatch` replaces `send`, so no middleware runs at all. `@aws-sdk/lib-dynamodb` is not a dependency of this repo yet and is not installed, so its source could not be read: the assumption is that its marshalling is middleware reached through `resolveMiddleware`, which an intercepted send never gets to. Confirm that when the devDependency is added, and if the marshalling turns out to sit somewhere else, say so and change this issue rather than working around it.

## Desired behaviour

A document client Command sent through an interception behaves like the equivalent attribute value Command. Input is marshalled from native JavaScript at the interception boundary and output is unmarshalled back, using the option defaults lib-dynamodb sets rather than the `util-dynamodb` defaults, so the same code produces the same values against the simulator and against AWS.

## Example

```typescript
using simSdk = new SimSdk();
const documents = DynamoDBDocumentClient.from(new DynamoDBClient({ region: "eu-west-2" }));
simSdk.intercept(documents);

await documents.send(
  new PutCommand({ TableName: "FoobarTable", Item: { id: "a", count: 1 } }),
);
const read = await documents.send(
  new GetCommand({ TableName: "FoobarTable", Key: { id: "a" } }),
);
```

## Proposed scope

Routes for `PutCommand`, `GetCommand`, `DeleteCommand`, `UpdateCommand`, `BatchWriteCommand` and `BatchGetCommand`, each one a route plus a marshalling entry, with the conversion in its own collaborator rather than inlined into the router. `@aws-sdk/lib-dynamodb` as a devDependency. Docs in the SDK page, whose Limitations section already says only `client.send` is intercepted, and in the DynamoDB page, listing which native types map to which descriptors.

`DynamoDBDocumentClient.from(client)` builds a separate object, so an instance interception on the base client does not cover it. Either patch the document client as well, or settle on which object a user intercepts and document that. Worth checking early: the dispatcher reads the service from `client.config.serviceId`, and whether the document client exposes the base client's config, and so reports `DynamoDB`, is an assumption rather than something verified.

## Acceptance criteria

- [ ] A `PutCommand` and a `GetCommand` round trip through an interception with native values going in and native values coming back
- [ ] Marshalling uses lib-dynamodb's own option defaults, pinned by a test covering a case where they differ from the `util-dynamodb` defaults
- [ ] Which object a user intercepts is settled, tested and documented
- [ ] A document client Command for an operation the simulator does not support yet throws `SimSdkUnsupportedCommandError` naming it, rather than failing somewhere inside marshalling
- [ ] Number handling on the round trip matches lib-dynamodb's default, and any precision it drops is stated in the docs rather than left to be discovered
- [ ] Both the SDK page and the DynamoDB page gain a section with an extractable example

## Out of scope

`Query`, `Scan`, transaction and PartiQL document client Commands, each of which arrives with the issue that adds the underlying operation. The served HTTP path, where the document client marshals before the request is built and so needs nothing from the simulator.

Builds on #259 and #260.

hughgrigg:
Routes lib-dynamodb's PutCommand, GetCommand, DeleteCommand, UpdateCommand, BatchWriteCommand and BatchGetCommand, converting native JavaScript values on the way in and out.

The real document client converts in middleware reached through resolveMiddleware, which an intercepted send replaces, so the conversion happens at the interception boundary instead. It uses lib-dynamodb's own option defaults rather than the util-dynamodb ones.

Intercept the document client itself: DynamoDBDocumentClient.from builds a separate object that does not extend DynamoDBClient, so intercepting the base client does not reach it.

Resolves https://github.com/KensioSoftware/yulin/issues/265

## Summary by CodeRabbit

* **New Features**
  * Added simulated DynamoDB document-client support for native JavaScript values.
  * Supports document operations including reads, writes, updates, deletes, batch requests, nested maps and lists, sets, binary values, and bigints.
  * Added conversion and validation for numbers, including safe handling of large integers.
  * Added clear errors for unsupported or invalid document values.
* **Documentation**
  * Added setup guides and examples covering document-client interception and common workflows.
* **Tests**
  * Added comprehensive coverage for conversion, routing, authorization, shared table state, and supported commands.

### DIFF
--- a/docs/sdk/README.md
+++ b/docs/sdk/README.md
@@ -170,6 +170,76 @@ intercept only specific Commands, pass an allow list of Command classes or names
 `simSdk.intercept(s3Client, { commands: [GetObjectCommand] })`. Commands outside the allow list
 throw a diagnostic error.
 
+## The DynamoDB document client
+
+`@aws-sdk/lib-dynamodb` lets application code write `{ id: "a", count: 1 }` rather than
+`{ id: { S: "a" }, count: { N: "1" } }`. Intercept the document client and its Commands reach
+simulated DynamoDB with the values converted, so code written against it needs no changes to run
+against the simulator.
+
+```typescript sim-sdk-document-client
+/**
+ * An intercepted DynamoDB document client, writing plain JavaScript values.
+ */
+
+import { CreateTableCommand, DynamoDBClient } from "@aws-sdk/client-dynamodb";
+import {
+  DynamoDBDocumentClient,
+  GetCommand,
+  PutCommand,
+} from "@aws-sdk/lib-dynamodb";
+
+import { SimSdk } from "@kensio/yulin/sdk";
+
+using simSdk = new SimSdk();
+
+const documents = DynamoDBDocumentClient.from(
+  new DynamoDBClient({ region: "eu-west-2" }),
+);
+
+// The document client is what gets intercepted, not the client it was built
+// from.
+simSdk.intercept(documents);
+
+// A document client forwards a Command it has no document form of, so the
+// table is created through the same client.
+await documents.send(
+  new CreateTableCommand({
+    TableName: "OrdersTable",
+    KeySchema: [{ AttributeName: "orderId", KeyType: "HASH" }],
+    AttributeDefinitions: [{ AttributeName: "orderId", AttributeType: "S" }],
+    BillingMode: "PAY_PER_REQUEST",
+  }),
+);
+await simSdk.simAws.backgroundTasksComplete();
+
+await documents.send(
+  new PutCommand({
+    TableName: "OrdersTable",
+    Item: { orderId: "order-1", total: 42, paid: true },
+  }),
+);
+
+const read = await documents.send(
+  new GetCommand({ TableName: "OrdersTable", Key: { orderId: "order-1" } }),
+);
+
+console.log(read.Item?.["total"]); // 42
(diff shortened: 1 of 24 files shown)


hughgrigg:
@coderabbitai full review

PR #289, opened 2026-08-01
feat: query a simulated DynamoDB table by partition key and sort key

### LINKED ISSUE #266: Query a simulated DynamoDB table by partition key and sort key
## Problem

A table with a sort key exists so an item collection can be read back in order, and `Query` is the operation that does it. Without it a test can write a collection and then only fetch items one known key at a time, which is not how the code under test reads them. Simulated DynamoDB has no `Query`: its SDK Command router lists `CreateTableCommand`, `DescribeTableCommand`, `ListTablesCommand` and `PutItemCommand`, and nothing else.

## Desired behaviour

`KeyConditionExpression` is parsed against a closed grammar, which is deliberately not the general condition grammar: exactly one equality test on the partition key, optionally combined with `AND` and exactly one sort key condition drawn from `=`, `<`, `<=`, `>`, `>=`, `BETWEEN` and `begins_with`. Matching items come back in sort key order, and `ScanIndexForward: false` reads the collection backwards. The response carries `Items`, `Count` and `ScannedCount`.

Anything outside that grammar is a `ValidationException` naming what was wrong: no partition key equality, a range operator applied to the partition key, an unsupported operator or function on the sort key, a non-key attribute anywhere in the condition, `OR` or `NOT` anywhere, or `begins_with` against a Number sort key.

Pagination comes with it, because a query without it is not the operation applications use. `Limit` counts items evaluated rather than items returned. `LastEvaluatedKey` is present whenever the walk stopped at a `Limit` boundary, and absent only when the key range ran out within budget, so hitting `Limit` on the last matching item still returns a token and the next call returns an empty page. `ExclusiveStartKey` resumes exclusively.

## Example

```typescript
const page = await simAws.dynamoDb().query(
  new QueryCommand({
    TableName: "FoobarTable",
    KeyConditionExpression: "customerId = :customer AND begins_with(orderId, :prefix)",
    ExpressionAttributeValues: {
      ":customer": { S: "c-1" },
      ":prefix": { S: "2026-" },
    },
    Limit: 10,
  }),
);
```

## Proposed scope

A `command/query/` group with the key condition parser as its own collaborator, and a pagination collaborator written so `Scan` can reuse it unchanged later. Sort key ordering needs comparison rules that belong with the item model rather than in the query handler: numeric ordering for `N`, UTF-8 byte order for `S`, unsigned byte order for `B`.

`IndexName`, `FilterExpression`, `ProjectionExpression` and `Select` are refused by name, in the style of `refuseUnsimulatedQueueInput` in simulated SQS, so a query that would mean something different in AWS fails here instead of quietly returning more than it should.

## Acceptance criteria

- [ ] A partition key equality on its own returns the whole item collection in ascending sort key order, and `ScanIndexForward: false` reverses it
- [ ] All seven sort key conditions are covered, including `begins_with` refused against a Number sort key
- [ ] Each rejected key condition shape gives a `ValidationException` that names the problem
- [ ] `Limit` counts evaluated items, and reaching `Limit` on the final matching item still returns a `LastEvaluatedKey` and then an empty page
- [ ] Paging a collection with `ExclusiveStartKey` returns every item exactly once
- [ ] `IndexName`, `FilterExpression`, `ProjectionExpression` and `Select` are each refused by name
- [ ] Docs section on querying an item collection, with the 1 MB page cap under Limitations as unmodelled, so pages break only on `Limit`

## Out of scope

Filter expressions, projection expressions and `Select`, each with its own issue. `Scan`. Secondary indexes, so `IndexName` stays refused. Consumed capacity.

## Implementation notes

The key condition grammar stays separate from the general condition expression grammar. Sharing one parser would let a query accept `OR`, which real DynamoDB rejects in a key condition, and that is the kind of divergence that passes here and fails on deploy.

Items are currently held in a `Map` keyed by a JSON string built by `DynamoDbKeySchema.makeItemKey`, which carries no ordering, so a collection has to be grouped by partition key and ordered at read time. The ordering has to compare decimal values and bytes: `list-tables.handler.ts:68` sorts table names with `localeCompare`, which is the wrong comparison for DynamoDB, and in `Query` that mistake would be far more visible.

Builds on #256.

hughgrigg:
Query reads one item collection: the items under one partition key, in sort key order. KeyConditionExpression is a closed grammar of its own rather than the general condition grammar, so a query cannot accept OR the way a condition can.

Limit counts items evaluated, LastEvaluatedKey is present whenever the walk stopped at that limit, and ExclusiveStartKey resumes exclusively by comparing sort keys.

IndexName, FilterExpression, ProjectionExpression and Select are refused by name. The document client's QueryCommand is refused too, since it shares its name with the client Command the interception routes by.

Resolves https://github.com/KensioSoftware/yulin/issues/266

## Summary by CodeRabbit

* **New Features**
  * Added support for DynamoDB `Query` operations, including partition and sort-key conditions, ascending or descending results, counts, and pagination.
  * Added support for continuation keys to resume queries and validate query boundaries.
  * Added SDK client routing and authorization coverage for Query requests.
* **Documentation**
  * Added Query usage examples, paging examples, supported functionality, syntax guidance, and limitations.
* **Bug Fixes**
  * Improved validation and handling of unsupported query options and document-client Query requests.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -1008,6 +1008,9 @@ The grammar is closed, and deliberately narrower than a `ConditionExpression`. E
 - the same key attribute tested twice
 - a `BETWEEN` whose upper bound is below its lower bound, or whose bounds are different types
 - a value written into the expression rather than supplied through `ExpressionAttributeValues`
+- a value whose type is not the one the table declared for that key attribute, such as comparing an
+  `S` sort key against an `N`. A key attribute has one type, so the condition could never hold, and
+  an empty page would read as a collection that happens to hold nothing.
 
 An attribute name that is a DynamoDB reserved word is written as a `#name` placeholder and defined in
 `ExpressionAttributeNames`, as in any other expression.
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -903,6 +903,192 @@ wanted. Naming one path twice counts the same way.
 A document path goes at most 32 levels deep, which is as far as an item nests. A negative index, a
 fractional index and a path past that depth are each a `ValidationException` naming the path.
 
+## Querying an item collection
+
+A table with a sort key holds an item collection under each partition key: the items with that
+partition key, ordered by their sort key. `Query` reads one of those collections.
+
+`KeyConditionExpression` says which. It is one equality on the partition key, optionally joined by
+`AND` to one condition on the sort key. The sort key condition is `=`, `<`, `<=`, `>`, `>=`,
+`BETWEEN` or `begins_with`, and both bounds of a `BETWEEN` are inside the range.
+
+```typescript sim-dynamodb-query
+/**
+ * Reading a customer's orders back in sort key order.
+ */
+
+import {
+  CreateTableCommand,
+  PutItemCommand,
+  QueryCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
+
+await dynamoDb.createTable(
+  new CreateTableCommand({
+    TableName: "OrdersTable",
+    KeySchema: [
+      { AttributeName: "customerId", KeyType: "HASH" },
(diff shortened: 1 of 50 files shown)


hughgrigg:
@coderabbitai full review

PR #290, opened 2026-08-01
feat: scan a simulated DynamoDB table, including parallel segments

### LINKED ISSUE #267: Scan a simulated DynamoDB table, including parallel segments
## Problem

Simulated DynamoDB has no way to read every item in a table. Query answers questions about one partition key at a time, so a test that does not know the keys, or that wants to assert what a table holds after some other component wrote to it, has nothing to call. Scan is also the operation most test setup and assertion code reaches for, because it needs no key knowledge.

Parallel scan matters for a different reason. Segment and TotalSegments change what the caller has to get right rather than what the simulator has to do, since there is no concurrency benefit in an in-process simulator. The value is that code splitting a scan across segments can be exercised without a real table.

## Desired behaviour

Scan walks every item in the table and returns Items, Count and ScannedCount, reusing the pagination collaborator added with Limit and LastEvaluatedKey on Query. ConsistentRead is accepted, and the simulator is always strongly consistent.

Ordering follows AWS rather than being convenient. Partition key values come back in an arbitrary order that is deliberately not sorted, and items within a single partition key value are ascending by sort key. A test that depended on a global sort would pass here and fail against the real service, so the simulator should not offer one.

Segment and TotalSegments must be supplied together. Segment is zero based and must be less than TotalSegments, TotalSegments is between 1 and 1000000, and a TotalSegments of 1 is simply a sequential scan. Items are assigned to a segment by hashing the partition key value, so every item sharing a partition key lands in the same segment, and segments may be uneven or empty.

## Example

```typescript
const segment = await simAws.dynamoDb().scan(
  new ScanCommand({
    TableName: "FoobarTable",
    Segment: 0,
    TotalSegments: 4,
  }),
);
```

## Proposed scope

A `command/scan/` group alongside `command/query/`, sharing the pagination collaborator and the sort key comparators rather than duplicating either. Segment assignment lives in its own small collaborator so the hashing rule can be tested on its own. Docs section in `docs/services/dynamodb/` with an extractable example covering a sequential scan and a worked parallel scan across four segments, including the standing advice that Scan is the wrong tool for most access patterns.

## Acceptance criteria

- [ ] Scan returns every item with Count and ScannedCount, paginating on Limit with LastEvaluatedKey and ExclusiveStartKey
- [ ] Items are ascending by sort key within one partition key value, and partition key values are not globally sorted
- [ ] Segment without TotalSegments, TotalSegments without Segment, a Segment at or above TotalSegments, and a TotalSegments outside 1 to 1000000 each give a ValidationException
- [ ] Every item sharing a partition key value lands in the same segment, and the segments together cover the whole table
- [ ] A paginated parallel scan resumes on the same segment, and an ExclusiveStartKey belonging to a different segment is rejected
- [ ] Segment and TotalSegments stay refused on Query, where they are not parameters at all

## Out of scope

FilterExpression, ProjectionExpression and Select on Scan, which stay refused until their own issues. Scanning a secondary index. The 1 MB page cap, which is not modelled, so pages break only on Limit. DynamoDB Streams and PartiQL, which stay out of this roadmap and belong under Limitations.

## Implementation notes

The arbitrary partition key order still has to be a stable total order within a process, otherwise ExclusiveStartKey cannot resume a scan. The requirement is that the order is not sorted, not that it varies between calls.

Builds on #266.

hughgrigg:
Scan walks the whole table, ordering partition key values by a hash so they come back arbitrarily rather than sorted, and items under one partition key ascending by sort key. Paging reuses the collaborator Query pages with.

Segment and TotalSegments divide a table by the same hash, so whole item collections move together. They are supplied together, and are refused on Query, which never had them.

Resolves https://github.com/KensioSoftware/yulin/issues/267

## Summary by CodeRabbit

* **New Features**
  * Added DynamoDB `ScanCommand` support, including full-table reads, deterministic ordering, pagination, continuation keys, and strong-read handling.
  * Added parallel scans using configurable segments.
  * Added native SDK routing and authorization for scan operations.
  * Added examples demonstrating standard and parallel scans.

* **Bug Fixes**
  * Query now explicitly rejects scan-only segment parameters.
  * Document-client scan commands are clearly reported as unsupported.

* **Documentation**
  * Updated DynamoDB feature lists, limitations, command references, and scan behavior documentation.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -1225,7 +1225,9 @@ for (const customerId of ["c-1", "c-2", "c-3", "c-4"]) {
 }
 
 const totalSegments = 4;
-const read: string[] = [];
+
+// Which segment each of a customer's orders came back in.
+const segmentsByCustomer = new Map<string, number[]>();
 
 for (let segment = 0; segment < totalSegments; segment++) {
   const segmentPage = await dynamoDb.scan(
@@ -1236,16 +1238,29 @@ for (let segment = 0; segment < totalSegments; segment++) {
     }),
   );
 
-  read.push(
-    ...(segmentPage.Items ?? []).map((item) => item["customerId"]?.S ?? ""),
-  );
+  const items = segmentPage.Items ?? [];
+
+  for (const item of items) {
+    const customerId = item["customerId"]?.S ?? "";
+    const segments = segmentsByCustomer.get(customerId) ?? [];
+
+    segmentsByCustomer.set(customerId, [...segments, segment]);
+  }
 }
 
 // The segments together are the whole table, with nothing read twice.
-console.log(read.length); // 8
+console.log(segmentsByCustomer.values().toArray().flat().length); // 8
+console.log(segmentsByCustomer.size); // 4
 
-// And each customer's orders arrived in one segment, both of them together.
-console.log(new Set(read).size); // 4
+// And each customer's two orders came back in one segment rather than split
+// between two.
+console.log(
+  segmentsByCustomer
+    .values()
+    .map((segments) => new Set(segments).size)
+    .toArray(),
+);
+// [ 1, 1, 1, 1 ]
 ```
 
 An item belongs to a segment by its partition key value, so every item of one item collection lands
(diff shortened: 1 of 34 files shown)


hughgrigg:
@coderabbitai full review

PR #291, opened 2026-08-01
feat: filter and count simulated DynamoDB query and scan results

### LINKED ISSUE #268: Filter and count simulated DynamoDB query and scan results
## Problem

Query and Scan return whole item collections, so a test covering an application that asks for the open orders belonging to a customer has to do the filtering itself. That works until the test asserts on Count or on pagination, at which point the difference between what DynamoDB evaluated and what it returned starts to matter.

Filtering is also the part of DynamoDB most often simulated wrongly, because it happens after the read. A filter saves no capacity, does not let a Limit fill up, and can leave a page with no items on it. A simulator that filters before applying Limit hands tests a page size the real service would never produce.

## Desired behaviour

The order of operations defines the semantics. Evaluate the key condition or walk the scan, cut the page at Limit, record ScannedCount, then apply the filter, then take Count from the survivors. A page can return zero Items with a LastEvaluatedKey present, because everything on it was filtered out.

The two commands are asymmetric, and this is the rule worth getting right. A Query FilterExpression may not reference the partition key or sort key attribute and is rejected with a ValidationException. A Scan FilterExpression may reference any attribute, including key attributes.

Select carries its own validation matrix. COUNT returns Count and ScannedCount with no Items array. SPECIFIC_ATTRIBUTES requires a ProjectionExpression. A ProjectionExpression combined with any Select other than SPECIFIC_ATTRIBUTES is a ValidationException. ALL_PROJECTED_ATTRIBUTES is valid only with an IndexName. The default for a table read is ALL_ATTRIBUTES.

## Example

```typescript
const page = await simAws.dynamoDb().query(
  new QueryCommand({
    TableName: "FoobarTable",
    KeyConditionExpression: "customerId = :customerId",
    FilterExpression: "#status = :open",
    ExpressionAttributeNames: { "#status": "status" },
    ExpressionAttributeValues: {
      ":customerId": { S: "c-1" },
      ":open": { S: "OPEN" },
    },
    Limit: 10,
  }),
);
```

`ScannedCount` is up to 10 and `Count` is how many of those survived the filter, so a Count below the Limit does not mean the collection is exhausted.

## Proposed scope

FilterExpression on Query and Scan, reusing the condition expression parser and evaluator built for ConditionExpression rather than writing a second grammar, plus the Select parameter on both commands. Lift the FilterExpression and Select refusals added by the Query and Scan issues, each by name so none is left behind. Docs section explaining that filtering happens after the read.

## Acceptance criteria

- [ ] FilterExpression is applied after Limit, so ScannedCount counts items evaluated and Count counts survivors
- [ ] A Query FilterExpression referencing a partition or sort key attribute gives a ValidationException, and the same expression on Scan is accepted
- [ ] A page whose items are all filtered out returns zero Items with a LastEvaluatedKey present
- [ ] Select COUNT returns Count and ScannedCount with no Items array
- [ ] Select SPECIFIC_ATTRIBUTES without a ProjectionExpression, and a ProjectionExpression with any other Select, each give a ValidationException
- [ ] ALL_PROJECTED_ATTRIBUTES without an IndexName gives a ValidationException

## Out of scope

Filtering on an index read, including whether a filter may reference index key attributes, which is settled once index reads land. Until then the filter fails closed against the table key schema and any index key schema. ProjectionExpression itself, which has its own issue.

## Implementation notes

The Select matrix can be enforced whether or not ProjectionExpression is simulated yet. If projections have not landed when this does, SPECIFIC_ATTRIBUTES stays unreachable through the existing ProjectionExpression refusal rather than being quietly accepted with nothing to project.

Builds on #257 and #267.

hughgrigg:
FilterExpression on Query and Scan, applied after the Limit so ScannedCount counts the items a read evaluated and Count the ones that survived. A page whose items were all dropped comes back empty with a LastEvaluatedKey.

A Query filter naming a key attribute is refused, and the same expression on a Scan is accepted. Select takes COUNT, and the rules tying SPECIFIC_ATTRIBUTES, ALL_PROJECTED_ATTRIBUTES and a projection together are enforced.

Resolves https://github.com/KensioSoftware/yulin/issues/268

## Summary by CodeRabbit

* **New Features**
  * Added `FilterExpression` support for DynamoDB queries and scans, including filtering, validation, pagination, and accurate scanned/result counts.
  * Added `Select` support, including count-only responses and supported projection modes.
  * Added filtering and selection behavior for parallel scans and missing attributes.
* **Documentation**
  * Expanded DynamoDB documentation and examples covering filters, selections, projections, pagination, validation, and response semantics.
* **Tests**
  * Added comprehensive coverage for query, scan, filtering, selection, validation, and pagination scenarios.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -1289,6 +1289,157 @@ These are the rules a request is held to, each a `ValidationException`:
 them. A query reads one item collection, which sits under one partition key and so inside one
 segment.
 
+## Filtering a read
+
+`FilterExpression` drops items a `Query` or a `Scan` read. It is the same grammar a
+[conditional write](#conditional-writes) is guarded by, evaluated against each item the read
+reached.
+
+It runs after the read rather than during it, and that order is what the counts report. The walk is
+cut at the `Limit` first, and the filter then drops items from the page that came back.
+`ScannedCount` is how many items the read evaluated, and `Count` how many of those survived.
+
+```typescript sim-dynamodb-query-filter
+/**
+ * Reading a customer's open orders, and counting what that cost.
+ */
+
+import {
+  CreateTableCommand,
+  PutItemCommand,
+  QueryCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
+
+await dynamoDb.createTable(
+  new CreateTableCommand({
+    TableName: "OrdersTable",
+    KeySchema: [
+      { AttributeName: "customerId", KeyType: "HASH" },
+      { AttributeName: "orderId", KeyType: "RANGE" },
+    ],
+    AttributeDefinitions: [
+      { AttributeName: "customerId", AttributeType: "S" },
+      { AttributeName: "orderId", AttributeType: "S" },
+    ],
+    BillingMode: "PAY_PER_REQUEST",
+  }),
+);
+await simAws.backgroundTasksComplete();
+
+const orders = [
+  { orderId: "2026-01", status: "OPEN" },
+  { orderId: "2026-02", status: "SHIPPED" },
+  { orderId: "2026-03", status: "OPEN" },
+  { orderId: "2026-04", status: "SHIPPED" },
+];
+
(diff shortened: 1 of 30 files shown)


hughgrigg:
@coderabbitai full review

Issue #269, opened 2026-07-30
Create and query simulated DynamoDB global secondary indexes


hughgrigg:
## Problem

Most real DynamoDB tables carry at least one global secondary index, because the access patterns applications are written against depend on them. A simulated table can only be read on its own primary key, so a test covering code that queries a secondary access pattern cannot run against the simulator at all.

Index declaration is also where CreateTable input goes wrong in practice, and the rule that catches people is not obvious: AttributeDefinitions must match exactly, in both directions, across the table key schema and every index key schema. Declaring an index without adding its key attribute definitions is a ValidationException in real DynamoDB.

## Desired behaviour

A secondary index domain model, plus GlobalSecondaryIndexes on CreateTable. Index names are unique within the table. A key schema is one HASH element, or HASH then RANGE, with key attributes of S, N or B. At most 20 indexes per table. Projections are ALL, KEYS_ONLY or INCLUDE with 1 to 20 NonKeyAttributes, and INCLUDE without NonKeyAttributes is rejected, as is NonKeyAttributes without INCLUDE. Per-index ProvisionedThroughput is required under PROVISIONED and refused under PAY_PER_REQUEST.

Then IndexName on Query and Scan. An unknown index name gives ResourceNotFoundException. Key conditions validate against the index key schema rather than the table's. The index is sparse, so an item missing an index key attribute is absent from it rather than an error. Index key values need not be unique, so items sharing one come back in no particular order. ConsistentRead set to true against a global secondary index is a ValidationException. Requesting an attribute the index does not project is rejected rather than silently omitted, Select defaults to ALL_PROJECTED_ATTRIBUTES, and LastEvaluatedKey carries the index key attributes together with the table key attributes.

The one thing the write path still checks is that an index key attribute's value type matches its declared definition, which is a ValidationException on the write.

## Example

```typescript
await simAws.dynamoDb().createTable(
  new CreateTableCommand({
    TableName: "FoobarTable",
    KeySchema: [{ AttributeName: "pk", KeyType: "HASH" }],
    AttributeDefinitions: [
      { AttributeName: "pk", AttributeType: "S" },
      { AttributeName: "status", AttributeType: "S" },
    ],
    BillingMode: "PAY_PER_REQUEST",
    GlobalSecondaryIndexes: [
      {
        IndexName: "byStatus",
        KeySchema: [{ AttributeName: "status", KeyType: "HASH" }],
        Projection: { ProjectionType: "KEYS_ONLY" },
      },
[...]

PR #292, opened 2026-08-01
feat: create simulated DynamoDB global secondary indexes

hughgrigg:
Declare global secondary indexes on CreateTable, with index name, key schema, projection and per-index throughput validation. AttributeDefinitions is matched against the table key schema and every index key schema, in both directions. Each index is reported in the CreateTable and DescribeTable description with its ARN, key schema, projection and status.

An index is sparse, so an item missing an index key attribute is absent from the index rather than refused. An item carrying one as a type the index did not declare is refused on the write.

Reading an index is not included: IndexName on Query and Scan is still refused by name.

Part of https://github.com/KensioSoftware/yulin/issues/269

## Summary by CodeRabbit

* **New Features**
  * Added support for creating and describing DynamoDB global secondary indexes.
  * Added index key schemas, projections, provisioned throughput, lifecycle statuses, ARNs, and sparse-item behavior.
  * Added validation for index names, schemas, projections, attributes, throughput, and item key types.
  * Added an example demonstrating global secondary index creation and usage.

* **Limitations**
  * Reads using `IndexName` and local secondary indexes remain unsupported.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -157,7 +157,8 @@ console.log(index?.IndexArn); // ".../table/OrdersTable/index/byStatus"
 
 await simAws.backgroundTasksComplete();
 
-// An item with no status is simply absent from the index, rather than refused.
+// This order carries neither index key attribute, so it is absent from
+// byStatus rather than refused. Missing either one is enough.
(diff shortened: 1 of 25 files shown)


hughgrigg:
@coderabbitai full review

PR #293, opened 2026-08-01
feat: query and scan simulated DynamoDB global secondary indexes

### LINKED ISSUE #269: Create and query simulated DynamoDB global secondary indexes

hughgrigg:
IndexName on Query and Scan reads an index rather than the table, through a read view the table hands out. The index is derived when it is read, so it is sparse: an item missing any index key attribute is not in it.

Key conditions and filters are held to the index key schema, reads answer with the attributes the index projects, and Select defaults to ALL_PROJECTED_ATTRIBUTES. Asking for more than the index carries is refused rather than quietly answered with less.

Index keys are not unique, so LastEvaluatedKey carries the index key and the table key together and the table key breaks ties in the walk order. Without that, paging a collection whose entries share an index key would repeat an item or skip one.

An unknown IndexName gives ResourceNotFoundException, and ConsistentRead against a global secondary index is a ValidationException.

Resolves https://github.com/KensioSoftware/yulin/issues/269

## Summary by CodeRabbit

* **New Features**
  * Added support for querying and scanning global secondary indexes.
  * Added sparse-index behavior, projected attributes, index-specific pagination, and validation.
  * Added support for `IndexName` in read operations.
  * Added examples demonstrating indexed queries and projected results.

* **Bug Fixes**
  * Improved validation for invalid indexes, unsupported attributes, pagination keys, and consistency settings.

* **Documentation**
  * Expanded DynamoDB documentation with index usage, projection behavior, pagination, and limitations.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -2608,11 +2608,12 @@ nothing is written.
 - `Expected`, `ConditionalOperator` and `AttributeUpdates` are not converted for the document
   client, because simulated DynamoDB refuses all three anyway. A request carrying one is refused by
   the operation rather than by the conversion.
-- An index read answers with the attributes the index projects, and there is no way to ask for more.
-  Real DynamoDB fetches the rest from the table when a `ProjectionExpression` names an attribute an
-  index does not project, at the cost of a read per item. `ProjectionExpression` is not simulated on
-  `Query` or `Scan` at all, so the case does not arise, and `Select: ALL_ATTRIBUTES` is refused
-  rather than served that way.
+- An index read answers with the attributes the index projects, and nothing fills in the rest. Real
+  DynamoDB does not either: a global secondary index never reads the base table for an attribute it
+  does not project, which is why `Select: ALL_ATTRIBUTES` against a partial projection is refused
+  rather than served. Fetching from the base table is a local secondary index behaviour, and those
+  are not simulated. `ProjectionExpression` is not simulated on `Query` or `Scan`, so naming a
+  non-projected attribute that way does not arise.
 - Local secondary indexes are not simulated. `LocalSecondaryIndexes` is refused rather than dropped,
   since a table missing an index it was asked for would answer queries differently to the real one.
   An empty list asks for no index, so it is accepted.
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -1181,6 +1181,111 @@ A token still works when the item it names has since been deleted: it says where
 than which position to return to. A token from a different partition key is refused, since it names
 a collection this query is not reading.
 
+## Reading a global secondary index
+
+`IndexName` on `Query` and `Scan` reads an index instead of the table. The key condition is held to
+the index key schema rather than the table's, which is the point: the index is how an access pattern
+the table key cannot serve gets served.
+
+```typescript sim-dynamodb-query-index
+/**
+ * Querying a global secondary index.
+ */
+
+import {
+  CreateTableCommand,
+  PutItemCommand,
+  QueryCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
+
+await dynamoDb.createTable(
+  new CreateTableCommand({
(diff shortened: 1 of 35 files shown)


hughgrigg:
@coderabbitai full review

hughgrigg:
@coderabbitai full review

PR #294, opened 2026-08-02
feat: create and query simulated DynamoDB local secondary indexes

### LINKED ISSUE #270: Create and query simulated DynamoDB local secondary indexes
## Problem

A local secondary index gives an item collection a second sort key, and `CreateTable` is the only place one can be declared. AWS has no call that adds, changes or removes one afterwards. A simulated table created without its index therefore stays different from the deployed table for the whole of its life, and no later command can bring the two back together.

Simulated DynamoDB models neither index kind today, so an access pattern such as "this customer's orders in date order" cannot be expressed against a simulated table at all.

## Desired behaviour

`LocalSecondaryIndexes` on `CreateTable`, then `Query` and `Scan` against one.

Validation: at most 5 indexes, an index name matching the same pattern as a table name and unique across both index kinds, a partition key identical to the table's partition key, and exactly one sort key that is a scalar attribute other than the table sort key. Per-index `ProvisionedThroughput` is refused, since a local secondary index shares the table's throughput. Projections follow the same `ALL`, `KEYS_ONLY` and `INCLUDE` rules as global secondary indexes, with the base table sort key present in the index entry whatever the projection asks for.

Reads reuse the index read path built for global secondary indexes, with two differences that matter. `ConsistentRead: true` is accepted rather than rejected, because a local secondary index sits in the same partition as the item it indexes. An attribute the index does not project is fetched from the base table rather than refused, which is the fetch AWS charges the extra read capacity for.

`LastEvaluatedKey` for a local secondary index query carries three attributes: the table partition key, the index sort key and the table sort key.

## Example

```typescript
  new CreateTableCommand({
    TableName: "Orders",
    KeySchema: [
      { AttributeName: "customerId", KeyType: "HASH" },
      { AttributeName: "orderId", KeyType: "RANGE" },
    ],
    AttributeDefinitions: [
      { AttributeName: "customerId", AttributeType: "S" },
      { AttributeName: "orderId", AttributeType: "S" },
      { AttributeName: "placedAt", AttributeType: "S" },
    ],
    LocalSecondaryIndexes: [
      {
        IndexName: "OrdersByDate",
        KeySchema: [
          { AttributeName: "customerId", KeyType: "HASH" },
          { AttributeName: "placedAt", KeyType: "RANGE" },
        ],
        Projection: { ProjectionType: "KEYS_ONLY" },
      },
    ],
  }),
);
```

A `Query` on `OrdersByDate` with `ConsistentRead: true` then returns that customer's orders ordered by `placedAt`.

## Acceptance criteria

- [ ] A partition key differing from the table's, a sort key equal to the table sort key, a non-scalar sort key attribute or a per-index `ProvisionedThroughput` is refused with a `ValidationException` naming the index
- [ ] More than 5 local secondary indexes is refused, and an index name colliding with a global secondary index name on the same table is refused
- [ ] `Query` and `Scan` accept `IndexName` for a local secondary index, and `ConsistentRead: true` succeeds rather than raising
- [ ] An attribute outside the index projection is returned by fetching it from the base table
- [ ] `LastEvaluatedKey` carries the table partition key, the index sort key and the table sort key, and passing it back as `ExclusiveStartKey` resumes the walk
- [ ] A docs section with an extractable example covers declaring and querying the index, and states that a local secondary index cannot be added, changed or removed after the table is created

## Implementation notes

Index entries are derived at read time rather than maintained on every write, following the way `SimSqsQueue.applyLifecycle` applies retention and redrive when messages are looked at instead of scheduling them. Nothing in the write path then has to know an index exists, and an index cannot drift from the table it derives from.

The base table fetch is why this is not the global secondary index path with one flag flipped. A read against a local secondary index needs the base item to hand, so the index walk should yield keys and let the read resolve the item, rather than yielding already projected entries.

If the diff runs long, index declaration and validation makes a clean first pull request, with `Query` and `Scan` support following.

## Out of scope

The 10 GB item collection size limit and `ItemCollectionSizeLimitExceededException`, which are not modelled and belong under Limitations. CloudFormation index properties. Any change to an index after table creation, which AWS does not offer for this index kind either.

Builds on #269.

hughgrigg:
LocalSecondaryIndexes on CreateTable, with the key schema rules that make an index local, and Query and Scan reading one. A local secondary index answers a strongly consistent read and fetches an attribute it does not project from the base table, which is where a read of it differs from a read of a global secondary index.

Resolves https://github.com/KensioSoftware/yulin/issues/270

## Summary by CodeRabbit

* **New Features**
  * Added support for local secondary indexes when creating and describing DynamoDB tables.
  * Added querying and scanning through local secondary indexes, including sorting, filtering, pagination, and strongly consistent reads.
  * Added support for `KEYS_ONLY`, `INCLUDE`, and `ALL_ATTRIBUTES` projections, including fetching non-projected attributes from the base table.
  * Added validation for index keys, projections, naming, limits, and unsupported throughput settings.
  * Added documentation and a working local-index example.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -198,6 +198,141 @@ rather than refused on the write. An index keyed on two attributes needs both. T
 is held to on account of an index is the type: an item carrying an index key attribute as a type the
 index did not declare is a `ValidationException`, since the index could never hold it.
 
+## Local secondary indexes
+
+`LocalSecondaryIndexes` on `CreateTable` gives an item collection a second sort key. The index shares
+the table's partition key, so an entry sits in the same partition as the item it indexes, and its
+sort key is some other attribute. That is what serves an access pattern such as "this customer's
+orders in date order" against a table keyed by customer and order id.
+
+`CreateTable` is the only place one can be declared. AWS has no call that adds, changes or removes a
+local secondary index afterwards, so a table created without one stays without it for the whole of
+its life.
+
+```typescript sim-dynamodb-local-secondary-index
+/**
+ * Declaring and querying a local secondary index.
+ */
+
+import {
+  CreateTableCommand,
+  PutItemCommand,
+  QueryCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
+
+await dynamoDb.createTable(
+  new CreateTableCommand({
+    TableName: "Orders",
+    KeySchema: [
+      { AttributeName: "customerId", KeyType: "HASH" },
+      { AttributeName: "orderId", KeyType: "RANGE" },
+    ],
+    AttributeDefinitions: [
+      { AttributeName: "customerId", AttributeType: "S" },
+      { AttributeName: "orderId", AttributeType: "S" },
+      { AttributeName: "placedAt", AttributeType: "S" },
+    ],
+    BillingMode: "PAY_PER_REQUEST",
+    LocalSecondaryIndexes: [
+      {
+        IndexName: "OrdersByDate",
+        // The partition key is the table's own. The sort key is the whole of
+        // what the index adds.
+        KeySchema: [
+          { AttributeName: "customerId", KeyType: "HASH" },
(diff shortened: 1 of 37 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #295, opened 2026-08-02
feat: deploy simulated DynamoDB secondary indexes from CloudFormation

### LINKED ISSUE #271: Deploy simulated DynamoDB secondary indexes from CloudFormation
## Problem

CDK's `Table` construct has `addGlobalSecondaryIndex` and `addLocalSecondaryIndex`, so an indexed table is the ordinary shape of a real template rather than an unusual one. By this point simulated DynamoDB validates and models both index kinds through `CreateTable`, but the CloudFormation resource refuses the two properties that declare them, so a template that deploys against AWS gets a skipped resource where its table should be.

Today there is no DynamoDB CloudFormation support of any kind: `sim-cfn-service-resolver.iso.test.ts:117` asserts that an `AWS::DynamoDB::*` resource resolves to "Unsupported sim CloudFormation Resource service DynamoDB".

## Desired behaviour

`GlobalSecondaryIndexes` and `LocalSecondaryIndexes` on `AWS::DynamoDB::Table` are parsed into the same values the `CreateTable` path already validates, and the two property refusals added when the table resource landed are removed. No index validation is duplicated in the CloudFormation layer, so a template declaring an index whose key attributes are missing from `AttributeDefinitions` fails the resource with the error the API gives for the same input.

Those refusals were a stopgap rather than the end state. They exist so an unmodelled property cannot deploy as a table that quietly differs from the template, and simulated CloudFormation records such a resource as skipped rather than refusing the whole stack.

The test that shows the property parsing, the index model and the read path line up is a CDK `.loc.test.ts` synthesising a table with one global and one local secondary index, deploying it, writing items and querying both indexes.

## Acceptance criteria

- [ ] Both index properties are parsed into the values `CreateTable` validates, with no second validation path in the CloudFormation layer
- [ ] A template declaring an index whose key attributes are absent from `AttributeDefinitions` fails that resource with the same error the API gives
- [ ] A table carrying either property deploys rather than being recorded as skipped
- [ ] A CDK `.loc.test.ts` synthesises, deploys, writes and queries against both a global and a local secondary index
- [ ] The one-indexed-table-at-a-time rule is either enforced or documented under Limitations, with the template shape that diverges spelled out
- [ ] A docs section with an extractable example covers deploying an indexed table from a template

## Implementation notes

One real divergence has to be settled here. AWS allows only one table with secondary indexes to be in `CREATING` at a time within an account and region. Simulated CloudFormation creates each dependency-ready batch of resources concurrently in `SimCfnStackResourceBatchCreator.create`, so a template with two indexed tables and no `DependsOn` between them deploys here and fails there.

Two ways to answer it. Serialise creation of indexed tables within the account and region, which puts the rule in simulated DynamoDB where the index state already lives, at the cost of changing stack timing. Or let the template deploy and state the divergence loudly under Limitations, naming the exact template shape that behaves differently. Either way it wants a code comment saying which was chosen and why, since the next reader will otherwise assume the omission was an oversight.

## Out of scope

Index changes on a stack update, which follow the `UpdateTable` work. `AWS::DynamoDB::GlobalTable`. Every other unmodelled `AWS::DynamoDB::Table` property, which stays refused.

Builds on #255 and #270.

hughgrigg:
GlobalSecondaryIndexes and LocalSecondaryIndexes are read off an AWS::DynamoDB::Table Resource and handed to CreateTable, so an index a template declared is the index an SDK caller would have got. The two property refusals are gone, and the index entries are held to the same simulated-property rule the table is: an unmodelled setting on a global index skips the table naming it, and a property neither index kind has fails the Resource.

AWS creates one indexed table at a time in an account and region, which sim CloudFormation does not, so that divergence is documented under Limitations rather than serialised.

Resolves https://github.com/KensioSoftware/yulin/issues/271

## Summary by CodeRabbit

* **New Features**
  * Added support for deploying DynamoDB tables with global and local secondary indexes through CloudFormation and CDK.
  * Added querying, projections, and independently provisioned capacity for supported indexes.
* **Bug Fixes**
  * Improved validation and clearer handling of invalid or unsupported index configurations.
  * Preserved key-schema ordering and optional projection settings during deployment.
* **Documentation**
  * Added indexed-table examples, supported functionality, validation guidance, CDK usage, and concurrency limitations.
* **Tests**
  * Added coverage for deployment, validation, index queries, projections, and resource status behavior.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -2644,9 +2644,15 @@ console.log(stack.outputs.get("OrdersTableArn")?.value);
 ```
 
 The properties that are read are `TableName`, `KeySchema`, `AttributeDefinitions`, `BillingMode`,
-`ProvisionedThroughput`, `TableClass`, `DeletionProtectionEnabled`, `Tags`, `TimeToLiveSpecification`,
-`GlobalSecondaryIndexes` and `LocalSecondaryIndexes`. Each one is passed to `CreateTable` rather than
-applied here, so a value the template gets wrong fails the same way it would for an SDK caller.
+`ProvisionedThroughput`, `TableClass`, `DeletionProtectionEnabled`, `Tags`,
+`GlobalSecondaryIndexes`, `LocalSecondaryIndexes` and `TimeToLiveSpecification`. All but the last are
+passed to `CreateTable` rather than applied here, so a value the template gets wrong fails the same
+way it would for an SDK caller.
+
+`TimeToLiveSpecification` is applied after the table is created, through `UpdateTimeToLive`. Real
+`CreateTable` has no parameter for it either, so real CloudFormation makes the table and then
+updates it. A specification the template got wrong is refused in the words `UpdateTimeToLive` refuses
+it in.
 
 A table with no `TableName` is named after the stack and its logical ID, so the table above with its
 name left out would be `orders-stack-OrdersTable`. Real CloudFormation adds random characters to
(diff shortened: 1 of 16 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #296, opened 2026-08-02
feat: add and remove a global secondary index with simulated UpdateTable

### LINKED ISSUE #272: Add and remove a global secondary index with simulated UpdateTable
## Problem

`UpdateTable` is where a table's definition changes after it exists, and adding a global secondary index is the change most likely to go wrong in a deployment. AWS backfills the new index, and while that runs the index is visible in `DescribeTable` and cannot be read. A test written against a simulator that activates a new index instantly passes, and the same code against AWS queries an index that is not ready yet.

Simulated DynamoDB has no `UpdateTable` at all: `CreateTable`, `DescribeTable`, `ListTables` and `PutItem` are the whole surface. So nothing can currently cover a table that changes shape, including a table switched between billing modes.

## Desired behaviour

`UpdateTable` accepting exactly one operation per call: a provisioned throughput or billing mode change, or one global secondary index creation, or one global secondary index deletion. More than one in a single call is a `ValidationException`.

The table moves `ACTIVE` to `UPDATING` to `ACTIVE` through the background scheduler, the same way `CreateTable` already schedules the move out of `CREATING` at `create-table.handler.ts:87`. It stays readable and writable throughout, since AWS does not take a table offline to update it. A second `UpdateTable` while one is in flight gives `ResourceInUseException`.

A new index reports `IndexStatus` `CREATING` with `Backfilling` true and then `ACTIVE`, and cannot be queried or scanned until it is `ACTIVE`. Its key attributes must be present in `AttributeDefinitions` on the same call, which is the only chance to declare them. Deleting an index takes it out of `DescribeTable`, after which reads against it give `ResourceNotFoundException`.

`DeletionProtectionEnabled` and `TableClass` changes flow through the same command and are covered here.

Which error a read against a backfilling index gets should be checked against real AWS before a test pins it, rather than assumed.

## Acceptance criteria

- [ ] A call combining a throughput change with a `GlobalSecondaryIndexUpdates` entry, or combining two index updates, is refused with a `ValidationException`
- [ ] Creating an index whose key attributes are absent from the `AttributeDefinitions` supplied on the same call is refused
- [ ] The table reports `UPDATING`, serves reads and writes while it does, and returns to `ACTIVE` through the background scheduler
- [ ] A second `UpdateTable` while one is in flight gives `ResourceInUseException`
- [ ] A new index reports `Backfilling` true while `CREATING`, cannot be queried or scanned until `ACTIVE`, and then answers for items written before it was added
- [ ] Deleting an index removes it from `DescribeTable`, and queries against the removed name give `ResourceNotFoundException`
- [ ] Stream updates, replica updates, `OnDemandThroughput` and local secondary index changes are each refused by name, and a docs section with an extractable example covers adding an index to a live table

## Implementation notes

Because index entries are derived at read time, there is no backfill to run. The `CREATING` and `Backfilling` window is a status the scheduler advances rather than work being done, and the index answers for older items the moment it goes `ACTIVE`. The observable behaviour matches AWS while the mechanism does not, so that wants a code comment where the status transition is scheduled.

`SimDynamoDbTable` is currently constructed from a whole `SimCreateTableCommand` (`sim-dynamodb-table.ts:42`) and stores no attribute definitions, so `UpdateTable` depends on the plain-values table model the `CreateTable` rebuild introduces before it can change one in place.

## Out of scope

Enabling and disabling streams, including the `ValidationException` for enabling a stream that already exists. Global table replicas. Auto scaling settings. Local secondary index changes, which AWS does not allow either.

Builds on #269.

hughgrigg:
UpdateTable does one thing per call, as AWS does: change what the table is billed and provisioned as, add one global secondary index, or remove one. A table class or deletion protection change rides along with any of them. The table moves through UPDATING while still serving reads and writes, and a second update in flight gives ResourceInUseException.

A global secondary index now has a status of its own, so a new one reports Backfilling and refuses reads until it is ACTIVE while the indexes already on the table stay readable.

Making a table updatable took SimDynamoDbTable over the line limit, so its mutable definition and its time to live are each held by a collaborator of their own.

Resolves https://github.com/KensioSoftware/yulin/issues/272

## Summary by CodeRabbit

* **New Features**
  * Added support for updating simulated DynamoDB tables, including billing settings, table class, deletion protection, and global secondary index creation or deletion.
  * Added realistic table and index lifecycle states, including `UPDATING`, backfilling, activation, and read restrictions while indexes are building.
  * Added SDK routing and IAM authorization support for table updates.
* **Documentation**
  * Documented supported operations, limitations, validation rules, and index lifecycle behavior.
  * Added an executable example demonstrating index creation, querying, and removal.
* **Tests**
  * Expanded coverage for validation, authorization, billing changes, lifecycle transitions, and unsupported settings.


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #306, opened 2026-08-02
feat: route a request to a simulated Lambda through a simulated HTTP API

### LINKED ISSUE #298: Route a request to a simulated Lambda through a simulated HTTP API
## Problem

A Node.js service that runs behind an API Gateway HTTP API cannot be tested against Yulin. There is no `src/service/apigatewayv2/`. The serve layer knows nothing about `execute-api` hostnames. `resolveSimSdkCommandRouter` has no case for the API Gateway v2 client, so intercepting an `ApiGatewayV2Client` throws `SimSdkUnknownServiceError`.

The only way to reach a simulated handler today is `InvokeCommand`. There is no API resource model, and no event built from a real HTTP request.

The payload format 2.0 event that an HTTP API sends to a Lambda proxy integration is already built in `src/service/lambda/serve/event/`, for Function URLs. The type's own comment names it as the API Gateway HTTP API payload format 2.0. It is Function-URL shaped rather than generic. `routeKey` and `stage` are the literal type `"$default"` in `sim-lambda-url-event.type.ts`. The `apiId`, `domainName` and `domainPrefix` values are read off `SimLambdaFunctionUrl`. `pathParameters` and `stageVariables` are absent.

## Desired behaviour

A `SimApiGatewayV2` facade reachable as `simAws.apiGatewayV2()`, scoped to an account and region. Create an API, an `AWS_PROXY` integration pointing at a simulated Lambda function, a `$default` route and a `$default` stage. Then fetch `https://<api-id>.execute-api.<region>.amazonaws.com/` through `serveSimAws` and get the handler's response back.

The handler receives a payload format 2.0 event. Its result is turned back into an HTTP response with real 2.0 semantics. That includes the documented inference: a handler returning `{ body: "hi" }` with no top-level `statusCode` produces a 200 whose body is `{"body":"hi"}` with `content-type: application/json`.

## Example

```typescript
const { ApiId, ApiEndpoint } = await simAws.apiGatewayV2().createApi(
  new CreateApiCommand({ Name: "orders", ProtocolType: "HTTP" }),
);
// ApiEndpoint === "https://<api-id>.execute-api.eu-west-2.amazonaws.com"

const { IntegrationId } = await simAws.apiGatewayV2().createIntegration(
  new CreateIntegrationCommand({
    ApiId,
    IntegrationType: "AWS_PROXY",
    IntegrationUri: functionArn,
    PayloadFormatVersion: "2.0",
  }),
);

  new CreateRouteCommand({
    ApiId,
    RouteKey: "$default",
    Target: `integrations/${IntegrationId}`,
  }),
);

await simAws.apiGatewayV2().createStage(
  new CreateStageCommand({ ApiId, StageName: "$default", AutoDeploy: true }),
);

const response = await fetch(srv.localUrl(`${ApiEndpoint}/orders?limit=10`));
```

`srv.localUrl` maps that to `http://<api-id>.execute-api.eu-west-2.sim-aws.localhost:<port>/orders?limit=10`, the same rewrite the Function URL host already gets.

## Proposed scope

- `CreateApi`, `GetApi`, `GetApis`, `DeleteApi`, `CreateIntegration`, `GetIntegrations`, `CreateRoute`, `GetRoutes`, `CreateStage`, `GetStages`. No `Update*` commands, no per-resource `Delete*`, no paging.
- `SimApiGatewayV2Api` as the aggregate root owning the route, integration and stage stores, since every child resource is `ApiId`-scoped on AWS. The facade delegates to per-resource command classes rather than holding the logic itself.
- API ids are 10 lowercase alphanumerics, allocated and registered together in one factory method, with deregistration on `DeleteApi`. A registry maps api id to owning account so a served hostname can find the account that owns the API.
- `CreateApi` accepts `Name`, `ProtocolType: "HTTP"`, `Description` and `DisableExecuteApiEndpoint`. It refuses `CorsConfiguration` and `Tags` by name, so the CloudFormation issue's property allow-list refuses them too rather than accepting them with no behaviour behind them.
- `DisableExecuteApiEndpoint: true` stops the generated endpoint serving, with 403 `{"message":"Forbidden"}`. AWS does not publish that status or body, so it is observed rather than documented and is recorded as such.
- Serve-layer wiring so `<api-id>.execute-api.<region>.amazonaws.com` reaches a new `SimApiGatewayV2ServiceController`, with the SigV4 signing name `execute-api`.
- Extraction of the payload 2.0 machinery out of `src/service/lambda/serve/` so both services share it, parameterising `routeKey`, `stage` and the endpoint identity, and adding optional `pathParameters` and `stageVariables`.
- Widening the shared event's `requestContext.authorizer` field to a union with optional `iam` and `jwt` members, so the JWT and `AWS_IAM` issues add an implementation rather than reopening a public type.
- Event omission rules matching AWS: `cookies`, `queryStringParameters`, `body`, `pathParameters` and `stageVariables` are omitted rather than null when empty, while `rawQueryString` is always present as `""`.
- The headers real API Gateway injects and the current builder does not: `host`, `x-amzn-trace-id`, `x-forwarded-for`, `x-forwarded-port`, `x-forwarded-proto`. The `cookie` header is removed from `headers` and surfaced only in `cookies`.
- Fail-closed refusals for the behaviour later issues in this milestone add: a `RouteKey` other than `$default`, a `StageName` other than `$default`, and an `AuthorizationType` other than `NONE` are each refused by name here. Issues 3, 6 and 7 accept them as they land.
- Control-plane IAM authorization for each command, through the shared `SimIamActionAuthorizer`, on `arn:aws:apigateway:<region>::/apis/<api-id>`.
- The `@aws-sdk/client-apigatewayv2` devDependency, the `./apigatewayv2` package export, the `docs/tsconfig.json` path entry, `docs/services/apigatewayv2/README.md` with `sim-apigatewayv2-*` labelled examples, `src/service/apigatewayv2/README.md`, and entries in `docs/README.md`, the root `README.md` and `docs/sdk/README.md`.

## Acceptance criteria

- [ ] `CreateApi` returns an `ApiEndpoint` of the form `https://<api-id>.execute-api.<region>.amazonaws.com` with no trailing slash and no stage, stable across `GetApi` calls and unique per API
- [ ] A `fetch` through `serveSimAws` to that endpoint on a `$default` route invokes the integration's function and returns its status, headers and body, with each `cookies[]` entry becoming one `set-cookie` header
- [ ] The event carries `version: "2.0"`, `routeKey: "$default"`, `rawPath`, `rawQueryString`, lowercased comma-joined `headers`, the `cookie` header removed from `headers` and present in `cookies`, and `requestContext.http.method`/`path`/`protocol`
- [ ] The event's `headers` include `host`, `x-amzn-trace-id`, `x-forwarded-for`, `x-forwarded-port` and `x-forwarded-proto` even when the client sent none of them
- [ ] `cookies`, `queryStringParameters` and `body` are absent from the event rather than null when empty, `rawQueryString` is `""` rather than absent, and `pathParameters` and `stageVariables` are always absent at this stage because only `$default` routes and stages exist
- [ ] A binary request `content-type` gives `isBase64Encoded: true` with a base64 body, and a text one gives the raw body, using the same rules the Function URL path already applies
- [ ] A handler returning a value with no numeric top-level `statusCode` yields 200 with the JSON serialisation of that value and `content-type: application/json`
- [ ] A handler that throws, and a handler returning a `statusCode` outside 100 to 599, each give 502 `{"message":"Internal Server Error"}`
- [ ] An unknown api id, a hostname whose region label does not match the API's region, and a request matching no route each give 404 `{"message":"Not Found"}`
- [ ] An API created with `DisableExecuteApiEndpoint: true` answers 403 `{"message":"Forbidden"}` on its generated endpoint and does not invoke the function
- [ ] `IntegrationType` other than `AWS_PROXY`, a missing or `1.0` `PayloadFormatVersion`, `ParameterMapping`, `ProtocolType: WEBSOCKET`, `CorsConfiguration`, `Tags`, a non-`$default` `RouteKey`, a non-`$default` `StageName` and a non-`NONE` `AuthorizationType` are each refused by name rather than accepted and ignored
- [ ] Two APIs with the same name in different simulated accounts each get their own api id, and each endpoint routes to the right account's API
- [ ] `ApiGatewayV2Client` is intercepted by `SimSdk`, including from inside a simulated Lambda handler
- [ ] A `docs/services/apigatewayv2/` page, with limitations stated

## Out of scope

- Route keys other than `$default`, path parameters, greedy `{proxy+}` routes, `ANY` and named stages, which have their own issue in this milestone.
- The `AWS::ApiGatewayV2::*` CloudFormation resources, which have their own issue in this milestone. Until that lands, a template naming them is unsupported CloudFormation input handled by the best-effort deployment model rather than being permanently refused.
- Authorizers of any kind. Every route is `AuthorizationType: NONE`. JWT authorizers and `AWS_IAM` route authorization each have their own issue in this milestone.
- Checking the integration function's resource policy before invoking it, which has its own issue in this milestone. Until then the integration invokes directly, which is a fail-open divergence and goes in Limitations from day one.
- `Update*` commands, and `DeleteRoute`, `DeleteIntegration` and `DeleteStage`, deferred beyond this milestone. Nothing in the milestone changes an API in place, and the CloudFormation issue's stack update and stack delete behaviour over `AWS::ApiGatewayV2::*` is deferred with them.
- Payload format 1.0, `HTTP_PROXY` and `IntegrationSubtype` integrations, `ParameterMapping`, inline OpenAPI (`Api.Body`), quick create (`Api.Target`), and every WebSocket-only resource type. All refused by name, all deferred beyond this milestone.
- `TimeoutInMillis` enforcement, deferred beyond this milestone. The 50 to 30000 range is validated and the value stored, but no 504 is produced, matching the fact that simulated Lambda does not interrupt a running handler either.
- CORS request handling, tagging and access logging, all deferred beyond this milestone.

## Implementation notes

This is expected to land as two PRs, in the style of #269, each ending `Part of` rather than `Resolves`. The first extracts the shared payload 2.0 code out of `src/service/lambda/serve/`. It is fully testable through the existing Function URL tests before any HTTP API exists, and the repo has shipped a source-extraction refactor that way before (#232). The pure pieces (`sim-lambda-url-body-encoding.ts`, `sim-lambda-url-request-parts.ts`, `sim-lambda-url-event-time.ts`) move to `src/util/http/`. The event builder in `sim-lambda-url-event-builder.ts`, the request-context builder in `sim-lambda-url-request-context.ts`, the response builder in `serve/response/sim-lambda-url-response-builder.ts` and the caller adapter in `sim-lambda-url-iam-authorizer.ts` move to a neutral home under `src/service/aws/`. Function URL behaviour is unchanged. The second PR is the new service.

There is one shared payload 2.0 event type after that extraction, owned by the neutral module. `@kensio/yulin/lambda` keeps exporting `SimLambdaFunctionUrlEvent` for it, from `src/service/lambda/index.ts:18`. That makes the extraction a user-visible breaking change rather than an internal one, for two reasons. `routeKey` and `stage` are currently the literal type `"$default"` on both the event and the request context, and widening them to `string` changes the shape a handler is typed against. `requestContext.authorizer` is currently `{ iam: ... }`, and widening it to a union is the second change. CLAUDE.md sanctions breaking changes; this one is worth calling out because handlers are typed against it. The `iam` block keeps the field types and values `SimLambdaUrlIamCaller.authorizerContext()` already produces (`accessKey: ""`, `callerId` and `userId` the caller ARN, `cognitoIdentity` and `principalOrgId` null), so the `AWS_IAM` route authorization issue reuses it rather than widening those fields again.

`SimAwsLocalUrl.awsHostnamePrefix` in `src/serve/http/url/sim-aws-local-url.ts` is an alternation of four regexes. They cover `<bucket>.s3|s3-website.<region>.amazonaws.com`, `s3.<region>.amazonaws.com`, `cognito-idp.<region>.amazonaws.com` and `<url-id>.lambda-url.<region>.on.aws`. None matches an execute-api host, so one today falls through to the whole-hostname rule and then fails to resolve. Add `/^(?<prefix>[^.]+\.execute-api\.[^.]+)\.amazonaws\.com$/`, where `[^.]+` is exact because an api id is one DNS label. Then add `"apiGatewayV2"` to the `SimAwsServiceTarget["service"]` union in `src/serve/controller/sim-service-controller.ts`, currently `"s3" | "cloudFront" | "lambda" | "route53" | "cognitoIdentityProvider"`. That forces a case in `simAwsServiceSigningName` returning `"execute-api"`, since that switch has no default arm. The switch in `SimAwsServiceControllerFactory` does have a throwing default, so its case fails at runtime rather than at compile time. Finally add a three-label `<api-id>.execute-api.<region>` branch to `SimRoute53ServiceTargetResolver` in `src/service/route53/resolve/sim-r53-service-target-resolver.ts`, modelled on the lambda-url branch there. Nothing under `src/serve/dns/` needs changing.

`src/service/aws/sim-aws.ts` is at 286 lines against the 300-line limit, so an `apiGatewayV2()` accessor needs a small extraction from that file first. `sim-r53-service-target-resolver.ts` at 182 lines is the other pressure point.

`SimLambdaFunctionUrlFactory.make` in `src/service/lambda/function/url/sim-lambda-function-url-factory.ts` allocates and registers the url id in one method, which is the shape the api id registry should copy. Wire it as an `@internal` field on `SimAwsServiceFactory` alongside `lambdaUrlRegistry`. The facade splits into per-resource command classes the way `src/service/sqs/command/queue/sim-sqs-queue-commands.ts` does, because one facade holding ten commands will not fit in 300 lines. The controller should check the region label in the hostname against the API's own region and return 404 otherwise, as `SimCognitoServiceController.servedPool` does.

Payload format selection should be an object with a single 2.0 implementation, and route authorization a polymorphic role with a `None` implementation. Payload 1.0 and the authorizer types then land as new implementations rather than edits to the controller.

Real HTTP APIs use a lowercase `message` key, unlike the capital-`Message` Function URL responses in `sim-lambda-url-error-response.ts`, so this needs its own error response class rather than reuse. The 500 and 502 split follows the Lambda developer guide, which states that a rejected invocation gives 500 and a function that errors or returns a wrongly formatted response gives 502. This issue only produces the 502 half; the 500 half arrives with the invoke permission issue. The alternative reading, extending the HTTP API Lambda authorizer page's documented 500 to integrations as well, was rejected because it is inferred rather than stated. Record that choice in a code comment and under Limitations. Body casing comes from the HTTP API troubleshooting page (`{"message":"Internal Server Error"}`), not the REST-oriented wording.

The SDK `serviceId` string is `"ApiGatewayV2"`, read as `serviceId: config?.serviceId ?? "ApiGatewayV2"` in `dist-es/runtimeConfig.shared.js` of the published `@aws-sdk/client-apigatewayv2` 3.1101.0 package. The package is not currently a devDependency, so add it and confirm the string against the installed copy. It is worth checking rather than guessing, because existing values include `"Cognito Identity Provider"` and `"Route 53"` and are not derivable from a package name.

Limitations to state from day one: `requestContext.http.sourceIp` is always `127.0.0.1`, and `requestId` is a UUID rather than the short opaque id AWS emits. Both are inherited from the Function URL builder. The integration invokes its function without checking the function's resource policy, and the invoke permission issue in this milestone removes that entry when it lands. The event omission rules for `cookies`, `queryStringParameters` and `body`, the injected `x-forwarded-*` headers, the removal of `cookie` from `headers`, the 404 status and the 403 for a disabled endpoint are all observed AWS behaviour rather than published, so note them as such at the point of construction. The null-or-omitted rule for `pathParameters` and `stageVariables` is documented.

The plan sizes this at 3000 to 3800 insertions across 45 to 55 files, split across the two PRs above. This issue depends on nothing else in the milestone.

hughgrigg:
Adds a simulated API Gateway v2 service covering HTTP APIs with an AWS_PROXY integration, reachable as simAws.apiGatewayV2(). An API, integration, $default route and $default stage make the generated execute-api endpoint serve real HTTP requests through serveSimAws, invoking the integrated function with a payload format 2.0 event.

The payload 2.0 event and response machinery moves out of the Lambda Function URL code into src/serve/payload-2/, so both services share it. The route key, stage and endpoint identity are now supplied by the endpoint, path parameters and stage variables are carried, and the authorizer block is a union with optional iam and jwt members.

Resolves https://github.com/KensioSoftware/yulin/issues/298

## Summary by CodeRabbit

* **New Features**
  * Added simulated API Gateway HTTP APIs with API, integration, route, and stage management.
  * Added Lambda proxy integrations using payload format 2.0.
  * Added local endpoint serving, request routing, IAM authorization, SDK support, and configurable stage variables.
  * Added support for disabled endpoints and AWS-style error responses.

* **Bug Fixes**
  * Improved Lambda Function URL handling when IAM authorization data is unavailable.

* **Documentation**
  * Added service documentation and working examples for API Gateway HTTP APIs.


hughgrigg:
@coderabbitai full review

PR #307, opened 2026-08-02
feat: match simulated HTTP API requests by route key, path parameter …

### LINKED ISSUE #299: Match simulated HTTP API requests by route key, path parameter and stage
## Problem

The simulated HTTP API serves a single `$default` route on the `$default` stage. Real APIs declare route keys like `GET /orders` and `GET /orders/{orderId}`. Which route a request matches decides what the handler does, and the matched route is what supplies the path parameters. Neither route keys nor path parameters are modelled. Named stages are not modelled either, and a named stage changes the path the client sends without changing `routeKey`.

## Desired behaviour

Route keys are parsed and validated at `CreateRoute`. A served request selects the route real API Gateway would select. Path parameters reach the handler in `event.pathParameters`. Named stages work alongside `$default`, with stage selection running before route selection.

## Example

```typescript
const apiGw = simAws.apiGatewayV2();

const { ApiId } = await apiGw.createApi(new CreateApiCommand({ Name: "pets", ProtocolType: "HTTP" }));

const { IntegrationId } = await apiGw.createIntegration(
  new CreateIntegrationCommand({
    ApiId,
    IntegrationType: "AWS_PROXY",
    IntegrationUri: functionArn,
    PayloadFormatVersion: "2.0",
  }),
);

for (const RouteKey of ["GET /pets", "GET /pets/{petId}", "ANY /admin/{proxy+}", "$default"]) {
  await apiGw.createRoute(
    new CreateRouteCommand({ ApiId, RouteKey, Target: `integrations/${IntegrationId}` }),
  );
}

await apiGw.createStage(new CreateStageCommand({ ApiId, StageName: "dev", AutoDeploy: true }));

await fetch(srv.localUrl(`https://${ApiId}.execute-api.eu-west-2.amazonaws.com/dev/pets/6`));
```

The handler receives `routeKey` `GET /pets/{petId}`, `pathParameters` `{ petId: "6" }`, `rawPath` `/dev/pets/6` and `requestContext.stage` `dev`.

## Proposed scope

- Route key grammar: `<METHOD> <path>` with an uppercase method token or `ANY`, or the literal `$default` with no method. Parsing throws at `CreateRoute`, which is where AWS rejects it too.
- One small class per segment kind (literal, `{name}`, `{proxy+}`) with polymorphic `matches` and `capture`.
- A matcher for one route against one request.
- A `SimApiGwRouteRank` value object with `compareTo` holding the whole precedence rule.
- `SimApiGwPathParameters` as a class, converted to a record only at the event boundary.
- Named stages, stage prefix stripping, and `StageVariables` on the stage surfaced as `event.stageVariables`.
- Docs sections on route keys, selection precedence with the worked table, path parameters and named stage URLs.

## Acceptance criteria

- [ ] A lowercase method, an unbalanced brace, a `{proxy+}` that is not the final segment, or any other malformed route key gives `BadRequestException`
- [ ] A duplicate route key gives `ConflictException`, including two keys that differ only in path parameter name
- [ ] A `Target` naming an integration that does not exist gives `NotFoundException`, an unknown `ApiId` on `CreateRoute` or `CreateStage` gives `NotFoundException`, and a second stage with an existing name gives `ConflictException`
- [ ] AWS's published five-route example is encoded verbatim as a test: with routes `GET /pets/dog/1`, `GET /pets/dog/{id}`, `GET /pets/{proxy+}`, `ANY /{proxy+}` and `$default` on the `$default` stage, `GET /pets/dog/1` selects `GET /pets/dog/1`, `GET /pets/dog/2` selects `GET /pets/dog/{id}`, `GET /pets/cat/1` selects `GET /pets/{proxy+}` and `POST /test/5` selects `ANY /{proxy+}`
- [ ] `{name}` matches exactly one segment, and `{proxy+}` is terminal only and captures the remainder with no leading slash, so `GET /pets/{proxy+}` matches `/pets/cat/1` with `pathParameters.proxy` of `cat/1` and does not match `/pets`
- [ ] `pathParameters` is omitted from the event when the matched route has no variable segments, including on `$default`
- [ ] A request whose path matches a route but whose method does not, on an API with no `ANY` route, no greedy route and no `$default` route to catch it, gives 404 `{"message":"Not Found"}` rather than 405
- [ ] Stage selection runs before route selection and an explicit stage match wins over any route match, so with a `$default` stage and a stage named `pets`, `GET /pets/dog` reaches stage `pets` on route path `/dog`
- [ ] For `/dev/pets/6` on stage `dev` matching `GET /pets/{petId}`, `rawPath` and `requestContext.http.path` are `/dev/pets/6`, `requestContext.stage` is `dev`, `routeKey` is `GET /pets/{petId}` and `pathParameters` is `{"petId":"6"}`
- [ ] A stage name is `$default`, or alphanumerics with hyphens and underscores up to 128 characters, and a request reaching an API with no matching stage and no `$default` stage gives 404 `{"message":"Not Found"}`
- [ ] `StageVariables` set on a stage arrive as `event.stageVariables`, omitted when the stage has none

## Out of scope

- `AWS::ApiGatewayV2::Route` and `AWS::ApiGatewayV2::Stage`, which have their own issue in this milestone.
- `AuthorizationType`, `AuthorizerId` and `AuthorizationScopes` on a route. `JWT` belongs to the JWT authorizer issue and `AWS_IAM` to the `AWS_IAM` route authorization issue, both in this milestone. Every route stays `NONE` here.
- `AutoDeploy: false`, `AWS::ApiGatewayV2::Deployment` and `Stage.DeploymentId`, all refused. AWS needs an explicit deployment for a stage that does not auto-deploy. What such a stage serves before that deployment exists is not established by the research, so refusing avoids guessing at it. CDK always sets `AutoDeploy: true`.
- `RouteSettings`, `DefaultRouteSettings` and `AccessLogSettings`, all refused.
- `UpdateRoute`, `DeleteRoute`, `UpdateStage` and `DeleteStage`, deferred beyond this milestone. A route or stage goes away with `DeleteApi`.
- Custom domain names and API mappings, deferred beyond this milestone. They change how `rawPath` behaves.

## Implementation notes

Three of the precedence rules are documented. AWS states the three tiers (full match, greedy, `$default`) and states that greedy beats `$default`. Literal-beats-variable within a full match comes from the first two rows of the worked example, and from the REST rule that a specific resource takes precedence over a generic one at the same level. Two further rules are observed rather than documented: an exact method beats `ANY`, and among greedy candidates at equal method specificity the longest literal prefix wins.

Several shaping and validation rules are also observed rather than documented. Those are the lowercase method rejected at `CreateRoute`, two route keys differing only in parameter name treated as a duplicate, the greedy capture holding the remainder with no leading slash, a greedy route not matching an empty remainder, `pathParameters` absent for a `$default` match, and the absence of 405. The 404 body for an unmatched stage or route is documented, the 404 status is not. The `NotFoundException` and `ConflictException` cases in the third acceptance criterion are not covered by the research at all, so check them against the API Gateway v2 API reference before pinning them. Mark each observed rule in the code next to the rule it governs, so it is visible when someone corrects one.

The same treatment applies to the stage prefix in `rawPath`. It is corroborated by the documented `$context.path` access-log variable ("The request path. For example, `/{stage}/root/child`") and by a community reproduction, not by the payload format page. Record that evidence in a comment at the point the event path fields are built.

Decomposing into one class per segment kind plus a separate rank object is what the repo's own limits require. ESLint `complexity` is on at its default of 20 per function (`eslint.config.ts:160`). `max-lines` is 300 for source (`eslint.config.ts:157`). `scripts/sh/fta.sh` fails at 50, and that score is dominated by token density rather than branching. A single matcher function with a switch over segment kinds and a scoring ternary will not pass `pnpm run check`. `noUncheckedIndexedAccess` also makes every `segments[i]` a `T | undefined`, which a segment class absorbs and a loop over indices does not.

Do not extend `SimCloudFrontPathPattern` (`src/service/cloudfront/resolver/sim-cloudfront-path-pattern.ts`). It is a glob compiled to a regex with no method dimension and no capture. `SimCloudFrontBehaviorResolver` picks a winner by counting non-wildcard graphemes (`sim-cloud-front-behavior-resolver.ts:72`). Character counting gives wrong answers here. `GET /a/{b}/ccccc` has more literal characters than `GET /a/b/{c}` and is less specific at the segment that decides. Match segment by segment instead of compiling regexes.

`routeKey` and `stage` are currently the literal type `"$default"` on `SimLambdaFunctionUrlEvent` (`src/service/lambda/serve/event/sim-lambda-url-event.type.ts:11` and `:33-34`). The foundation issue widens them to `string` and adds the optional `pathParameters` and `stageVariables` fields, so nothing further is needed here beyond populating them.

The foundation issue refuses a non-`$default` route key and a named stage by name, and records both in the API Gateway Limitations section. This issue implements them, so those two refusals go and that Limitations entry is deleted.

Sized at 2200 to 2800 insertions across 30 to 38 files.

Builds on the simulated HTTP API foundation issue in this milestone.

Depends on #298.

hughgrigg:
…and stage

Route keys are parsed and validated at CreateRoute, one class per segment kind. A served request selects the route real API Gateway would, with the precedence rule in SimHttpApiRouteRank.compareTo. Named stages work alongside $default, and stage selection runs before route selection.

Resolves https://github.com/KensioSoftware/yulin/issues/299

## Summary by CodeRabbit

- **New Features**
  - Added support for custom HTTP API routes, including literal, parameterized, greedy, `ANY`, and `$default` routes.
  - Added named-stage routing and stage-name validation.
  - Requests now capture route path parameters and include them in responses.
  - Route matching follows AWS precedence rules for methods, paths, and stages.
  - Added an executable API Gateway v2 routing example.

- **Documentation**
  - Expanded routing, stage behavior, path parameters, precedence, and known limitations.

- **Tests**
  - Added broad coverage for route parsing, matching, stages, validation, and served responses.

### DIFF
--- a/docs/services/apigatewayv2/README.md
+++ b/docs/services/apigatewayv2/README.md
@@ -134,6 +134,183 @@ The `$default` route matches any method and path, so every request to the endpoi
 function. The integration URI is the function's ARN, and the function may be in another Account or
 Region: it is looked up where its ARN says it is.
 
+## Route keys
+
+A route key is either the literal `$default` or an upper-case HTTP method and a path separated by one
+space:
+
+```text
+GET /pets
+GET /pets/{petId}
+ANY /admin/{proxy+}
+$default
+```
+
+The method is one of `GET`, `POST`, `PUT`, `PATCH`, `DELETE`, `HEAD`, `OPTIONS` or `ANY`, where `ANY`
+matches whatever method the request used.
+
+A path is made of three kinds of segment:
+
+- A literal, such as `pets`, matching that one segment.
+- A parameter, such as `{petId}`, matching exactly one segment, whatever is in it.
+- A greedy parameter, such as `{proxy+}`, matching everything left of the path. It is only ever the
+  last segment, and it needs at least one segment to match, so `GET /pets/{proxy+}` matches
+  `/pets/cat/1` but not `/pets`.
+
+A route key that cannot be read is refused by `CreateRoute` with a `BadRequestException`, which is
+where real API Gateway refuses it too. That covers a lower-case method, an unbalanced brace, and a
+greedy parameter anywhere but the end.
+
+A parameter name is not part of a route's identity, so `GET /pets/{id}` and `GET /pets/{petId}` are
+the same route key and creating the second gives a `ConflictException`.
+
+## Which route serves a request
+
+More than one route may match a request, and the most specific one takes it. In order:
+
+1. A route matching the whole path beats a route ending in a greedy parameter, which beats
+   `$default`.
+2. An exact method beats `ANY`.
+3. The path decides, segment by segment from the left, with a literal beating a `{name}` parameter
+   and a `{name}` beating a `{name+}`. Comparing left to right is also what makes the longest literal
+   prefix win between two greedy routes, so `GET /pets/dog/{proxy+}` takes `/pets/dog/collars/1`
+   ahead of `GET /pets/{proxy+}`.
+
+AWS's worked example, which is encoded as a test here:
+
+| Request           | Route selected       |
+| ----------------- | -------------------- |
+| `GET /pets/dog/1` | `GET /pets/dog/1`    |
+| `GET /pets/dog/2` | `GET /pets/dog/{id}` |
+| `GET /pets/cat/1` | `GET /pets/{proxy+}` |
+| `POST /test/5`    | `ANY /{proxy+}`      |
+
+from the routes `GET /pets/dog/1`, `GET /pets/dog/{id}`, `GET /pets/{proxy+}`, `ANY /{proxy+}` and
+`$default`.
+
+Rule 1 is documented by AWS, as is the literal-beating-parameter part of rule 3. Rule 2, the
+longest-literal-prefix part of rule 3, and the place of the method comparison above the path
+comparison rather than below it, are observed rather than documented. Each is marked in the code next
+to the rule it governs.
+
+A request whose path matches a route but whose method does not simply matches nothing. An API with no
+`ANY` route, no greedy route and no `$default` route to catch it answers 404, not 405.
+
+## Path parameters and named stages
+
+What a route captured reaches the handler as `event.pathParameters`. A named stage is served under
+its own path segment, and stage selection runs before route selection, so the routes never see the
+stage name.
+
+```typescript sim-apigatewayv2-routes
+/**
+ * Matching a simulated HTTP API request by route key, path parameter and stage.
+ */
+
+import {
+  CreateApiCommand,
+  CreateIntegrationCommand,
+  CreateRouteCommand,
+  CreateStageCommand,
+} from "@aws-sdk/client-apigatewayv2";
+import { CreateFunctionCommand } from "@aws-sdk/client-lambda";
+
+import { SimAws } from "@kensio/yulin";
+import type { SimPayload2Event } from "@kensio/yulin/apigatewayv2";
+import { makeLambdaZipFileInput } from "@kensio/yulin/lambda";
+import { serveSimAws } from "@kensio/yulin/serve";
+
+const simAws = new SimAws();
+
+const { FunctionArn } = await simAws.lambda().createFunction(
+  new CreateFunctionCommand({
+    FunctionName: "pets",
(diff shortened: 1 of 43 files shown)


hughgrigg:
@coderabbitai full review

PR #308, opened 2026-08-02
feat: require an invoke permission before a simulated HTTP API calls …

### LINKED ISSUE #300: Require an invoke permission before a simulated HTTP API calls its Lambda integration
## Problem

A Lambda proxy integration on a real HTTP API does not work until the function's resource policy allows `apigateway.amazonaws.com` to invoke it. AWS documents that a Lambda integration created through CloudFormation, the CLI or an SDK needs that permission granted explicitly. It documents the result of leaving it out as a 500 with the body `{"message":"Internal Server Error"}`, with the function never invoked. The simulated integration invokes its function unconditionally, so a template with no `AWS::Lambda::Permission` serves normally here.

The permission cannot be honoured today even when it is present. `SimLambdaPermission.condition()` writes `ArnLike { "AWS:SourceArn": ... }` into the statement (`src/service/lambda/function/policy/sim-lambda-permission.ts:106-108`). `SimIamConditionOperatorParser` registers seven operators, and none of them is `ArnLike` or `ArnEquals` (`src/service/iam/authorize/match/condition/sim-iam-condition-operator-parser.ts:12-35`). `SimIamPolicyConditionMatcher` fails closed on an operator it does not recognise. `docs/services/lambda/README.md:1332-1334` records the consequence: a statement carrying `SourceArn` never matches. Nothing supplies a value for the key at request time either.

## Desired behaviour

A matched route authorizes `lambda:InvokeFunction` on the function ARN before invoking it. The caller is the service principal `apigateway.amazonaws.com`, and the request supplies `AWS:SourceArn`. A denied or absent permission returns 500 `{"message":"Internal Server Error"}`, and the handler does not run. `ArnLike` and `ArnEquals` become supported IAM condition operators, so a `SourceArn` written by `AddPermission` or by `AWS::Lambda::Permission` starts matching.

## Example

```typescript
const created = await simAws
  .apiGatewayV2()
  .createApi(new CreateApiCommand({ Name: "orders", ProtocolType: "HTTP" }));
// Plus an AWS_PROXY integration, a "GET /orders/{orderId}" route and a $default stage.

const url = `https://${created.ApiId}.execute-api.eu-west-2.amazonaws.com/orders/42`;

const refused = await fetch(srv.localUrl(url));
// 500 with {"message":"Internal Server Error"}. The handler was not invoked.

await simAws.lambda().addPermission(
  new AddPermissionCommand({
    FunctionName: "orders",
    StatementId: "api-gateway-invoke",
    Action: "lambda:InvokeFunction",
    Principal: "apigateway.amazonaws.com",
    SourceArn: `arn:aws:execute-api:eu-west-2:111111111111:${created.ApiId}/*/*/orders/{orderId}`,
  }),
);

const allowed = await fetch(srv.localUrl(url));
// 200 from the handler.
```

## Proposed scope

- `ArnLike` and `ArnEquals` condition operators. `ArnLike` is `StringLike` semantics over an ARN, so `simIamWildcardMatch` already does the matching and the operator classes are thin.
- The integration authorizes `lambda:InvokeFunction` on the function ARN with caller `{ kind: "service", service: "apigateway.amazonaws.com" }`. It passes the function's resource policy and a request-time `conditionContext` carrying `AWS:SourceArn`.
- A source ARN builder producing `arn:aws:execute-api:<region>:<account>:<apiId>/<stage>/<METHOD>/<routePath>`. The method comes from the request. The path comes from the matched route key template, braces intact. For the `$default` route the method and path collapse to the single literal `$default`, which AWS gives as a documented CLI example.
- One builder class, with the method and path supplied by its caller. The `AWS_IAM` route authorization issue in this milestone builds the same ARN string from the concrete request path for a different purpose, so the two policies belong side by side with the difference stated in one place rather than in two unrelated comments.
- Service principal matching exercised for real, replacing the `/* v8 ignore */` markers on `SimIamPolicyPrincipalMatcher.servicePatternMatches` (`src/service/iam/authorize/match/sim-iam-policy-principal-matcher.ts:92`).
- Docs. The API Gateway page gets a section on the invoke permission with the exact source ARN shape. The Limitations entry added by the routing foundation issue, saying the integration invokes without checking the function's resource policy, is deleted. The `SourceArn` never-matches sentence at `docs/services/lambda/README.md:1332-1334` is corrected. The IAM page's supported-operator list gains the two operators.

## Acceptance criteria

- [ ] A route whose function has no matching permission returns 500 `{"message":"Internal Server Error"}` and the handler is not invoked
- [ ] `AddPermission` with `Principal: "apigateway.amazonaws.com"` and no `SourceArn` allows the invocation
- [ ] A permission whose `SourceArn` names a different route gives 500 `{"message":"Internal Server Error"}` and the handler is not invoked
- [ ] The source ARN uses the route key template path, so a `SourceArn` ending `/orders/{orderId}` allows a request to `/orders/42`
- [ ] A `$default` route builds `<apiId>/<stage>/$default`, so CDK's `<apiId>/*/*` matches it
- [ ] `ArnLike` and `ArnEquals` evaluate in an identity policy and in a resource policy, and a pattern with a wildcard in the region, account or resource-path position matches
- [ ] A Function URL grant made by `AddPermission` with a `SourceArn` matches at request time rather than never matching
- [ ] Docs state that `AWS:SourceAccount`, `aws:PrincipalOrgID` and `lambda:InvokedViaFunctionUrl` are still unsupplied, so a statement carrying one never matches

## Out of scope

- `AWS:SourceAccount`, `aws:PrincipalOrgID` and `lambda:InvokedViaFunctionUrl`. Two of them already have their operator: `SimLambdaPermission.stringEqualsConditions()` writes `SourceAccount` and `PrincipalOrgID` under `StringEquals`, which is registered. What all three lack is a request-time value, and that does not change here. `Bool`, which `InvokedViaFunctionUrl` needs, is not added.
- Integration `CredentialsArn`, the IAM role alternative to a resource policy.
- Any change to how a direct SDK `Invoke` or an SQS event source mapping authorizes.
- `AWS_IAM` route authorization, which has its own issue in this milestone. It is a different caller and a different decision.

## Implementation notes

The operator classes follow the existing per-operator directory layout under `src/service/iam/authorize/match/condition/`, so `arn/like/` and `arn/equals/` sit alongside `string/like/` and `numeric/less-than-equals/`, with two entries added to the map in `sim-iam-condition-operator-parser.ts`. Condition context keys are lowercased before comparison (`SimIamPolicyConditionMatcher.normalizeContextKey`), so `AWS:SourceArn` supplied at request time matches the key as written into the statement.

`SimLambdaUrlAuthorizer` (`src/service/lambda/serve/auth/sim-lambda-url-authorizer.ts`) is the model for the call. It returns a decision rather than throwing, passes `simLambdaResourcePolicies(simFunction)`, and supplies a request-time `conditionContext`. The integration wants the same shape, with `lambda:InvokeFunction` in place of `lambda:InvokeFunctionUrl`. `InvokeAuthorizer` (`src/service/lambda/command/invoke/invoke-authorizer.ts`) shows the throwing variant of the same authorization.

Both parts of the source ARN are inferred rather than documented, and both need a comment at the point of construction. The path is the better supported of the two. A CDK synth emits `SourceArn` as `<apiId>/*/*/orders/{orderId}`, and `simIamWildcardMatch` escapes regex metacharacters including braces (`src/service/iam/authorize/sim-iam-wildcard.ts:23-26`), so a concrete request path cannot match that pattern. The method segment is weaker. AWS's documented CLI example for an `ANY /test` route on the `prod` stage is `api-id/prod/*/test`, which is a permission pattern using a wildcard rather than the value API Gateway supplies, so it does not settle what the method segment holds at request time. A concrete verb matches any wildcard pattern, while a permission written with a literal `ANY` would not match it. Put that reasoning in the comment. A wrong guess makes every CDK-deployed route return 500.

`src/service/lambda/cfn/permission/sim-cfn-lambda-permission.iso.test.ts:158-193` asserts the current never-matches behaviour. It does so for an S3 source ARN, which still has no request-time value, so the assertion holds and its wording needs revisiting rather than the test being deleted.

`ArnLike`, `ArnEquals` and the Function URL `SourceArn` fix depend on nothing else in this milestone, and they are testable through the existing Lambda permission tests. If this runs past a reviewable size they split off cleanly, and can land ahead of the API Gateway work.

Builds on the route key, path parameter and stage matching issue in this milestone, because the matched route key template is what the source ARN path comes from.

Depends on #299.

hughgrigg:
…its Lambda integration

A matched route authorizes lambda:InvokeFunction on the function ARN before invoking it, as the service principal apigateway.amazonaws.com and with the matched route supplied as AWS:SourceArn. A denied or absent permission answers 500 {"message":"Internal Server Error"} and the handler does not run.

ArnLike and ArnEquals become supported condition operators, so a SourceArn written by AddPermission or by AWS::Lambda::Permission starts matching. Both compare the six components of an ARN separately, as AWS documents.

Resolves https://github.com/KensioSoftware/yulin/issues/300

## Summary by CodeRabbit

* **New Features**
  * API Gateway HTTP APIs now verify Lambda invocation permissions before invoking integrated functions.
  * Added support for route-, method-, stage-, and account-specific permission matching.
  * Added IAM `ArnEquals` and `ArnLike` condition support with wildcard ARN matching.
  * Lambda proxy examples now grant API Gateway permission by default, with configurable function accounts and permission behavior.

* **Bug Fixes**
  * Unauthorized Lambda integrations now return HTTP 500 responses instead of executing functions.

* **Documentation**
  * Expanded guidance on Lambda permissions, source ARN construction, matching behavior, and supported limitations.

### DIFF
--- a/docs/services/apigatewayv2/README.md
+++ b/docs/services/apigatewayv2/README.md
@@ -52,8 +52,11 @@ share a name, and only the id tells them apart.
 ## Routing requests to a Lambda function
 
 An API needs three more resources before it serves anything: an integration naming the function, a
-route pointing at that integration, and a stage to serve it from. Once they exist, `serveSimAws`
-answers requests to the generated endpoint by invoking the function.
+route pointing at that integration, and a stage to serve it from. The function also has to allow API
+Gateway to invoke it, which is a permission on the function rather than anything on the API. See
+[Granting the API permission to invoke the function](#granting-the-api-permission-to-invoke-the-function).
+Once all of that exists, `serveSimAws` answers requests to the generated endpoint by invoking the
+function.
 
 Pass the API endpoint through `srv.localUrl(...)`, which keeps the endpoint's hostname but sends the
 request to the local server, in the same way it adapts simulated S3 website and Lambda Function URL
@@ -70,7 +73,10 @@ import {
   CreateRouteCommand,
   CreateStageCommand,
 } from "@aws-sdk/client-apigatewayv2";
-import { CreateFunctionCommand } from "@aws-sdk/client-lambda";
+import {
+  AddPermissionCommand,
+  CreateFunctionCommand,
+} from "@aws-sdk/client-lambda";
 
 import { SimAws } from "@kensio/yulin";
 import type { SimPayload2Event } from "@kensio/yulin/apigatewayv2";
@@ -120,6 +126,16 @@ await apiGateway.createStage(
   new CreateStageCommand({ ApiId, StageName: "$default", AutoDeploy: true }),
 );
 
+await simAws.lambda().addPermission(
+  new AddPermissionCommand({
+    FunctionName: "orders",
+    StatementId: "api-gateway-invoke",
+    Action: "lambda:InvokeFunction",
+    Principal: "apigateway.amazonaws.com",
+    SourceArn: `arn:aws:execute-api:us-east-1:888888888888:${ApiId}/*/*`,
+  }),
+);
+
 const srv = await serveSimAws({ simAws });
 
 const response = await fetch(srv.localUrl(`${ApiEndpoint}/orders?limit=10`));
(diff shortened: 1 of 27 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #309, opened 2026-08-02
feat: deploy a simulated HTTP API from AWS::ApiGatewayV2 CloudFormati…

### LINKED ISSUE #301: Deploy a simulated HTTP API from AWS::ApiGatewayV2 CloudFormation resources
## Problem

None of the `AWS::ApiGatewayV2::*` resource types create anything. `resolveSimCloudFormationServiceResourceFactory` in `src/service/cloudformation/resource/resolve/service/sim-cfn-service-resolver.ts` has no `ApiGatewayV2` case. Its default branch throws `Unsupported sim CloudFormation Resource service ApiGatewayV2`. That message contains both `Unsupported sim` and `CloudFormation`, which is what `SimCfnResourceCreateOperation.isUnsupportedResourceError` matches, so the resource is skipped instead of failing the stack.

A real CDK-synthesized `HttpApi` template was deployed into the simulator during research. The stack reached `CREATE_COMPLETE` with all ten `AWS::ApiGatewayV2::*` resources skipped and no API created. The outputs resolved to stand-in strings. `Fn::GetAtt Api ApiEndpoint` came back as `HttpApiF5A9A8A7.ApiEndpoint`. The `AWS::Lambda::Permission` `SourceArn` came back as `arn:aws:execute-api:eu-west-2:111111111111:HttpApiF5A9A8A7/*/*/orders`, with the logical ID standing in for the api id.

## Desired behaviour

`AWS::ApiGatewayV2::Api`, `Integration`, `Route` and `Stage` create simulated resources. A CDK-synthesized or hand-written template then deploys an HTTP API that serves requests. `Ref` and `Fn::GetAtt` resolve the way real CloudFormation resolves them for each of the four types.

Every property outside the simulated subset is refused by name. A refusal fails the stack, naming the resource type, the logical ID, the property and the simulated set. Nothing is accepted and then ignored.

## Example

```typescript
  stackName: "orders-stack",
  template: {
    Resources: {
      // The AWS::Lambda::Function handler and its AWS::Lambda::Permission are omitted.
      Api: {
        Type: "AWS::ApiGatewayV2::Api",
        Properties: { Name: "orders", ProtocolType: "HTTP" },
      },
      Stage: {
        Type: "AWS::ApiGatewayV2::Stage",
        Properties: { ApiId: { Ref: "Api" }, StageName: "$default", AutoDeploy: true },
      },
      Integration: {
        Type: "AWS::ApiGatewayV2::Integration",
        Properties: {
          ApiId: { Ref: "Api" },
          IntegrationType: "AWS_PROXY",
          IntegrationUri: { "Fn::GetAtt": ["Handler", "Arn"] },
          PayloadFormatVersion: "2.0",
        },
      },
      Route: {
        Type: "AWS::ApiGatewayV2::Route",
        Properties: {
          ApiId: { Ref: "Api" },
          RouteKey: "GET /orders",
          AuthorizationType: "NONE",
          Target: { "Fn::Join": ["", ["integrations/", { Ref: "Integration" }]] },
        },
      },
    },
    Outputs: { ApiUrl: { Value: { "Fn::GetAtt": ["Api", "ApiEndpoint"] } } },
  },
});

await stack.waitForDeployComplete();

// https://<api-id>.execute-api.eu-west-2.amazonaws.com
const apiUrl = stack.outputs.get("ApiUrl")?.value;

try {
  const response = await fetch(srv.localUrl(`${apiUrl}/orders`));
} finally {
  srv.close();
}
```

## Proposed scope

- A CloudFormation resource factory for the service, plus an `ApiGatewayV2` case in `sim-cfn-service-resolver.ts`.
- A creator and an allow-list property parser for `Api`, accepting `Name`, `Description` and `ProtocolType: HTTP`, plus `DisableExecuteApiEndpoint`.
- A creator and property parser for `Integration`, accepting `IntegrationType: AWS_PROXY`, `IntegrationUri`, `PayloadFormatVersion: "2.0"` and `TimeoutInMillis`.
- A creator and property parser for `Route`, accepting `RouteKey`, `Target` and `AuthorizationType: NONE`.
- A creator and property parser for `Stage`, accepting `StageName`, `AutoDeploy: true`, `StageVariables` and `Description`.
- `IntegrationUri` accepted in the bare function ARN form CDK emits, and in the `arn:<partition>:apigateway:<region>:lambda:path/2015-03-31/functions/<function-arn>/invocations` form.
- `Route.Target` parsed from the string `integrations/<integration-id>`. CDK builds it with `Fn::Join`, and there is no id property to read instead.
- A value adapter per resource type, including the `Stage` case where `Ref` is the stage name and there are no attributes.
- A `.loc.test.ts` synthesizing a real `aws-cdk-lib` `HttpApi` with `HttpLambdaIntegration` and `addRoutes`, deploying it, and calling the served endpoint.
- The four resource types added to the supported list in `docs/services/cloudformation/README.md`, and a CloudFormation section in the API Gateway docs page.

## Acceptance criteria

- [ ] A template with `Api`, `Integration`, `Route` and `Stage` deploys, and a request to the deployed endpoint reaches the integration's simulated Lambda
- [ ] `Ref` returns the api id, the route id and the integration id, and `Fn::GetAtt` resolves `ApiId`, `RouteId` and `IntegrationId`
- [ ] `Fn::GetAtt Api ApiEndpoint` returns `https://<api-id>.execute-api.<region>.amazonaws.com`, with no trailing slash and no stage segment
- [ ] That endpoint and CDK's `httpApi.url`, which resolves through `AWS::URLSuffix` to the `sim-aws.localhost` form, both reach the same served API
- [ ] `Ref` on a `Stage` returns the stage name, and `Fn::GetAtt` on a `Stage` throws, since AWS documents no attributes for it
- [ ] One `Integration` referenced by several `Route` targets works, and a `Stage` created before its routes still serves them
- [ ] An `Api` deployed with `DisableExecuteApiEndpoint: true` serves 403 `{"message":"Forbidden"}` on the default endpoint
- [ ] A `Route` with an `AuthorizationType` other than `NONE`, and other than a type a later issue in this milestone adds, fails the stack rather than deploying
- [ ] A `Route` whose `AuthorizerId` resolves to the logical ID of a skipped resource fails the stack, naming both logical IDs
- [ ] Each refusal fails the stack with an error naming the resource type, the logical ID, the property and the simulated set
- [ ] A CDK `HttpApi` stack synthesizes, deploys and serves, with the `AWS::Lambda::Permission` gating the invocation
- [ ] Docs list the four resource types, with the `Body` refusal and the `CorsConfiguration` refusal recorded under Limitations as deliberate divergences

## Out of scope

- `AWS::ApiGatewayV2::Authorizer`, and `AuthorizerId` and `AuthorizationScopes` on `Route`, which have their own issue in this milestone.
- `AuthorizationType: JWT` and `AuthorizationType: AWS_IAM`, each added to this issue's `Route` allow-list by its own issue in this milestone.
- `Api.CorsConfiguration`, refused by name. CORS request handling is deferred beyond this milestone, and the research recommends refusing the property rather than accepting it with no effect. A CDK stack using `corsPreflight` will not deploy until CORS lands.
- `Tags` on `Api` and on `Stage`, refused by name. No tagging command exists on the simulated service, so an accepted tag would be unreadable.
- `Api.Body`, refused by name here. OpenAPI import has its own issue in this milestone, which replaces this refusal with a document translator. `Api.BodyS3Location` stays refused beyond that issue as well.
- `Api.Target` and `Api.RouteKey`. Quick create is deferred beyond this milestone and is refused by name.
- A template combining `Api.Body` with `Route`, `Integration` or `Authorizer` resources for the same `ApiId`. AWS documents that the two modes must not be combined but does not reject the template, so refusing it is stricter than AWS. The OpenAPI import issue carries this refusal forward, since both declaration paths exist once it lands.
- `ProtocolType: WEBSOCKET`, and the WebSocket-only `Model`, `RouteResponse` and `IntegrationResponse` types. All refused, with a message saying they are WebSocket only.
- Any `IntegrationType` other than `AWS_PROXY`, a missing or `1.0` `PayloadFormatVersion`, `IntegrationMethod`, `ParameterMapping`, `AutoDeploy: false`, `Stage.DeploymentId`, `AccessLogSettings`, `RouteSettings` and `DefaultRouteSettings`. All refused, all deferred beyond this milestone.
- `Deployment`, `DomainName`, `ApiMapping`, `VpcLink`, `RoutingRule` and `ApiGatewayManagedOverrides`. Custom domain names are the natural next milestone.
- Stack update and stack delete over these resource types. `docs/services/cloudformation/README.md` already records that stack updates and deletes are not supported at all, so there is nothing resource-type-specific to define here.

## Implementation notes

Follow the split the other services use. The factory, creators and property parsers live in a new `src/service/apigatewayv2/cfn/`. Mirror `src/service/cognito/cfn/`, which has `sim-cfn-cognito-resource-factory.ts` switching on the bare type name, one `*-creator.ts` and one `*-properties.ts` per type, and a shared property parser. The value adapters live under CloudFormation instead, in a new `src/service/cloudformation/resource/cfn/apigatewayv2/`, beside the existing `cognito/` and `lambda/` directories. One term is added to the `??` chain in `simCfnResourceValueAdapter`.

Refusal wording should match `unsimulatedPropertyError` in `src/service/cognito/cfn/sim-cfn-cognito-property-parser.ts`, which names the resource type, the logical ID, the property and the simulated set. The `Stage` adapter has a direct precedent in `src/service/cloudformation/resource/cfn/cognito/sim-cognito-user-pool-group-cfn.ts`, whose `attributeValue` throws for every attribute name.

The `Route` allow-list is written here with `AuthorizationType: NONE` only. The JWT authorizer issue adds `JWT` to that one enum, and the `AWS_IAM` issue adds `AWS_IAM`. Each amends the acceptance criterion above rather than replacing the parser.

The skipped-authorizer case needs care. `SimCfnResourceCreateOperation.isUnsupportedResourceError` treats any error containing both `Unsupported sim` and `CloudFormation` as a skip. The factory's default branch for an out-of-scope type such as `Authorizer` therefore marks the resource `CREATE_COMPLETE` with `simResource` undefined. A skipped resource matches no adapter, so its `Ref` falls through to `SimCfnDefaultResourceValueAdapter` and returns the logical ID string. A `Route` carrying `AuthorizationType: JWT` would otherwise deploy pointing at an authorizer that does not exist. The route creator has `context.resources`, a `ReadonlyMap<string, SimCfnResource>` with a `skipped` flag per resource, and that is the only signal available at creation time.

Ordering needs no `DependsOn` handling. A real synth carries none. The `Ref` and `Fn::GetAtt` edges alone give Api before Integration, Authorizer and Stage, then Route. The stage's only dependency is the api, which is why a stage created before the routes must still serve them.

The two endpoint forms come from different places. `Fn::GetAtt ApiEndpoint` is the `amazonaws.com` host, matching how `SimLambdaFunctionUrl` reports the real AWS hostname. CDK's `httpApi.url` is a `Fn::Join` over `AWS::URLSuffix`, which `src/service/cloudformation/parameters/pseudo/sim-cfn-pseudo-parameters.ts` deliberately resolves to `sim-aws.localhost`. The serving-layer hostname rule added earlier in this milestone is what makes both reach the same API, so assert both.

The 403 body for a disabled default endpoint is recorded in the research as observed rather than documented by AWS. Note that in a comment where the response is built.

The CDK test belongs in a new `src/service/cloudformation/cdk/apigatewayv2/` directory, following `src/service/cloudformation/cdk/lambda/sim-cdk-lambda-function-url.loc.test.ts` and its use of `TestCdkProject`.

Around 1800 to 2400 insertions across 24 to 30 files, in line with the other CloudFormation-only PRs in the repo.

Builds on the route key, path parameter and stage matching issue, and on the Lambda invoke permission issue, both in this milestone. With the permission issue landed first, the CDK test asserts an invocation that a real permission gates.

Depends on #299 and #300.

hughgrigg:
…on resources

AWS::ApiGatewayV2::Api, Integration, Route and Stage now create simulated resources, so a CDK-synthesized or hand-written template deploys an HTTP API that serves requests. Ref and Fn::GetAtt resolve per resource type, and every property outside the simulated set is refused by name rather than ignored.

Adding the service case pushed the CloudFormation service resolver over the complexity limit, so its switch is now a map of per-service delegations.

Resolves https://github.com/KensioSoftware/yulin/issues/301

## Summary by CodeRabbit

* **New Features**
  * Added CloudFormation support for simulated API Gateway v2 HTTP APIs, integrations, routes, and stages.
  * Supports API deployment, Lambda integrations, multiple routes, stages, stage variables, and generated endpoints.
  * Added support for CloudFormation `Ref` and `Fn::GetAtt` values, including API, integration, and route identifiers.
* **Bug Fixes**
  * Improved validation and diagnostics for unsupported properties, invalid values, unresolved authorizers, and unsupported resource types.
* **Documentation**
  * Added deployment examples and documented supported properties, limitations, URI formats, and endpoint behavior.

### DIFF
--- a/docs/services/apigatewayv2/README.md
+++ b/docs/services/apigatewayv2/README.md
@@ -496,6 +496,160 @@ only through a custom domain is configured. A request to it is answered with a 4
 `{"message":"Forbidden"}`. AWS publishes neither the status nor the body for that case, so both are
 what a disabled endpoint was observed to answer rather than something documented.
 
+## CloudFormation
+
+[Simulated CloudFormation](../cloudformation/ "Simulated CloudFormation docs") deploys
+`AWS::ApiGatewayV2::Api`, `AWS::ApiGatewayV2::Integration`, `AWS::ApiGatewayV2::Route` and
+`AWS::ApiGatewayV2::Stage`, so a synthesized or hand-written template produces an API that serves
+requests.
+
+`Ref` and `Fn::GetAtt` return what real CloudFormation returns for each type:
+
+| Resource type | `Ref`              | `Fn::GetAtt`                |
+| ------------- | ------------------ | --------------------------- |
+| `Api`         | the API id         | `ApiId`, `ApiEndpoint`      |
+| `Integration` | the integration id | `IntegrationId`             |
+| `Route`       | the route id       | `RouteId`                   |
+| `Stage`       | the stage name     | none, as AWS documents none |
+
+`Fn::GetAtt: ["Api", "ApiEndpoint"]` is the generated endpoint with no trailing slash and no stage
+segment, on the real `amazonaws.com` hostname. CDK's `httpApi.url` is built from `AWS::URLSuffix`
+instead, which resolves to the local `sim-aws.localhost` form. Both reach the same served API.
+
+An integration's `IntegrationUri` is accepted as the bare Lambda function ARN CDK emits, and as the
+`arn:aws:apigateway:<region>:lambda:path/2015-03-31/functions/<function-arn>/invocations` form. A
+route's `Target` is the `integrations/<integration-id>` string, which CDK builds with `Fn::Join` over
+a `Ref` to the integration.
+
+```typescript sim-apigatewayv2-cloudformation
+/**
+ * Deploying a simulated HTTP API from a CloudFormation template.
+ */
+
+import { SimAws } from "@kensio/yulin";
+import { serveSimAws } from "@kensio/yulin/serve";
+
+const simAws = new SimAws();
+
+const stack = await simAws.cloudFormation().deployTemplate({
+  stackName: "orders-stack",
+  template: {
+    Resources: {
+      HandlerRole: {
+        Type: "AWS::IAM::Role",
+        Properties: {
+          RoleName: "orders-role",
+          AssumeRolePolicyDocument: {
+            Version: "2012-10-17",
+            Statement: [
+              {
+                Effect: "Allow",
+                Principal: { Service: "lambda.amazonaws.com" },
+                Action: "sts:AssumeRole",
+              },
+            ],
+          },
+        },
+      },
+      Handler: {
+        Type: "AWS::Lambda::Function",
+        Properties: {
+          FunctionName: "orders",
+          Role: { "Fn::GetAtt": ["HandlerRole", "Arn"] },
+          Handler: "index.handler",
+          Runtime: "nodejs20.x",
+          Code: {
+            ZipFile:
+              "exports.handler = async () => ({ statusCode: 200, body: 'orders' });",
+          },
+        },
+      },
+      HandlerPermission: {
+        Type: "AWS::Lambda::Permission",
+        Properties: {
+          Action: "lambda:InvokeFunction",
+          FunctionName: { "Fn::GetAtt": ["Handler", "Arn"] },
+          Principal: "apigateway.amazonaws.com",
+          SourceArn: {
+            "Fn::Join": [
+              "",
+              [
+                "arn:aws:execute-api:us-east-1:111111111111:",
+                { Ref: "Api" },
+                "/*/*",
+              ],
+            ],
+          },
+        },
+      },
+      Api: {
+        Type: "AWS::ApiGatewayV2::Api",
+        Properties: { Name: "orders", ProtocolType: "HTTP" },
+      },
+      Stage: {
+        Type: "AWS::ApiGatewayV2::Stage",
+        Properties: {
+          ApiId: { Ref: "Api" },
+          StageName: "$default",
+          AutoDeploy: true,
+        },
+      },
+      Integration: {
+        Type: "AWS::ApiGatewayV2::Integration",
+        Properties: {
+          ApiId: { Ref: "Api" },
+          IntegrationType: "AWS_PROXY",
+          IntegrationUri: { "Fn::GetAtt": ["Handler", "Arn"] },
+          PayloadFormatVersion: "2.0",
+        },
+      },
+      Route: {
+        Type: "AWS::ApiGatewayV2::Route",
+        Properties: {
+          ApiId: { Ref: "Api" },
+          RouteKey: "GET /orders",
+          AuthorizationType: "NONE",
+          Target: {
+            "Fn::Join": ["", ["integrations/", { Ref: "Integration" }]],
+          },
+        },
+      },
+    },
+    Outputs: {
+      ApiEndpoint: { Value: { "Fn::GetAtt": ["Api", "ApiEndpoint"] } },
+    },
+  },
+});
+
+await stack.waitForDeployComplete();
+
+// https://<api-id>.execute-api.us-east-1.amazonaws.com
+const apiEndpoint = stack.outputs.get("ApiEndpoint")?.value as string;
+
+const srv = await serveSimAws({ simAws });
+
+const response = await fetch(srv.localUrl(`${apiEndpoint}/orders`));
+
+console.log(response.status);
+console.log(await response.text());
+
+srv.close();
+```
+
+Every property outside the simulated set is refused by name, and the refusal fails the stack. The
+simulated properties are:
+
+- `Api`: `Name`, `ProtocolType`, `Description`, `DisableExecuteApiEndpoint`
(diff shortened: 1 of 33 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #310, opened 2026-08-02
feat: authorize simulated HTTP API routes with a Cognito JWT authorizer

### LINKED ISSUE #302: Authorize simulated HTTP API routes with a Cognito JWT authorizer
## Problem

A simulated HTTP API route with a Lambda integration is open to anyone. `AuthorizationType` is `NONE` and no token is checked. Simulated Cognito already issues real RS256-signed tokens against a real per-pool JWKS. Nothing in `src/` verifies one: a grep for `createVerify` and `createPublicKey` across `src/` returns nothing. A test covering a protected route has to skip the authorization or mock it.

There is a second gap worth closing at the same time. A JWT authorizer never inspects `token_use`, which is not in AWS's documented validation list. An ID token sent to an authorizer that configures only `Audience` therefore passes, because its `aud` is the app client id. AWS documents this and recommends route scopes as the workaround. A Cognito ID token has no `scope` claim, so `AuthorizationScopes` is the only thing that discriminates the two token types.

## Desired behaviour

A JWT authorizer attached to a route verifies a signed Cognito token in process, against the pool's own JWKS, before the integration is invoked. Sign a user in through the simulated Cognito auth flows and call the API with the access token. The claims arrive in `requestContext.authorizer.jwt`. Verification reads `SimClock`, so advancing the simulation clock past the token's `exp` turns a 200 into a 401.

Failures follow the order AWS documents:

1. identity source present
2. decode
3. algorithm and signature
4. `kid`
5. `iss`
6. `aud`, or `client_id` when `aud` is absent
7. `exp`, `nbf`, `iat`
8. route scopes

Everything up to and including claim validation returns 401 `{"message":"Unauthorized"}` with a `www-authenticate: Bearer` header. An unmet route scope returns 403 `{"message":"Forbidden"}`.

## Example

```typescript
const authorizer = await simAws.apiGatewayV2().createAuthorizer(
  new CreateAuthorizerCommand({
    ApiId: apiId,
    Name: "pool-authorizer",
    AuthorizerType: "JWT",
    IdentitySource: ["$request.header.Authorization"],
    JwtConfiguration: {
      Issuer: `https://cognito-idp.eu-west-2.amazonaws.com/${userPoolId}`,
      Audience: [clientId],
    },
  }),
);

  new CreateRouteCommand({
    ApiId: apiId,
    RouteKey: "GET /orders",
    Target: `integrations/${integrationId}`,
    AuthorizationType: "JWT",
    AuthorizerId: authorizer.AuthorizerId,
  }),
);

const signedIn = await simAws.cognitoIdentityProvider().initiateAuth(
  new InitiateAuthCommand({
    ClientId: clientId,
    AuthFlow: "USER_PASSWORD_AUTH",
    AuthParameters: { USERNAME: "ada", PASSWORD: "Correct-horse-1" },
  }),
);
const accessToken = signedIn.AuthenticationResult?.AccessToken ?? "";

const url = srv.localUrl(
  `https://${apiId}.execute-api.eu-west-2.amazonaws.com/orders`,
);

await fetch(url);
// 401 {"message":"Unauthorized"}, www-authenticate: Bearer

await fetch(url, { headers: { authorization: `Bearer ${accessToken}` } });
// 200, and the handler saw requestContext.authorizer.jwt.claims

await simAws.clock().advanceBy({ hours: 2 });

await fetch(url, { headers: { authorization: `Bearer ${accessToken}` } });
// 401 again, because exp has passed
```

## Proposed scope

- `CreateAuthorizer`, `GetAuthorizers` and `DeleteAuthorizer` for `AuthorizerType: JWT`. `JwtConfiguration.Issuer` and `Audience` are required. `IdentitySource` takes a single entry, either `$request.header.<name>` or `$request.querystring.<name>`. The `REQUEST`-only fields are refused by name.
- A `Bearer ` prefix on the identity source value is stripped case-insensitively. It is not required.
- A JWT parse and RS256 verification module in a new `src/util/jwt/`. Three-part split, base64url decode and structural guards. An RS256-only `alg` allow-list refusing `none` and `HS*`. `kid` selection, then signature verification from a published JWK.
- Claim validation against `SimClock` with zero skew on `exp`, `nbf` and `iat`. `client_id` is consulted only when `aud` is entirely absent, so an access token carrying a resource binding fails rather than falling back to `client_id`.
- `AuthorizationScopes` as any-of over the whitespace-split `scope` or `scp` claim.
- `requestContext.authorizer.jwt.claims`, with `.scopes` as a string array or `null` when the token carries no `scope` claim. This widens the shared payload 2.0 event's `authorizer` field from `{ iam: ... }` to a union, which is a public breaking change on a type reachable from the exported `SimLambdaFunctionUrlEvent`.
- `AWS::ApiGatewayV2::Authorizer` with a JWT-only allow-list. `JWT` is added to the `AuthorizationType` values the `AWS::ApiGatewayV2::Route` property parser accepts, along with `AuthorizerId` and `AuthorizationScopes`. CDK's `HttpJwtAuthorizer` and `HttpUserPoolAuthorizer` then both deploy.
- Docs section on protecting a route with a Cognito user pool, covering the access token versus ID token behaviour.
- If this runs past a reviewable size, `src/util/jwt/` splits cleanly from the authorizer work. It is testable against a token from `SimCognitoTokenIssuer` before any authorizer exists.

## Acceptance criteria

- [ ] `CreateAuthorizer` with `AuthorizerType: JWT` requires `JwtConfiguration.Issuer` and `Audience`, accepts one header or query-string `IdentitySource`, and refuses the `REQUEST`-only fields by name
- [ ] A route with `AuthorizationType: JWT` returns 401 `{"message":"Unauthorized"}` with `www-authenticate: Bearer` when the token is missing, undecodable, signed with an unsupported algorithm, signed by an unknown `kid`, or fails signature, `iss`, audience, `exp`, `nbf` or `iat`, and the integration is never invoked
- [ ] A valid access token from a simulated Cognito sign-in reaches the integration, with `requestContext.authorizer.jwt.claims` present and `scopes` split from the `scope` claim
- [ ] A route with `AuthorizationScopes: ["aws.cognito.signin.user.admin"]` reaches the integration for that access token, because a route scope matches a claimed scope
- [ ] A route with `AuthorizationScopes` returns 403 `{"message":"Forbidden"}` when no route scope matches a claimed scope, which is what an ID token gets
- [ ] An ID token passes an authorizer that configures only `Audience`, because `token_use` is not checked
- [ ] A token invalidated by `GlobalSignOut` also still passes, because the pool's issued-token store is not consulted
- [ ] `simAws.clock().advanceBy({ hours: 2 })` turns a 200 into a 401 on an already-issued token, with nothing reissued
- [ ] `AWS::ApiGatewayV2::Authorizer` deploys with a JWT configuration, and a `Route` with `AuthorizationType: JWT` whose `AuthorizerId` did not resolve to a deployed authorizer fails the stack rather than deploying an open route
- [ ] A `.loc.test.ts` deploys one CDK stack with a `UserPool`, a `disableOAuth` client, an `HttpUserPoolAuthorizer` and a protected route, then asserts 401, 200, and 401 after the clock advances
- [ ] Docs section on protecting a route with a user pool, with the claim string coercion and the `null` `scopes` recorded under Limitations as observed rather than documented

## Out of scope

- `AWS_IAM` route authorization, which has its own issue in this milestone and adds `AWS_IAM` to the same set of accepted `AuthorizationType` values.
- Lambda `REQUEST` authorizers of either payload version, refused by name at `CreateAuthorizer` and in CloudFormation. Deferred beyond this milestone.
- Authorizer result caching. `AuthorizerResultTtlInSeconds` is one of the `REQUEST`-only fields refused above, so caching cannot be configured at all. Deferred beyond this milestone.
- Cognito resource servers, custom scopes and the client credentials grant, deferred beyond this milestone. No simulated flow can put any scope other than `aws.cognito.signin.user.admin` in a token, so that is the only satisfiable route scope. It goes under Limitations.
- Multiple identity sources, JWKS caching, key rotation and more than one signing key per pool.

## Implementation notes

`SimCognitoSigningKey` (`src/service/cognito/user-pool/token/sim-cognito-signing-key.ts`) keeps only the private key. It does expose `publicJwk()` carrying `n` and `e`. `createPublicKey({ key: jwk, format: "jwk" })` reconstructs the public key from that, so the verifier works off any issuer's published JWKS and the key class needs no change.

`aws-jwt-verify` cannot be used in implementation code. The CLAUDE.md import rule names `@aws-sdk/*` packages, so it does not literally cover it. It is a devDependency, and the package's runtime dependencies are `@faker-js/faker`, `@kensio/part-factory` and `lodash-es`. Shipping it would mean adding a runtime dependency. Write the verifier against `node:crypto`. Tests and docs examples may keep using `aws-jwt-verify`.

Reaching a pool follows the ACM to Route53 precedent. `SimAcmDnsRecords` (`src/service/acm/validation/sim-acm-dns-records.ts`) is the port the consumer owns. `SimRoute53AcmDnsRecords` is the adapter, constructed in `SimAwsAccountRegionServiceBuilder.createAcm` at `src/service/aws/factory/sim-aws-account-region-service-builder.ts:79`. Mirror that here. API Gateway owns the issuer port. A Cognito adapter over `SimCognitoUserPoolRegistry` (`src/service/cognito/registry/sim-cognito-user-pool-registry.ts`) parses `https://cognito-idp.<region>.amazonaws.com/<poolId>` back to a pool, and checks region agreement the way `SimCognitoServiceController.servedPool` does. A refusing implementation covers a standalone `SimApiGatewayV2`. API Gateway does not import `SimCognitoUserPool`, because a JWT authorizer accepts any OIDC issuer.

The authorizer never fetches JWKS or discovery over HTTP. That is deliberate. `SimCognitoOpenIdConfiguration` (`src/service/cognito/serve/sim-cognito-openid-configuration.ts`) publishes an `issuer` naming the localhost origin, while a token's `iss` names the real AWS URL, so an OIDC-conformant discovery client would reject its own pool's tokens. Resolving in process compares `pool.issuerUrl` against the token's `iss`, and both come from the same getter. `Fn::GetAtt <UserPool>.ProviderURL` already returns that string (`src/service/cloudformation/resource/cfn/cognito/sim-cognito-user-pool-cfn.ts:44-45`), so a CDK-declared authorizer's issuer and a token's issuer agree with no special handling.

Two things are deliberately not done. `token_use` is not checked, matching AWS, and the consequence belongs in the docs rather than being corrected in code. The pool's issued-token store (`src/service/cognito/user-pool/auth/sim-cognito-issued-token-store.ts`) is not consulted, because real API Gateway knows nothing about revocation. Consulting it would reject a token here that AWS would accept.

Three details are observed rather than published by AWS. The 403 body `{"message":"Forbidden"}`. The string coercion of claim values, where array claims render Go-style as `[GroupA GroupB]`, not JSON and not comma-separated. And `scopes` being `null` rather than `[]` when the token has no `scope` claim. Emit only the one `error_description` AWS publishes, `"the token does not have a valid audience"`, and omit the description elsewhere rather than inventing text. Record each of these in Limitations as observed.

The event type widening lands in this issue. `SimLambdaFunctionUrlAuthorizerContext` (`src/service/lambda/serve/event/sim-lambda-url-event.type.ts:46-48`) is `{ iam: ... }` today, and the shared payload 2.0 extraction leaves it that way. Adding a `jwt` member makes it a union, which changes the type handlers are written against.

For the CDK test, pass the app client explicitly through `userPoolClients`. `HttpUserPoolAuthorizer.bind` otherwise calls `pool.addClient("UserPoolAuthorizerClient")` with defaults, which emits the OAuth properties that stay refused.

This amends the CloudFormation issue's acceptance criterion that a `Route` with an `AuthorizationType` other than `NONE` fails the stack. `JWT` becomes an accepted value, and the failure narrows to a `JWT` route whose `AuthorizerId` did not resolve to a deployed authorizer.

Builds on the route key and stage matching issue, for route selection, and on the `AWS::ApiGatewayV2` CloudFormation issue, for the `Route` and `Authorizer` property parsers. The CDK Cognito CloudFormation issue is needed only by the `.loc.test.ts`, which deploys the pool and its client from one stack. The behaviour work does not wait on it.

Depends on #297, #299 and #301.

hughgrigg:
A JWT authorizer attached to a route verifies a signed token in process, against the keys its issuer publishes, before the integration is invoked. Verification reads the simulation's clock, so advancing past a token's exp turns a 200 into a 401.

Adds CreateAuthorizer, GetAuthorizers and DeleteAuthorizer for JWT authorizers, AuthorizationType JWT with AuthorizerId and AuthorizationScopes on a route, AWS::ApiGatewayV2::Authorizer, and a JWT parse and RS256 verification module in src/util/jwt.

Resolves https://github.com/KensioSoftware/yulin/issues/302

## Summary by CodeRabbit

* **New Features**
  * Added JWT authorization for simulated API Gateway HTTP APIs, including Cognito token verification, issuer and audience validation, expiration checks, scopes, identity sources, and handler-visible claims.
  * Added authorizer creation, listing, deletion, and route association through API Gateway APIs, SDK commands, CloudFormation, and CDK.
  * Added JWT-protected API examples and support for standard authorization responses, including 401 and 403 errors.

* **Documentation**
  * Expanded API Gateway documentation to describe JWT authorizers, supported configuration, limitations, validation, and authorization behavior.

* **Tests**
  * Added comprehensive coverage for JWT parsing, verification, route scopes, CloudFormation, CDK, SDK commands, and token expiration.


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #313, opened 2026-08-03
feat: authorize simulated HTTP API routes with AWS_IAM and SigV4-sign…

### LINKED ISSUE #303: Authorize simulated HTTP API routes with AWS_IAM and SigV4-signed requests
## Problem

Every simulated HTTP API route is either open or JWT-protected. A route guarded by IAM cannot be covered by a test here. Real callers of such a route sign their requests with SigV4, and an identity policy grants them `execute-api:Invoke`.

There is a CloudFormation failure alongside it. CDK's `HttpIamAuthorizer` emits only `AuthorizationType: AWS_IAM` on the `AWS::ApiGatewayV2::Route`. A synth against the `aws-cdk-lib` 2.262.1 installed in this repo produced no `AWS::ApiGatewayV2::Authorizer` resource and no `AuthorizerId`. The route property parser accepts only `NONE`, so a stack using that authorizer does not deploy.

## Desired behaviour

A route declared `AuthorizationType: AWS_IAM` reaches its integration only when the caller is allowed `execute-api:Invoke` on the ARN of the route being invoked. A refusal is 403 and the integration is not invoked. An allowed request carries the caller's identity into the payload 2.0 event as `requestContext.authorizer.iam`.

Most of the machinery exists. SigV4 verification and principal resolution run at the shared serving boundary before any controller sees a request. `simIamWildcardMatch` already matches the `execute-api` ARN shape, including multi-segment path wildcards. The new work is a route ARN built from the routed request, an authorization call against it, and the event and refusal shaping around that.

## Example

```typescript
  new CreateRouteCommand({
    ApiId,
    RouteKey: "GET /orders/{orderId}",
    Target: `integrations/${IntegrationId}`,
    AuthorizationType: "AWS_IAM",
  }),
);

await simAws.iam().putRolePolicy(
  new PutRolePolicyCommand({
    RoleName: "Reporter",
    PolicyName: "InvokeOrders",
    PolicyDocument: JSON.stringify({
      Version: "2012-10-17",
      Statement: [
        {
          Effect: "Allow",
          Action: "execute-api:Invoke",
          Resource: `arn:aws:execute-api:eu-west-2:111111111111:${ApiId}/$default/GET/orders/*`,
        },
      ],
    }),
  }),
);

const url = srv.localUrl(
  `https://${ApiId}.execute-api.eu-west-2.amazonaws.com/orders/42`,
);

const anonymous = await fetch(url);
// 403 {"message":"Forbidden"}. An unsigned request resolves to an anonymous
// caller, nothing allows it execute-api:Invoke, and the handler is not run.

const reporter = await fetch(url, {
  headers: { "x-sim-aws-caller": "arn:aws:iam::111111111111:role/Reporter" },
});
// 200 with the handler's response.
```

A SigV4-signed request from the same role behaves the same way. The caller header is the shorthand documented for the other served services, and the signed path is what the tests exercise.

## Proposed scope

- `CreateRoute` accepting `AuthorizationType: AWS_IAM`, reported back by `GetRoutes`. `AWS_IAM` takes no authorizer, so an `AuthorizerId` supplied alongside it is a `BadRequestException`.
- `AuthorizationScopes` alongside `AWS_IAM` refused with a `BadRequestException`. This is stricter than AWS, which documents route scopes as meaningful only for `JWT` and ignored for `AWS_IAM`. Accepting them here would let a test assert on a scope restriction that is not applied.
- `AWS_IAM` added to the `AuthorizationType` values the `AWS::ApiGatewayV2::Route` property parser accepts, so a CDK stack using `HttpIamAuthorizer` deploys and serves. This amends the CloudFormation issue's criterion that any `AuthorizationType` other than `NONE` fails the stack.
- A route ARN builder giving `arn:aws:execute-api:<region>:<account>:<apiId>/<stage>/<METHOD>/<path>`. Region and account come from the API's owning scope rather than the caller's. `<stage>` is the matched stage's name, so `$default` for the default stage.
- The builder takes its method and path from its caller, and this issue passes the request's own method in uppercase and the request path with the stage prefix and the leading slash removed. A request to the API root therefore gives an ARN ending `/GET/`. The Lambda invoke permission issue passes the route key template instead, and the two callers are reconciled in the notes below.
- A decision-returning authorizer under a new `src/service/apigatewayv2/serve/auth/`, evaluating `execute-api:Invoke` against that ARN with the API account's own IAM. No resource policies are passed, because HTTP APIs do not have them.
- `requestContext.authorizer.iam` on the payload 2.0 event, in the shape the Function URL path already produces.
- `Policy` on `AWS::ApiGatewayV2::Api` refused with its own message saying HTTP APIs do not support resource policies. The generic allow-list refusal reports a property as not simulated, which reads as a gap that will be filled later. There is no such property on the real resource, so the only template carrying one was copied from a REST API.
- Docs section on IAM-authorized routes with the ARN shape written out, plus the cross-account rule, the wildcard divergence and the unpopulated `iam` fields under Limitations.

## Acceptance criteria

- [ ] `CreateRoute` accepts `AuthorizationType: AWS_IAM`, and refuses an `AuthorizerId` or `AuthorizationScopes` supplied alongside it with a `BadRequestException` naming the property
- [ ] A CDK stack using `HttpIamAuthorizer` deploys through `AWS::ApiGatewayV2::Route`, and its route then requires IAM authorization when served
- [ ] The route ARN is built from the API's own account and region, the matched stage name, the request's own method rather than `ANY`, and the request path with the stage prefix and leading slash removed
- [ ] Identity policies naming `<apiId>/*`, `<apiId>/$default/*` and `<apiId>/*/GET/orders/*` each allow a request to `GET /orders/42` on the `$default` stage
- [ ] A policy naming a different stage does not allow that request, and neither does one naming a lowercase method
- [ ] An unsigned request to an `AWS_IAM` route gets `403 {"message":"Forbidden"}` and the integration is not invoked
- [ ] A SigV4-signed request from a principal with no matching Allow gets 403, an explicit `Deny` beats an `Allow`, and an allowed principal gets the handler's response
- [ ] `requestContext.authorizer.iam` reports `accountId` and `userArn` from the resolved principal's ARN for an allowed caller
- [ ] A caller in another simulated account is denied even with a full identity policy Allow, and the same caller succeeds after assuming a role in the API's account through simulated STS
- [ ] A route with `AuthorizationType: NONE` on the same API still serves both signed and unsigned requests

## Out of scope

- Lambda `REQUEST` authorizers, which stay refused by name and are deferred beyond this milestone.
- HTTP API resource policies. AWS does not support them, so there is nothing to build beyond the refusal above.
- `execute-api:ManageConnections`, which is WebSocket-only. Nothing constrains the action string in simulated IAM, so it belongs in Limitations rather than being enforced.
- Anything already covered by the Cognito CloudFormation, HTTP API foundation, route matching, Lambda invoke permission, CloudFormation resource and JWT authorizer issues in this milestone.

## Implementation notes

`SimApiGwRouteAuthorization` is the polymorphic role introduced by the HTTP API foundation issue with a `None` implementation. This issue adds a further implementation of it. The controller is not edited, and nothing here depends on the JWT authorizer issue landing first.

The authorizer follows `src/service/lambda/serve/auth/sim-lambda-url-authorizer.ts`, which returns a `SimIamAuthorizationDecision` rather than throwing, because a refusal is an ordinary HTTP outcome. IAM comes from the API's owning scope, the way `SimLambdaUrlRouter.route` resolves the account from the registry and then hands on `scope.iam()` (`src/service/lambda/serve/sim-lambda-url-router.ts:62,77`). `resourcePolicies` is omitted.

Simulated IAM needs no change. `simIamWildcardMatch` in `src/service/iam/authorize/sim-iam-wildcard.ts` converts `*` to `.*`, which crosses `/` boundaries. A research probe ran every wildcard shape an `execute-api` policy uses and all of them matched, including `arn:aws:execute-api:*:*:abc123/*/GET/*` against a two-segment path. Resource matching is case-sensitive (`src/service/iam/authorize/match/sim-iam-policy-statement-matcher.ts:106`), so a policy naming `prod/get/*` does not match a `GET` request. The criteria above pin both properties so a regression in the matcher surfaces here. One divergence goes in Limitations: AWS distinguishes `*` from `*/*` in some ARN path positions, while the simulator's `*` is uniformly greedy, which is more permissive than AWS.

AWS documents the ARN as `arn:aws:execute-api:{region}:{account-id}:{api-id}/{stage-name}/{HTTP-VERB}/{resource-path-specifier}`. Two details are inferred rather than documented: that the path segment carries no leading slash, and that under a named stage it is the path after the stage prefix rather than the full request path. Comment both at the point of construction.

The Lambda invoke permission issue builds the same ARN string for `AWS:SourceArn`, and the two callers differ in what they pass for the method and path. That one is matched against a pattern CDK writes into a permission, so it passes the route key template with braces intact and collapses the `$default` route to the literal `$default`. This one is matched against a resource a human writes in an identity policy, so it passes the concrete request method and path. Share one builder between them and record the pair as a single named divergence in Limitations, since AWS documents one format.

A malformed or wrongly scoped signature never reaches this code. `SimAwsRequestAuthenticator` refuses it at the boundary and `SimAwsAuthFailureResponse` answers `403 {"Message":"Forbidden"}` with an `x-amzn-errortype` header, or 400 for a signature too incomplete to parse. Real API Gateway is observed, not documented, to answer a missing or bad signature with a 403 carrying an `x-amzn-ErrorType` of `IncompleteSignatureException` or `InvalidSignatureException`. The simulator resolves an unsigned request to an anonymous caller instead, so such a request is denied by the route authorizer and gets `403 {"message":"Forbidden"}`. Both that and the capital-M body from the boundary go in Limitations.

The boundary also checks the signature's credential scope against `simAwsServiceSigningName` (`src/serve/controller/sim-aws-service-signing-name.ts`), which returns `execute-api` from the foundation issue. A request signed for another service is refused before route matching, including on a `NONE` route, where real API Gateway would ignore the signature. That divergence already exists for Function URLs and should be stated rather than fixed here.

Cross-account invocation is denied. `SimIamCrossAccountCaller` applies `SimIamBothSidesAllowRequirement` (`src/service/iam/authorize/caller-account/sim-iam-cross-account-caller.ts:37-38`), which needs a resource-side Allow, and an HTTP API has nowhere to put one. Real AWS behaves the same way, so the route through it is to assume a role in the API's account and sign with the session credentials. Cover that with a test rather than only a Limitations line.

`requestContext.authorizer.iam` is `{ accessKey, accountId, callerId, cognitoIdentity, principalOrgId, userArn, userId }`, observed rather than documented, matching the AWS `aws-lambda-java-events` type. Use the block the Function URL path already produces so both services report the same field with the same semantics: `accessKey` empty because the simulator does not carry the signing access key id past authentication, `callerId` and `userId` carrying the caller ARN because there is no `AIDA`/`AROA` unique-id namespace at the request boundary, and `cognitoIdentity` and `principalOrgId` null. `accountId` and `userArn` come from the resolved principal's ARN via `parseSimArn`. The reasoning is already written out on `SimLambdaFunctionUrlIamAuthorizer` in `src/service/lambda/serve/event/sim-lambda-url-event.type.ts:50-71`, and the values are produced by `SimLambdaUrlIamCaller` in `src/service/lambda/serve/event/sim-lambda-url-iam-authorizer.ts`. That file is not in the foundation issue's move list, so move it into the shared home here if that issue has not. All of it goes in Limitations. The shared event type already carries the `iam` arm of the `authorizer` field, so nothing is widened here; the JWT issue adds the `jwt` arm.

The two research reports disagree on whether `execute-api:InvalidateCache` applies to HTTP APIs. Nothing constrains the action string in simulated IAM either way, so state in Limitations that any `execute-api` action can be written in a policy and only `execute-api:Invoke` is evaluated.

`test/sigv4/sign-aws-request.ts` already takes `service` and `region` as parameters, so the signed-request tests need no new harness. `SimAwsResponseHints.stampCaller` puts `x-sim-aws-caller` and `x-sim-aws-auth` on every response, so a test can assert whether the simulator authenticated by signature or by header.

The work is roughly 1400 to 1800 insertions across 18 to 24 files, which is at the narrow end of the repo's measured feature PRs, because the SigV4, principal resolution and wildcard matching work already exists.

Builds on the route key, path parameter and stage matching issue, and on the `AWS::ApiGatewayV2` CloudFormation issue.

Depends on #299 and #301.

hughgrigg:
…ed requests

A route declared AuthorizationType AWS_IAM now reaches its integration only when the caller is allowed execute-api:Invoke on the ARN of the route being called, built from the API's own Account and Region, the stage that served the request, and the request's own method and path. A refusal is 403 and the integration is not invoked. An admitted caller reaches the handler as requestContext.authorizer.iam.

CDK's HttpIamAuthorizer emits only AuthorizationType on the Route, so the property parser accepts it on its own. AWS::ApiGatewayV2::Api refuses Policy with its own message, since an HTTP API has no resource policy.

Resolves https://github.com/KensioSoftware/yulin/issues/303

## Summary by CodeRabbit

* **New Features**
  * Added IAM authorization support for simulated API Gateway HTTP API routes.
  * Supports signed requests, caller identity propagation, route-specific permissions, and cross-account authorization behavior.
  * Added IAM-protected HTTP API examples and CDK deployment coverage.
* **Bug Fixes**
  * Improved authorization responses, including consistent handling of denied and unauthorized requests.
* **Documentation**
  * Documented IAM routes, permissions, ARN matching, caller details, limitations, and unsupported resource policies.

### DIFF
--- a/docs/services/apigatewayv2/README.md
+++ b/docs/services/apigatewayv2/README.md
@@ -636,6 +636,205 @@ claim. None of that is published by AWS; all of it is what the real endpoint was
 A route with `AuthorizationType: "NONE"` has no caller to describe, so `requestContext.authorizer` is
 left out of its events entirely.
 
+## Protecting a route with IAM
+
+A route declared `AuthorizationType: "AWS_IAM"` reaches its integration only when the caller is
+allowed `execute-api:Invoke` on the ARN of the route being called. The route takes no authorizer, and
+naming one is refused: IAM itself is what decides.
+
+The caller comes from the request, through either a SigV4 signature or an `x-sim-aws-caller` header
+naming a principal directly. A request that offers neither is anonymous, owns no policies, and is
+refused. See [callers of HTTP requests](../iam/#callers-of-http-requests) in the IAM docs for how that
+resolution works and how to sign a served request.
+
+The ARN a request is authorized against is:
+
+```text
+arn:aws:execute-api:<region>:<account>:<apiId>/<stage>/<METHOD>/<path>
+```
+
+- The Account and Region are the API's own, not the caller's.
+- The stage is the one that served the request, so `$default` for the default stage.
+- The method is the one the client sent, upper case, so a `GET` reaching a route keyed `ANY /orders`
+  gives `GET`.
+- The path is the request path with the stage segment and the leading slash taken off, so `/dev/orders/42`
+  served from stage `dev` gives `orders/42`. It is the path asked for rather than the route key, because
+  this is the resource a policy is written against by hand. A request to the API root gives an ARN
+  ending `/GET/`.
+
+An identity policy may wildcard any part of that. `<apiId>/*`, `<apiId>/$default/*` and
+`<apiId>/*/GET/orders/*` all allow a `GET` of `/orders/42` on the default stage.
+
+```typescript sim-apigatewayv2-iam-authorizer
+/**
+ * Protecting a simulated HTTP API route with IAM.
+ */
+
+import {
+  CreateApiCommand,
+  CreateIntegrationCommand,
+  CreateRouteCommand,
+  CreateStageCommand,
+} from "@aws-sdk/client-apigatewayv2";
+import { CreateRoleCommand, PutRolePolicyCommand } from "@aws-sdk/client-iam";
+import {
+  AddPermissionCommand,
+  CreateFunctionCommand,
+} from "@aws-sdk/client-lambda";
+
+import { SimAws } from "@kensio/yulin";
+import type { SimPayload2Event } from "@kensio/yulin/apigatewayv2";
+import { makeLambdaZipFileInput } from "@kensio/yulin/lambda";
+import { serveSimAws } from "@kensio/yulin/serve";
+
+const simAws = new SimAws();
+
+const { FunctionArn } = await simAws.lambda().createFunction(
+  new CreateFunctionCommand({
+    FunctionName: "orders",
+    Role: "arn:aws:iam::888888888888:role/OrdersRole",
+    Code: {
+      ZipFile: makeLambdaZipFileInput((event: SimPayload2Event) => ({
+        statusCode: 200,
+        headers: { "content-type": "text/plain" },
+        body: `orders for ${
+          event.requestContext.authorizer?.iam?.userArn ?? "nobody"
+        }`,
+      })),
+    },
+  }),
+);
+
+const apiGateway = simAws.apiGatewayV2();
+
+const { ApiId, ApiEndpoint } = await apiGateway.createApi(
+  new CreateApiCommand({ Name: "orders", ProtocolType: "HTTP" }),
+);
+
+const { IntegrationId } = await apiGateway.createIntegration(
+  new CreateIntegrationCommand({
+    ApiId,
+    IntegrationType: "AWS_PROXY",
+    IntegrationUri: FunctionArn,
+    PayloadFormatVersion: "2.0",
+  }),
+);
+
+await apiGateway.createRoute(
+  new CreateRouteCommand({
+    ApiId,
+    RouteKey: "GET /orders/{orderId}",
+    Target: `integrations/${IntegrationId}`,
+    AuthorizationType: "AWS_IAM",
+  }),
+);
+
+await apiGateway.createStage(
+  new CreateStageCommand({ ApiId, StageName: "$default", AutoDeploy: true }),
+);
+
+await simAws.lambda().addPermission(
+  new AddPermissionCommand({
+    FunctionName: "orders",
+    StatementId: "api-gateway-invoke",
+    Action: "lambda:InvokeFunction",
+    Principal: "apigateway.amazonaws.com",
+    SourceArn: `arn:aws:execute-api:us-east-1:888888888888:${ApiId}/*/*`,
+  }),
+);
+
+// A Role of the API's own Account, allowed to call the orders routes of this
+// API on the default stage.
+await simAws.iam().createRole(
+  new CreateRoleCommand({
+    RoleName: "Reporter",
+    AssumeRolePolicyDocument: JSON.stringify({
+      Version: "2012-10-17",
+      Statement: [
+        {
+          Effect: "Allow",
+          Principal: { AWS: "arn:aws:iam::888888888888:root" },
+          Action: "sts:AssumeRole",
+        },
+      ],
+    }),
+  }),
+);
+
+await simAws.iam().putRolePolicy(
+  new PutRolePolicyCommand({
+    RoleName: "Reporter",
+    PolicyName: "InvokeOrders",
+    PolicyDocument: JSON.stringify({
+      Version: "2012-10-17",
+      Statement: [
+        {
+          Effect: "Allow",
+          Action: "execute-api:Invoke",
+          Resource: `arn:aws:execute-api:us-east-1:888888888888:${ApiId}/$default/GET/orders/*`,
+        },
+      ],
+    }),
+  }),
+);
+
+const srv = await serveSimAws({ simAws });
+const url = srv.localUrl(`${ApiEndpoint}/orders/42`);
+
+const anonymous = await fetch(url);
+
+console.log(anonymous.status); // 403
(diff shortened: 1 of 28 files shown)


hughgrigg:
@coderabbitai full review

PR #315, opened 2026-08-03
feat: create a simulated HTTP API from an OpenAPI 3 definition

### LINKED ISSUE #304: Create a simulated HTTP API from an OpenAPI 3 definition
## Problem

Yulin has no way to accept an OpenAPI document. There is no `ImportApi` command, and the CloudFormation issue in this milestone refuses `AWS::ApiGatewayV2::Api` `Body` by name.

An HTTP API is often declared as an OpenAPI 3 document rather than as separate route and integration resources. A test that wants those routes has to restate the document as `CreateRoute` and `CreateIntegration` calls first.

## Desired behaviour

`ImportApi` takes a serialised OpenAPI 3.0 document. It creates the API, one route and one integration per operation, and one JWT authorizer per referenced security scheme. `AWS::ApiGatewayV2::Api` `Body` carries the same document inline as a JSON object and reaches the same translator through the same command.

Most members of an OpenAPI document are ignored. AWS classifies `requestBody`, the content schemas under `responses` and `components.schemas` as valid OpenAPI that HTTP APIs do not support, and ignores them silently, because HTTP APIs perform no request validation. This does the same.

`FailOnWarnings: true` is accepted on both entry points. `FailOnWarnings: false` is refused.

An import creates no stage. `AWS::ApiGatewayV2::Stage` or `CreateStage` is still declared separately, and an imported API with no stage answers 404.

## Example

```typescript
const openApi = {
  openapi: "3.0.1",
  info: { title: "orders", version: "1.0" },
  paths: {
    "/orders/{orderId}": {
      get: {
        security: [{ "pool-authorizer": ["aws.cognito.signin.user.admin"] }],
        // Ignored, as on AWS: HTTP APIs do no request validation.
        responses: { "200": { description: "200 response" } },
        "x-amazon-apigateway-integration": {
          type: "aws_proxy",
          httpMethod: "POST",
          uri: `arn:aws:apigateway:eu-west-2:lambda:path/2015-03-31/functions/${functionArn}/invocations`,
          payloadFormatVersion: "2.0",
        },
      },
    },
  },
  components: {
    securitySchemes: {
      "pool-authorizer": {
        type: "oauth2",
        "x-amazon-apigateway-authorizer": {
          type: "jwt",
          jwtConfiguration: {
            issuer: `https://cognito-idp.eu-west-2.amazonaws.com/${userPoolId}`,
            audience: [clientId],
          },
          identitySource: "$request.header.Authorization",
        },
      },
    },
  },
};

const { ApiId, ApiEndpoint } = await simAws
  .apiGatewayV2()
  .importApi(new ImportApiCommand({ Body: JSON.stringify(openApi) }));

// The import created a route, an integration and an authorizer, but no stage.
await simAws
  .apiGatewayV2()
  .createStage(new CreateStageCommand({ ApiId, StageName: "$default", AutoDeploy: true }));

try {
  const response = await fetch(srv.localUrl(`${ApiEndpoint}/orders/42`), {
    headers: { authorization: `Bearer ${accessToken}` },
  });
  // 200, routeKey GET /orders/{orderId}, pathParameters { orderId: "42" }
} finally {
  srv.close();
}
```

## Proposed scope

- An `ImportApi` command. `Body` is a string and is the only required member of the SDK input. `FailOnWarnings` is accepted when `true`. The handler parses `Body` before anything else looks at it.
- `Api.Body` and `Api.FailOnWarnings` added to the `AWS::ApiGatewayV2::Api` allow-list, where CloudFormation carries the document as an inline JSON object rather than a string. The creator serialises the resolved property and calls `importApi`, so both entry points share one translator and one set of refusals. That follows the rule that CloudFormation orchestrates and services create.
- `Name` and `ProtocolType` become optional on `AWS::ApiGatewayV2::Api` when `Body` is present, which is what AWS documents. `ProtocolType` alongside a `Body` must be `HTTP`. `Name` alongside a `Body` names the API, and `info.title` names it otherwise.
- A translator taking an already parsed document. The two entry points differ only in how the document arrives, and both have parsed it by the time the translator runs.
- `openapi: "3.0.x"` required. `swagger` and `3.1.x` refused, naming the version found.
- One route per operation, with the route key built as the uppercased method key and the path taken verbatim. OpenAPI path templating is already API Gateway's path parameter syntax, so `/orders/{orderId}` needs no translation.
- The operation keys read are `get`, `put`, `post`, `delete`, `options`, `head` and `patch`. Other path item members are handled explicitly rather than uppercased into a route key.
- Route keys, integration inputs and JWT configurations go through the ordinary `CreateRoute`, `CreateIntegration` and `CreateAuthorizer` commands, so nothing about route key grammar, payload format or issuer validation is restated in the translator.
- `x-amazon-apigateway-integration` read into a `CreateIntegration` input, including the `$ref` form into `components.x-amazon-apigateway-integrations`. A referenced definition is created once and shared by every operation that references it.
- The `IntegrationUri` normalisation the CloudFormation issue put in its property parser moves into the integration model, so the translator applies the same one. An imported `uri` is the long `arn:<partition>:apigateway:<region>:lambda:path/2015-03-31/functions/<function-arn>/invocations` form.
- `components.securitySchemes` carrying `x-amazon-apigateway-authorizer` with `type: "jwt"` read into a `CreateAuthorizer` input, whose required `Name` is the security scheme key. One authorizer per scheme name, shared by every operation that names it.
- A route created for an operation carrying a security requirement gets `AuthorizationType: "JWT"`, the shared authorizer's `AuthorizerId`, and the requirement's scope list as its `AuthorizationScopes`. An operation with no `security` gets `AuthorizationType: "NONE"`.
- A JSON pointer on every refusal, so the reader can find the member in their own document.

## Acceptance criteria

- [ ] An imported document creates the API, one route per operation with `routeKey` of the form `GET /orders/{orderId}`, one integration per `x-amazon-apigateway-integration`, and returns `ApiId` and `ApiEndpoint`
- [ ] `requestBody`, the content schemas under `responses` and `components.schemas` are ignored without a refusal, and a request whose body contradicts a declared schema still reaches the handler
- [ ] A document declaring `swagger: "2.0"` or `openapi: "3.1.0"` is refused, naming the version, and no API is created
- [ ] An `aws_proxy` integration whose `httpMethod` is not `POST` is refused, and `integrationMethod` is refused rather than read
- [ ] A missing `payloadFormatVersion` is refused, `"1.0"` is refused, and a `type` other than `aws_proxy` is refused naming `http_proxy` as deferred
- [ ] A `$ref` into `components.x-amazon-apigateway-integrations` resolves and two operations referencing one definition share a single integration, while a `$ref` anywhere else is refused naming its pointer
- [ ] A security scheme's comma-separated `identitySource` becomes the single-entry array `CreateAuthorizer` takes, and a value carrying more than one entry is refused
- [ ] A protected operation gives 401 without a token and reaches the handler with a Cognito access token, and its security requirement's scope list arrives as the route's `AuthorizationScopes`
- [ ] An operation carrying more than one security requirement is refused, matching AWS, and so is a security requirement naming a scheme `components.securitySchemes` does not define
- [ ] A `trace` operation key, a path item `$ref` and `x-amazon-apigateway-any-method` are each refused, while `parameters`, `summary`, `description` and `tags` are skipped without a refusal
- [ ] `FailOnWarnings: false` is refused by name on both entry points, `Basepath` and `BasePath` are refused, and `ReimportApi` is refused
- [ ] An imported API with no stage answers 404, and serves once a stage is created separately
- [ ] `Api.Body` deploys an equivalent template through the same translator, with `Name` and `ProtocolType` both absent, and again with `ProtocolType: HTTP` present
- [ ] `Api.BodyS3Location` stays refused, and a template combining `Api.Body` with a `Route`, `Integration` or `Authorizer` resource for the same `ApiId` fails the stack naming both logical IDs
- [ ] Every refusal names the JSON pointer of the offending member, such as `#/paths/~1orders~1{orderId}/get/x-amazon-apigateway-integration/payloadFormatVersion`
- [ ] Docs section on importing an OpenAPI definition, with the ignored members and the always-on `FailOnWarnings` behaviour under Limitations

## Out of scope

- `Api.BodyS3Location`. Reading a document out of a simulated S3 bucket adds a fetch path and nothing about OpenAPI. It stays refused by name after this issue.
- `ReimportApi`. Nothing in this milestone changes an API in place, and the same reasoning that keeps `UpdateRoute` and `UpdateStage` out applies here.
- `Basepath` on `ImportApi`, `BasePath` on `AWS::ApiGatewayV2::Api`, and `servers` in the document. All refused by name. Base path handling changes the path every route matches on, which belongs with custom domain names beyond this milestone.
- A security scheme of type `openIdConnect`, refused by name. AWS auto-detects the issuer by fetching the discovery document at `openIdConnectUrl`. The JWT authorizer issue in this milestone deliberately never fetches discovery or JWKS over HTTP, and taking the URL itself as the issuer would produce a mismatched `iss` and a silent 401. Only `oauth2` with an explicit issuer is accepted.
- `operationId`, which AWS maps to the route's `OperationName`. No command in this milestone takes `OperationName` and nothing reads it, so an imported `operationId` is dropped rather than stored. Recorded in Limitations.
- `http_proxy` integrations, `integrationSubtype`, `requestParameters`, `credentials`, `tlsConfig`, `responseTransferMode`, and private integrations through `connectionId` or `connectionType: VPC_LINK`. All refused by name, all deferred beyond this milestone, matching the refusals the other issues in this milestone already carry.
- `x-amazon-apigateway-cors`, refused by name, alongside the `Api.CorsConfiguration` refusal the CloudFormation issue records. CORS request handling is deferred beyond this milestone.
- Lambda `REQUEST` authorizers expressed as `x-amazon-apigateway-authorizer` with `type: "request"` or `"token"`, refused by name and deferred beyond this milestone.
- `AWS_IAM` authorization expressed through a security scheme. The research does not establish how an HTTP API document declares it, so a security scheme that is not a JWT authorizer is refused rather than guessed at, including after the `AWS_IAM` route authorization issue in this milestone lands.
- Document-level `security` applied to every operation, and `x-amazon-apigateway-any-method`. Neither is established by the research for HTTP APIs, so both are refused by name.
- Quick create through `Api.Target` and `Api.RouteKey`, which the CloudFormation issue already refuses and which this issue does not change.
- Request validation, `AWS::ApiGatewayV2::Model` and any use of `components.schemas`. HTTP APIs do not validate requests, so there is nothing to build.

## Implementation notes

AWS sorts import problems into three categories. Error is malformed input, and nothing is created. Warning is invalid per the specification but non-blocking, and `FailOnWarnings` decides whether the import rolls back. Info is valid OpenAPI that HTTP APIs do not support, and it is ignored silently.

That classification is what keeps this issue small. The Info category covers `requestBody`, the content schemas under `responses` and `components.schemas`, so there is no JSON Schema to understand and no `$ref` to resolve across schema trees. Real documents carry schemas, so ignoring them is what AWS does and what makes the feature usable here. The one `$ref` form that must resolve is the single-level local reference into `components.x-amazon-apigateway-integrations`.

Refusing rather than warning behaves as `FailOnWarnings: true` at all times. That is the fail-closed reading, so it is the one taken here. `FailOnWarnings: true` is accepted and has no further effect, and `FailOnWarnings: false` is refused by name because the lenient behaviour it selects is not simulated. Omitting it is accepted. The research does not establish AWS's own default for the property, so the docs should say the property is honoured only in its strict sense rather than name a default.

The translator lives in a new `src/service/apigatewayv2/openapi/`, and the command in `src/service/apigatewayv2/command/import-api/` following the `*.cmd.ts` and `*.handler.ts` split the repo uses. The CloudFormation side is a `Body` branch on the `Api` creator added by the CloudFormation issue in `src/service/apigatewayv2/cfn/`, calling `importApi` rather than reaching into the model. The precedent is `SimCfnCognitoUserPoolCreator` at `src/service/cognito/cfn/user-pool/sim-cfn-cognito-user-pool-creator.ts`, which creates through `createUserPool` so that a pool a template deployed is the same thing an SDK caller would have got. The creator serialises the resolved `Body` to a string, `importApi` parses it, and the translator takes the parsed document. Serialising is lossless, because `SimCfnTemplateValue` in `src/service/cloudformation/template/value/sim-cfn-template-value.ts` is null, boolean, number, string, arrays and records, all of which are JSON.

A refusal has to fail the stack rather than skip the resource. `SimCfnResourceCreateOperation.isUnsupportedResourceError` at `src/service/cloudformation/resource/create/sim-cfn-resource-create-operation.ts:78` treats any error whose message contains both `Unsupported sim` and `CloudFormation` as a skip, so the import errors must not be worded that way. A `BadRequestException` on the service's own error type propagates and fails the stack, which is what the `Api.Body` case wants. `BadRequestException` is also what the real `ImportApi` reports a bad document with. Nothing in `src/` defines it today, so this issue relies on the service error module the earlier issues in this milestone create rather than adding it.

`ImportApi` has no api id at call time, so it authorizes on the same control-plane resource `CreateApi` uses rather than on `arn:aws:apigateway:<region>::/apis/<api-id>`.

The decomposition matters more here than usual. A single document reader will not survive `max-lines` at 300 (`eslint.config.ts:157`), `complexity` at 20 (`eslint.config.ts:160`) or `scripts/sh/fta.sh` at 50, and the CLAUDE.md convention against ternaries pushes the same way. Proposed pieces, one responsibility each:

- `SimApiGwOpenApiDocument` for the root and the version check.
- `SimApiGwOpenApiPointer` for RFC 6901 pointers, escaping `~` as `~0` and `/` as `~1`, built up as the readers descend.
- `SimApiGwOpenApiValue` for the narrowing helpers that throw pointer-named errors, mirroring `SimCfnCognitoValueParser` at `src/service/cognito/cfn/sim-cfn-cognito-value-parser.ts`.
- `SimApiGwOpenApiPaths` and `SimApiGwOpenApiPathItem` for the path map and the keys within one path.
- `SimApiGwOpenApiOperation` for one operation's integration and security requirement.
- `SimApiGwOpenApiIntegration` for the extension object.
- `SimApiGwOpenApiIntegrationRefs` for the single-level local reference and the created-once integration ids.
- `SimApiGwOpenApiSecuritySchemes` for the scheme to authorizer mapping and its own created-once ids.
- An importer that drives the commands in order.

Route key assembly needs the uppercasing done in the translator. OpenAPI method keys are lowercase, and `CreateRoute` refuses a lowercase method, which the route matching issue in this milestone pins as an acceptance criterion. Nothing else about the path is touched, so `{orderId}` reaches the segment classes that issue introduces unchanged, and `SimApiGwPathParameters` populates the event as it already does.

Four behaviours here are chosen rather than established, and each belongs in Limitations. Whether AWS creates one integration per `$ref` use or shares a single one is not established by the research. This shares one, which is how a reusable integration definition reads, and what CDK's own deduplication does for the resource path. The observable difference is the count `GetIntegrations` returns. Second, which of `Name` and `info.title` names the API when both are present is not established. This picks `Name`, which affects only the name `GetApi` reports and nothing that routes. Third, `trace` is a valid OpenAPI operation key and AWS HTTP API route keys have no `TRACE` method, but the research does not pin that, so `trace` is refused rather than turned into a route key the route grammar would accept. Fourth, `parameters` is listed by the research as part of the supported subset, but HTTP APIs derive path parameters from the path template and validate nothing else, so it is skipped with no effect. A terminal `{proxy+}` in an imported path is passed through to `CreateRoute` unchanged, which is also chosen rather than established, since the research covers greedy segments for route keys and not for OpenAPI path templating.

This issue amends the CloudFormation issue in three places. `Api.Body` leaves its refusal list and its Out of scope entry. Its docs acceptance criterion covers the `Body` refusal and the `CorsConfiguration` refusal under Limitations, and the `Body` half of that entry is deleted. Its `Api` allow-list gains `Body` and `FailOnWarnings`, and its requirement of `Name` and `ProtocolType` narrows to the case where no `Body` is present.

The refusal surface is the bulk of the work. Counting the version checks, the integration fields, the security scheme fields, the pointer-named structural guards and the two entry points, there are roughly forty branches that each need a test at 99.9% statements and 95% branches. Budget several hundred test lines for that, and split the test files by concern rather than raising `max-lines`.

Around 1400 to 2000 insertions across 20 to 26 files.

Builds on the `AWS::ApiGatewayV2` CloudFormation issue, for the `Api` creator and its property parser, and on the Cognito JWT authorizer issue, for `CreateAuthorizer` and route scopes. It is the last issue in this milestone.

Depends on #301 and #302.

hughgrigg:
ImportApi takes a serialised OpenAPI 3.0 document and creates the API, one route and one integration per operation, and one JWT authorizer per security scheme an operation names. AWS::ApiGatewayV2::Api Body carries the same document inline and reaches the same translator through the same command.

Everything is created through the ordinary CreateRoute, CreateIntegration and CreateAuthorizer commands, and a refusal one of them makes is given the JSON pointer of the member that produced it.

Resolves https://github.com/KensioSoftware/yulin/issues/304

## Summary by CodeRabbit

* **New Features**
  * Added API Gateway HTTP API import from OpenAPI 3.0 documents.
  * Imports routes, Lambda integrations, and supported JWT authorizers, including reusable definitions.
  * Added CloudFormation support for inline OpenAPI API definitions through `Body`.
  * Added SDK support and an executable example for imported HTTP APIs.

* **Bug Fixes**
  * Improved handling of API Gateway-wrapped Lambda integration URIs.
  * Added rollback and clear validation errors for invalid or conflicting imports.

* **Documentation**
  * Documented supported OpenAPI features, limitations, warning behavior, and CloudFormation integration.


hughgrigg:
@coderabbitai full review

PR #326, opened 2026-08-03
feat: cache simulated HTTP API Lambda authorizer results

### LINKED ISSUE #323: Cache simulated HTTP API Lambda authorizer results for AuthorizerResultTtlInSeconds
## Problem

Real API Gateway caches a Lambda authorizer's answer for `AuthorizerResultTtlInSeconds`, keyed on the authorizer's identity source values, and does not invoke the authorizer again while that answer is cached. A session that is revoked part-way through a test still gets in until the TTL expires, and an authorizer counting its own invocations sees one call for several requests. Neither is visible without the cache being simulated, and both are the kind of thing a test written against the simulator would get wrong about the deployed system.

Caching is also not optional if a CDK-defined authorizer is ever to deploy. `HttpLambdaAuthorizer` sets `resultsCacheTtl` to `Duration.minutes(5)` when the caller does not, so every CDK stack emits a non-zero TTL.

## Current behaviour

`AuthorizerResultTtlInSeconds` falls outside the accepted option list in `SimHttpApiAuthorizerCommands`, so `CreateAuthorizer` refuses it by name. Nothing caches an authorization decision.

## Desired behaviour

`CreateAuthorizer` accepts `AuthorizerResultTtlInSeconds` on a `REQUEST` authorizer that has at least one identity source, which is what real API Gateway requires to enable caching. Two requests presenting the same identity source values within the TTL invoke the authorizer once, and the second request is served from the cached decision, including its `context`. Advancing simulated time past the TTL invokes the authorizer again.

The cache is keyed on the identity source values only, so by default one decision covers every route of the API that uses the authorizer. Adding `$context.routeKey` to the identity sources caches per route instead, as it does on AWS.

Expiry is checked against the injectable `SimClock`, so a test advances simulated time rather than waiting.

## Acceptance criteria

- [ ] Two requests with the same identity source values inside the TTL invoke the authorizer once and both get the same decision and `context`.
- [ ] Different identity source values invoke the authorizer separately.
- [ ] Advancing simulated time past the TTL invokes the authorizer again.
- [ ] A cached decision covers other routes of the API using the same authorizer, and adding `$context.routeKey` as an identity source separates them.
- [ ] `AuthorizerResultTtlInSeconds` with no identity source is refused by name, since real API Gateway cannot key the cache.
- [ ] A denial is cached the same way an allow is.

## Out of scope

Deploying a cached authorizer from a template, which is https://github.com/KensioSoftware/yulin/issues/324 and needs this one first. `$stageVariables.*` identity sources. Only `$context.routeKey` is added here, since it is the one AWS documents for per-route caching; the rest of `$context` needs the whole variable set and is a separate question.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/322, since there is nothing to cache until a Lambda `REQUEST` authorizer exists.

`SimHttpApiIdentitySource` reads a value out of a `Request` today, which is most of what a cache key needs. `$context.routeKey` is the first identity source that reads from the matched route rather than the request, so the identity source will need the match alongside the request.

hughgrigg:
CreateAuthorizer takes AuthorizerResultTtlInSeconds on a REQUEST authorizer, and a request presenting the same identity source values within it is served from the decision already made, context and all, without the function running again. Expiry is checked against the simulation's clock, so a test advances time rather than waiting.

The key is the identity source values and nothing else, so one decision covers every route of the API using the authorizer, and $context.routeKey as an identity source puts the route in the key. Each kind of identity source is now its own class, since that one reads the matched route rather than the request.

Resolves https://github.com/KensioSoftware/yulin/issues/323

## Summary by CodeRabbit

* **New Features**
  * Added configurable Lambda authorizer result caching for HTTP APIs.
  * Cached approvals and denials are reused until the configured TTL expires.
  * Supports header, query-string, and route-key identity sources.
  * Added route-specific caching through `$context.routeKey`.
  * Added validation for cache TTLs and identity-source requirements.
  * Added documentation and an executable example demonstrating cache reuse and expiration.
* **Bug Fixes**
  * Authorizer invocation failures are no longer cached and are retried as needed.


hughgrigg:
@coderabbitai full review

hughgrigg:
@coderabbitai full review

hughgrigg:
@coderabbitai full review

PR #348, opened 2026-08-03
feat: supply the source Account when a simulated service invokes a La…

### LINKED ISSUE #328: Supply the source Account when a simulated service invokes a Lambda function
## Problem

Nothing in the simulator gives `AWS:SourceAccount` a value at request time. `AddPermission` writes
it into a statement (`src/service/lambda/function/policy/sim-lambda-permission.ts:126-128`), nothing
supplies it, and the condition matcher fails closed on a missing key.
`docs/services/apigatewayv2/README.md` records the consequence:

> `AWS:SourceArn` is the only condition key supplied here. A permission that also carries
> `SourceAccount`, `PrincipalOrgID` or `InvokedViaFunctionUrl` never matches, since nothing gives
> those keys a value at request time, so the request is refused with the same 500.

That is about to stop being a footnote. CDK's `LambdaDestination` for an S3 event notification
always sets `sourceAccount`, so the permission every CDK-authored notification depends on could
never match:

```json
{
  "Action": "lambda:InvokeFunction",
  "FunctionName": { "Fn::GetAtt": ["Handler886CB40B", "Arn"] },
  "Principal": "s3.amazonaws.com",
  "SourceAccount": "111111111111",
  "SourceArn": { "Fn::GetAtt": ["SiteBucket397A1860", "Arn"] }
}
```

## Desired behaviour

A simulated AWS service invoking a Lambda function supplies both `AWS:SourceArn` and
`AWS:SourceAccount`, so a resource policy statement carrying either or both matches when the values
match and is refused when they do not.

## Proposed scope

- A new `src/service/lambda/command/authorize/sim-lambda-service-invoke-authorizer.ts`, taking the
  function, the service principal, an optional source ARN and an optional source Account, returning
  a `SimIamAuthorizationDecision` and reusing `simLambdaResourcePolicies`.
- `SimHttpApiInvokeAuthorizer` rewritten as a thin wrapper over it, supplying the API's own Account
  as `AWS:SourceAccount`.
- The divergence paragraph in `docs/services/apigatewayv2/README.md` removed, and the class comment
  at `sim-lambda-permission.ts:89-98` corrected, since it currently says those keys never match.

## Acceptance criteria

- [ ] A permission granted with a matching `SourceAccount` allows the invoke.
- [ ] A permission granted with a different `SourceAccount` refuses it.
- [ ] A permission granted with `SourceArn` alone behaves as it does today, so the existing API
      Gateway tests pass unedited.
- [ ] The API Gateway docs no longer claim `SourceAccount` never matches.

## Out of scope

Any S3 code. `PrincipalOrgID` and `InvokedViaFunctionUrl`, which still have no value at request time
and stay refused under the same fail-closed rule. Widening `InvokeCommandHandlerOptions` to carry a
condition context: real Lambda's Invoke API has no such parameter, and it would let a caller assert
its own source ARN.

## Implementation notes

Where this lives matters. "Who may invoke me" is Lambda's rule, so it belongs in the Lambda tree
with API Gateway and S3 both calling it. The alternative, S3 forking a second near-identical copy of
`SimHttpApiInvokeAuthorizer`, guarantees the two drift.

Worth doing before the S3 notification work rather than inside it, since it changes simulated API
Gateway behaviour and retires a documented divergence, and is easier to review on its own.

Needed by https://github.com/KensioSoftware/yulin/issues/331.

hughgrigg:
…mbda function

Who may invoke a function is Lambda's rule, so the decision moves into SimLambdaServiceInvokeAuthorizer, which every simulated service calls. API Gateway now supplies its own Account as AWS:SourceAccount alongside the execute-api source ARN, so a permission carrying either or both is evaluated on what it says.

Resolves https://github.com/KensioSoftware/yulin/issues/328

### DIFF
--- a/docs/services/apigatewayv2/README.md
+++ b/docs/services/apigatewayv2/README.md
@@ -181,8 +181,13 @@ arn:aws:execute-api:<region>:<account>:<apiId>/<stage>/<METHOD>/<route path>
 A `SourceArn` may wildcard any part of that, which is what the usual grant does:
 `<apiId>/*/*` allows every route of the API on every stage.
 
-`AWS:SourceArn` is the only condition key supplied here. A permission that also carries
-`SourceAccount`, `PrincipalOrgID` or `InvokedViaFunctionUrl` never matches, since nothing gives those
+The API's own Account is supplied as `AWS:SourceAccount`, so a permission carrying a `SourceAccount`
+matches when it names the Account the API belongs to. That is the Account the source ARN names, not
+the one owning the function, which matters for an integration reaching across Accounts. CDK writes
+both keys for some grants, and a permission carrying either or both is evaluated on what it says.
+
+`AWS:SourceArn` and `AWS:SourceAccount` are the only condition keys supplied here. A permission that
+also carries `PrincipalOrgID` or `InvokedViaFunctionUrl` never matches, since nothing gives those
 keys a value at request time, so the request is refused with the same 500.
 
 Neither the method nor the path is documented by AWS as the value API Gateway supplies. Both are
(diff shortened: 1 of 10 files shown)


hughgrigg:
@coderabbitai rate limit

PR #350, opened 2026-08-03
fix: apply a simulated TransactWriteItems atomically

### LINKED ISSUE #336: Apply a simulated TransactWriteItems atomically
## Problem

A simulated `TransactWriteItems` can apply partially. Real DynamoDB transactions are all or nothing.

`applySimDynamoDbTransaction` runs `assertApplies` and then applies in a loop
(`src/service/dynamodb/command/transact/sim-dynamodb-transact-cancellation.ts:36-41`):

```typescript
assertApplies(targets);

for (const { item, table } of targets) {
  item.applyTo(table);
}
```

`assertApplies` checks only the `ConditionExpression` of each action, through `reasonFor` into
`SimDynamoDbTransactPutAction.assertApplicableTo` (`sim-dynamodb-transact-put-action.ts:36-38`). Two
other pre-write assertions are not checked there and instead throw from inside `applyTo`:

- the secondary-index key-type assertion, `this.indexes.assertItemKeyTypes(...)` at
  `src/service/dynamodb/table/sim-dynamodb-table.ts:228`
- the 400 KB item size check in `SimDynamoDbItem.ofUpdatedAttributes`
  (`src/service/dynamodb/item/sim-dynamodb-item.ts:74-78`), reached through `plan.applyTo`

So a transaction whose third action carries a wrong-typed index key attribute, or whose update would
exceed 400 KB, writes the first two actions and then throws. The function's own doc comment says
"Nothing has been written by the time this throws", which is true only of the checks `assertApplies`
performs.

## Desired behaviour

A `TransactWriteItems` that fails for any reason leaves the table exactly as it was.

## Proposed scope

Either move every pre-write assertion into `assertApplicableTo` so `applyTo` cannot throw, or stage
the writes and commit only once all of them have succeeded. The first keeps the existing shape and
lets the failure be reported through the cancellation reasons a caller already reads; the second is
more robust against a future assertion being added in the wrong place. Pick one and say why in a
code comment, and correct the doc comment on `applySimDynamoDbTransaction` either way.

## Acceptance criteria

- [ ] A transaction whose second action carries a wrong-typed secondary-index key attribute leaves
      the table wholly unchanged.
- [ ] A transaction whose second action would exceed the 400 KB item size limit leaves the table
      wholly unchanged.
- [ ] The existing condition-failure behaviour is unchanged: it already fails before the first
      mutation and its cancellation reasons already name the failing action.

## Out of scope

Cancellation reason wording, which is already correct. Any change to condition-check behaviour.

## Implementation notes

Worth fixing before simulated DynamoDB Streams capture lands
(https://github.com/KensioSoftware/yulin/issues/339). Capture goes on the item store, so a partially
applied transaction stops being a latent state bug and starts emitting stream records for a
transaction real AWS rejects atomically. At that point the instinct is to paper over it in the
stream layer, which would be fixing the wrong thing.

hughgrigg:
A transaction could write its first actions and then throw, because the secondary index key type check and the 400 KB item size check both fired from inside applyTo rather than from assertApplicableTo.

Every action now prepares its write before any of them is committed. SimDynamoDbTableWrites splits a write into a prepare that can be refused and a commit that cannot, and putItem and deleteItem run the two together, so no write reaches the item store unprepared.

Resolves https://github.com/KensioSoftware/yulin/issues/336

## Summary by CodeRabbit

* **New Features**
  * DynamoDB transactions now validate all writes before committing, preserving atomicity when a later write is invalid.
  * Transactions reject invalid secondary-index key types and items exceeding the 400 KB size limit.
  * TTL scheduling now occurs only for successfully committed writes.
  * Transaction cancellation reasons are preserved while failed transactions leave existing data unchanged.
* **Documentation**
  * Updated DynamoDB documentation to describe transactional write preparation, atomic commits, validation, and cancellation behavior.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -2244,13 +2244,15 @@ the `ConditionalCheckFailedException` a single `PutItem` throws.
 An action that sets `ReturnValuesOnConditionCheckFailure` to `ALL_OLD` gets `Item` on its
 cancellation reason, holding the item as it was, so a retry needs no second read.
 
-Five things refuse the request outright rather than cancelling it, with nothing written either way:
+These refuse the request outright rather than cancelling it, with nothing written either way:
 
 - more than 100 actions
 - an action carrying more than one of `Put`, `Update`, `Delete` and `ConditionCheck`, or none of them
 - two actions on the same item of one table
 - a table that is not there, or a key that does not match its key schema
 - an update that would move the item's primary key
+- an item carrying a secondary index key attribute as a type the index did not declare
+- an update that would take the item past the 400 KB one item holds
 
 One table may be named as often as the transaction likes, which is the difference from a batch. What
 it may not do is touch one item twice.
--- a/src/service/dynamodb/README.md
+++ b/src/service/dynamodb/README.md
@@ -153,8 +153,16 @@ For a table with partition key `pk` and sort key `sk`, the internal key includes
 { "pk": "user-123", "sk": "profile" }
 ```
 
-This is an implementation detail, but it is important when changing item storage because `putItem()`
+This is an implementation detail, but it is important when changing item storage because a write
 uses the key schema to overwrite items with the same primary key.
+
+`SimDynamoDbTableWrites` is how a table takes a write, in two steps. `prepareItem` and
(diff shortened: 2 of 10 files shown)


hughgrigg:
@coderabbitai full review

PR #351, opened 2026-08-03
feat: notify a simulated Lambda function when an Object is created or…

### LINKED ISSUE #331: Notify a simulated Lambda function when an Object is created or deleted in a simulated S3 Bucket
## Problem

Simulated S3 has no event notifications, so the single most common S3 architecture cannot be tested
against it: upload an object, a Lambda function runs. A test has to invoke the handler directly with
a hand-written event, which skips the configuration, the permission and the event document, so the
three things the deployment actually depends on are the three things never exercised. Issue #320
already works around this by saying the function is invoked directly.

The absence is silent rather than loud. `PutBucketNotificationConfiguration` is not a command sim S3
has, and a CDK stack containing a notification deploys clean and wires nothing, because the
`Custom::S3BucketNotifications` refusal is worded in a way sim CloudFormation downgrades to a skip.

## Desired behaviour

Configure a simulated S3 Bucket to notify a simulated Lambda function, put or delete an Object, and
the function is invoked with the S3 event document. The function's resource policy decides whether
S3 may invoke it, checked when the configuration is applied and again when the event is delivered,
as real S3 does.

## Example

```typescript
await simAws.lambda().addPermission(
  new AddPermissionCommand({
    FunctionName: "thumbnailer",
    StatementId: "AllowS3",
    Action: "lambda:InvokeFunction",
    Principal: "s3.amazonaws.com",
    SourceArn: "arn:aws:s3:::uploads",
    SourceAccount: simAws.defaultAccountId,
  }),
);

await simAws.s3().putBucketNotificationConfiguration(
  new PutBucketNotificationConfigurationCommand({
    Bucket: "uploads",
    NotificationConfiguration: {
      LambdaFunctionConfigurations: [
        {
          Events: ["s3:ObjectCreated:*"],
          LambdaFunctionArn: thumbnailerArn,
          Filter: { Key: { FilterRules: [{ Name: "prefix", Value: "raw/" }] } },
        },
      ],
    },
  }),
);

await simAws.s3().putObject(
  new PutObjectCommand({ Bucket: "uploads", Key: "raw/cat.jpg", Body: bytes }),
);

// Delivery happens in the background, so wait for the simulation to settle.
```

## Proposed scope

**Configuration.** A model under `src/service/s3/bucket/notification/`, held on `SimS3Bucket`
beside the website configuration, the policy and the public access block. All four destination
groups plus a separate EventBridge flag, generated configuration ids, prefix and suffix filters, the
event type vocabulary and the overlap rule. `PutBucketNotificationConfiguration` replaces the whole
configuration; `GetBucketNotificationConfiguration` answers an empty one for a Bucket with none.
Authorized as `s3:PutBucketNotification` and `s3:GetBucketNotification`, which are the real IAM
action names and do not match the API names. `sim-s3-public-access-block-authorizer.ts:15-22`
comments the same mismatch and is the precedent.

**Delivery.** A destination port under `src/service/s3/notification/destination/`, in the S3 tree
because S3 is what pushes, mirroring Lambda owning `SimLambdaEventSourceQueues` for the pull case.
A `ReadonlyMap` keyed on the destination ARN's service segment resolves an ARN to a destination that
owns both `validate` and `deliver`, since configuration-time validation and delivery-time re-check
are the same per-destination question.

**The event.** A destination-agnostic `SimS3ObjectEvent`, with the `Records` document as a separate
serialiser so a later SNS or EventBridge destination shapes it differently without a rewrite.
`eventTime` from `background.now()`, so a frozen clock produces a fixed timestamp. `size` from the
Buffer length and `eTag` as its hex MD5, both computed at the raise point. `versionId` omitted, which
is what real S3 does for an unversioned Bucket. A per-key hex `sequencer`. The object key
form-URL-encoded, as real S3 encodes it.

**The raise point.** Every real write path funnels through `await bucket.putObject(object)` at
`src/service/s3/command/put-object/put-object.handler.ts:90`, which is also the only layer holding
the caller and knowing which `ObjectCreated:*` sub-event this is. One call there and one in the
`DeleteObject` handler covers the SDK, the intercepted SDK client and the REST endpoint.

## Acceptance criteria

- [ ] Putting an Object into a configured Bucket invokes the function with the event document, after
      `await simAws.backgroundTasksComplete()`.
- [ ] Deleting an Object raises `s3:ObjectRemoved:Delete`, with `size` and `eTag` absent from the
      record.
- [ ] A prefix filter and a suffix filter each admit a matching key and exclude a non-matching one.
- [ ] `PutBucketNotificationConfiguration` naming a function whose resource policy does not admit
      `s3.amazonaws.com` from this Bucket is refused, and nothing is stored.
- [ ] A permission removed after the configuration was applied stops delivery, rather than the
      configuration-time check standing in for the delivery-time one.
- [ ] `GetBucketNotificationConfiguration` reports what was applied, and an empty configuration for a
      Bucket with none.
- [ ] Two configurations whose expanded event sets intersect and whose prefixes overlap are refused.
- [ ] A handler that throws does not fail the `PutObject` and does not reject
      `backgroundTasksComplete()`.
- [ ] Every unsimulated destination, event type and filter rule name is refused by name.

**The SDK output is not symmetric with the input.** `GetBucketNotificationConfigurationCommandOutput
extends NotificationConfiguration`, so the four destination keys sit at the top level of the
response, while `PutBucketNotificationConfigurationRequest` nests them under a required
`NotificationConfiguration`. TypeScript will not catch getting this wrong, because SDK interception
replaces `send` and the caller is typed by the SDK regardless of what the simulator returns. A docs
example asserting `result.LambdaFunctionConfigurations` would compile, pass `examples:check` and read
`undefined`, which looks exactly like "no configurations are stored". Worth a comment on the output
type, since the neighbouring `GetBucketPolicy` pair has no such asymmetry.

**The overlap rule keys on event sets, not event strings.** AWS's own invalid example pairs
`s3:ObjectCreated:*` with `s3:ObjectCreated:Put`. Compared as strings the simulator would accept a
configuration real S3 refuses with `InvalidArgument`. Expand `:*` to its concrete members before
comparing and treat any intersection as a conflict. AWS's three invalid examples and its valid
`images` + `.jpg` / `images` + `.png` case are worth porting verbatim as tests, since a naive
implementation refuses the valid one.

**A handler writing back into its trigger Bucket hangs the run.** `BackgroundTasks.complete()`
drains `while (this.pending.size > 0)`, so an in-process loop is a hung test with no stack. That is
worse than the real-AWS version, where concurrency limits make it visible. Prefix filtering is the
standard guard and is why filters belong here rather than in a follow-up. A bounded delivery ceiling
that throws a named error is worth adding on top, and is the one delivery failure allowed to reject
its background task.

**Delivery errors must be caught.** `BackgroundTasks.schedule` deliberately keeps a failed task's
promise rejected so `complete()` can surface it. A handler exception escaping the delivery task
would make an unrelated `await simAws.backgroundTasksComplete()` throw, which real S3 never does to a
`PutObject` caller. Swallowing it entirely makes a broken handler invisible, so record the outcome
somewhere inspectable and warn once per cause, following `SimLambdaEnvironmentConflicts`.

**The construction hazard.** `createLambda` already calls `scope.s3()` eagerly, so an eager
`scope.lambda()` in `createS3` is an immediate stack overflow: `Memo.getOrCreate` populates its
cache only after the factory returns and has no re-entrancy guard. Resolve destinations lazily from
`this.simAws`, which `createLambda` already passes safely as `runAsOwner`. Worth a regression test
constructing `simAws.lambda()` first in one case and `simAws.s3()` first in the other, since the bug
reproduces under one order only. Note that an eager `scope.sqs()` would work today and only break
when a future service closes the cycle, so it should be refused on principle rather than accepted
because it passes.

**`userIdentity.principalId` has no honest source.** `PutObjectAuthorizer.authorize` returns `void`
and discards the decision whose `.caller` is the only resolved principal in the request, and
simulated IAM carries ARNs rather than `AIDA...` unique ids. Return the decision and emit the ARN,
with the divergence recorded. `src/serve/payload-2/sim-payload-2-event.type.ts:75-88` is the
accepted precedent for exactly this. Do not invent an AIDA-shaped string.
`responseElements` and `requestParameters.sourceIPAddress` get the same treatment.

**`eventVersion` is a moving target.** AWS increments the minor version as the schema evolves and
advises consumers to compare the major for equality and the minor as a floor. Emit what the S3 event
message structure page documents now, in one named constant carrying the doc URL, and tell readers
in Limitations to assert on the major. Do not copy the `2.1` in the older Lambda developer guide
sample.

**A standalone `new SimS3()` cannot deliver.** It builds its own `BackgroundTasks`, which no
`simAws.backgroundTasksComplete()` can drain, so even a delivered notification could not be awaited.
`PutBucketNotificationConfiguration` should refuse outright there, naming standalone construction and
pointing at `SimAws`, as `SimLambdaNoEventSourceQueues` does.

**CDK `BucketDeployment` and `mountBucketFilesystem` raise nothing.** Both swap the whole storage
backend rather than putting Objects, so no notification is possible and none should be invented. Real
CDK `BucketDeployment` fires one `ObjectCreated:Put` per file, so this belongs in Limitations rather
than being worked around.

## Out of scope

SQS, SNS and EventBridge destinations, refused by name. Every event type simulated S3 cannot
produce: `Copy`, `Post`, `CompleteMultipartUpload`, `DeleteMarkerCreated`, `ObjectRestore:*`,
`Replication:*`, `LifecycleExpiration:*`, `LifecycleTransition`, `IntelligentTiering`,
`ObjectTagging:*`, `ObjectAcl:Put` and `ReducedRedundancyLostObject`. Any CloudFormation or CDK path.
`s3:TestEvent`, which is an SQS and SNS mechanism. Surfacing `eTag` and `lastModified` on
`GetObject` and `ListObjects` responses, which is a separate realism improvement with a real cost:
`FilesystemS3BucketStorage.listObjects` maps every key through `getObject`, so storing an eTag on
the model would hash every file on disk on every list.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/329, which gives `s3:ObjectRemoved:*` a
producer, and on https://github.com/KensioSoftware/yulin/issues/328, without which a permission
carrying `SourceAccount` can never match.

This is deliberately larger than the usual steer. It does not split honestly: separating the
configuration commands from delivery ships a store that does nothing, and separating destination
validation ships a configuration that accepts a permission-less destination and then silently
delivers nothing, which is the exact failure the feature exists to catch. The nearest structural
analogue in this repo, the SQS Lambda event source mapping, landed in one commit.

`docs/services/s3/README.md` has no page-level Limitations section and needs one.

hughgrigg:
… deleted in a simulated S3 Bucket

Configure a simulated S3 Bucket with PutBucketNotificationConfiguration and put or delete an Object, and a simulated Lambda function is invoked with the S3 event document. The function's resource policy decides whether S3 may invoke it, checked when the configuration is applied and again for every event.

Resolves https://github.com/KensioSoftware/yulin/issues/331

### DIFF
--- a/docs/services/s3/README.md
+++ b/docs/services/s3/README.md
@@ -278,6 +278,180 @@ failures come back.
 - A Bucket using filesystem-backed storage refuses deletion. See
   [Filesystem-backed Bucket storage](#filesystem-backed-bucket-storage).
 
+## Event notifications
+
+A simulated S3 Bucket can notify a simulated Lambda function when an Object is created or removed.
+The configuration is applied with `PutBucketNotificationConfigurationCommand` and read back with
+`GetBucketNotificationConfigurationCommand`.
+
+The function's resource policy decides whether S3 may invoke it. That is checked when the
+configuration is applied, and again for every event, as real S3 does.
+
+```typescript sim-s3-event-notifications
+/**
+ * Notifying a simulated Lambda function when an Object is created.
+ */
+
+import {
+  AddPermissionCommand,
+  CreateFunctionCommand,
+} from "@aws-sdk/client-lambda";
+import {
+  CreateBucketCommand,
+  PutBucketNotificationConfigurationCommand,
+  PutObjectCommand,
+} from "@aws-sdk/client-s3";
+import { SimAws } from "@kensio/yulin";
+import { makeLambdaZipFileInput } from "@kensio/yulin/lambda";
+
+interface S3EventDocument {
+  Records: [{ eventName: string; s3: { object: { key: string } } }];
+}
+
+const simAws = new SimAws();
+const thumbnailerArn = `arn:aws:lambda:${simAws.defaultRegionName}:${simAws.defaultAccountId}:function:thumbnailer`;
+
+await simAws.s3().createBucket(new CreateBucketCommand({ Bucket: "uploads" }));
+
+await simAws.lambda().createFunction(
+  new CreateFunctionCommand({
+    FunctionName: "thumbnailer",
+    Role: `arn:aws:iam::${simAws.defaultAccountId}:role/ThumbnailerRole`,
+    Code: {
+      ZipFile: makeLambdaZipFileInput((event: S3EventDocument) => {
+        console.log(event.Records[0].eventName, event.Records[0].s3.object.key);
+
+        return "thumbnailed";
+      }),
+    },
+  }),
+);
+
+await simAws.lambda().addPermission(
+  new AddPermissionCommand({
+    FunctionName: "thumbnailer",
+    StatementId: "AllowS3",
+    Action: "lambda:InvokeFunction",
+    Principal: "s3.amazonaws.com",
+    SourceArn: "arn:aws:s3:::uploads",
+    SourceAccount: simAws.defaultAccountId,
+  }),
+);
+
+await simAws.s3().putBucketNotificationConfiguration(
+  new PutBucketNotificationConfigurationCommand({
+    Bucket: "uploads",
+    NotificationConfiguration: {
+      LambdaFunctionConfigurations: [
+        {
+          Id: "thumbnail-raw-uploads",
+          Events: ["s3:ObjectCreated:*"],
+          LambdaFunctionArn: thumbnailerArn,
+          Filter: { Key: { FilterRules: [{ Name: "prefix", Value: "raw/" }] } },
+        },
+      ],
+    },
+  }),
+);
+
+await simAws.s3().putObject(
+  new PutObjectCommand({
+    Bucket: "uploads",
+    Key: "raw/cat.jpg",
+    Body: "cat picture",
+  }),
+);
+
+// Delivery happens in the background, so wait for the simulation to settle.
+await simAws.backgroundTasksComplete();
+```
+
+The event types a configuration can name are `s3:ObjectCreated:*`, `s3:ObjectCreated:Put`,
+`s3:ObjectRemoved:*` and `s3:ObjectRemoved:Delete`. Any other S3 event type is refused by name rather
+than stored and never raised.
+
+A configuration can filter on an object key prefix, a suffix, or both. Two configurations that share
+an event type and whose filters could both match the same key are refused with `InvalidArgument`, as
+real S3 refuses them. Overlapping prefixes are fine when the suffixes do not overlap, so one function
+can take the `.jpg` files under a prefix while another takes the `.png` files under the same one.
+
+`PutBucketNotificationConfigurationCommand` replaces the whole configuration rather than adding to
+it. `GetBucketNotificationConfigurationCommand` answers an empty configuration for a Bucket that has
+none. Note that the response carries the destination groups at the top level, while the request nests
+them under `NotificationConfiguration`:
+
+```typescript
+const read = await simAws
+  .s3()
+  .getBucketNotificationConfiguration(
+    new GetBucketNotificationConfigurationCommand({ Bucket: "uploads" }),
+  );
+const configurations = read.LambdaFunctionConfigurations ?? [];
+```
+
+The two commands are authorized as `s3:PutBucketNotification` and `s3:GetBucketNotification`. Those
+are the real IAM action names, and they do not match the API names.
+
+### What arrives at the handler
+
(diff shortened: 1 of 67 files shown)


hughgrigg:
@coderabbitai rate limit

PR #353, opened 2026-08-03
feat: config sim S3 event notifications CDK bucket.addEventNotification

### LINKED ISSUE #332: Configure simulated S3 event notifications from a CDK bucket.addEventNotification
## Problem

CDK never emits `AWS::S3::Bucket` `NotificationConfiguration`. Every notification path goes through
a `Custom::S3BucketNotifications` resource, including `eventBridgeEnabled: true`. Verified by
synthesising with the repo's own aws-cdk-lib 2.263.0: the Bucket resource carries only
`BucketName`, and in the EventBridge case it has no `Properties` at all.

Simulated CloudFormation throws
`Unsupported sim CloudFormation Custom Resource S3BucketNotifications` for it. That message contains
both `Unsupported sim` and `CloudFormation`, which is the pair
`SimCfnResourceCreateOperation.isUnsupportedResourceError` matches to downgrade a failure into a
skip. So a CDK stack with a notification deploys today with status `CREATE_COMPLETE`, wires nothing,
and says nothing.

## Desired behaviour

A CDK stack calling `bucket.addEventNotification(...)` deploys against simulated CloudFormation and
the notification works, so a test uploads an object through the deployed Bucket and the deployed
function runs.

## Example

```typescript
bucket.addEventNotification(
  s3.EventType.OBJECT_CREATED,
  new s3n.LambdaDestination(handler),
  { prefix: "raw/" },
);
```

## Current CDK output

The custom resource carries the SDK shape, so it is close to a pass-through to
`PutBucketNotificationConfiguration`:

```json
{
  "ServiceToken": { "Fn::GetAtt": ["BucketNotificationsHandler050a...", "Arn"] },
  "BucketName": { "Ref": "SiteBucket397A1860" },
  "NotificationConfiguration": {
    "LambdaFunctionConfigurations": [
      {
        "Events": ["s3:ObjectCreated:*"],
        "LambdaFunctionArn": { "Fn::GetAtt": ["Handler886CB40B", "Arn"] }
      }
    ],
    "QueueConfigurations": [
      {
        "Events": ["s3:ObjectRemoved:*"],
        "Filter": {
          "Key": {
            "FilterRules": [
              { "Name": "suffix", "Value": ".jpg" },
              { "Name": "prefix", "Value": "uploads/" }
            ]
          }
        },
        "QueueArn": { "Fn::GetAtt": ["Queue4A7E3555", "Arn"] }
      }
    ]
  },
  "Managed": true,
  "SkipDestinationValidation": false
}
```

CDK also emits the `AWS::Lambda::Permission` the destination needs, with `Principal:
"s3.amazonaws.com"`, a `SourceArn` and a `SourceAccount`, which sim CloudFormation already deploys.

## Proposed scope

- Extract the hand-written `Custom::` branches at
  `src/service/cloudformation/resource/resolve/service/sim-cfn-service-resolver.ts:18-30` into a
  small `ReadonlyMap` of supported CDK custom resource type names, mirroring
  `simCfnServiceResourceFactories`, then add the second entry.
- A factory under `src/service/cloudformation/cdk/s3/bucket-notifications/` reading `BucketName`,
  `NotificationConfiguration`, `Managed` and `SkipDestinationValidation` from the resolved
  properties, and calling the ordinary `putBucketNotificationConfiguration` command path, so a
  template and an SDK caller get identical validation.
- `ServiceToken` ignored. CDK's `BucketNotificationsHandler` is a Python function, so it stays
  skipped on its runtime, as CDK's `BucketDeployment` provider already is.

## Acceptance criteria

- [ ] A CDK stack with `bucket.addEventNotification(EventType.OBJECT_CREATED, new
      LambdaDestination(fn))` deploys, and putting an Object into the deployed Bucket invokes the
      deployed function.
- [ ] The same stack with a `prefix` filter delivers only matching keys.
- [ ] `Managed: false` is refused by name, worded so the Stack fails rather than skipping the
      resource.
- [ ] A stack whose function has no matching `AWS::Lambda::Permission` fails to deploy, rather than
      deploying and delivering nothing.

Every refusal added here must be worded `Invalid Custom::S3BucketNotifications Resource
<logicalId>: ...`. Anything containing both `Unsupported sim` and `CloudFormation` is silently
downgraded to a skip with `CREATE_COMPLETE`, which is how the current situation arose.
`sim-cfn-lambda-event-source-mapping-property-rules.ts` already dodges this deliberately and is the
pattern to follow. Worth a test asserting one wording skips and the other fails the Stack.

`Managed: false` means merge with the existing configuration rather than replace it, which CDK emits
for an imported Bucket. Treating it as a replace would silently discard configurations that survive
on real AWS, so it should be refused rather than approximated.

The local test needs to deploy into a scope whose Account and Region match the CDK synth
environment, because `SourceAccount` is a synth-time literal while the default simulated scope is a
different Account. That is worth a sentence in the docs, since it is the first thing a user will hit.

## Out of scope

The `Managed: false` merge itself. `Custom::S3AutoDeleteObjects`. The `AWS::S3::Bucket`
`NotificationConfiguration` template property, which is
https://github.com/KensioSoftware/yulin/issues/333. Queue and topic destinations, so the test uses a
Lambda destination only.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/331.

Ordering this before the template property is deliberate. The custom resource carries the SDK shape
already, so it does not need the template-shape translation layer at all, which makes it the
earliest point the feature is demonstrable the way most users will reach it.

hughgrigg:
CDK writes every event notification as a Custom::S3BucketNotifications Resource carrying the PutBucketNotificationConfiguration request, so sim CloudFormation applies it through the ordinary command path. Managed: false is refused by name rather than approximated, as are the destination groups simulated S3 cannot deliver to.

Resolves https://github.com/KensioSoftware/yulin/issues/332

## Summary by CodeRabbit

- **New Features**
  - Added support for deploying CDK S3 bucket notifications through simulated CloudFormation.
  - S3 event notifications can now trigger Lambda handlers, including prefix filtering and destination validation.
  - Added support for CDK custom resources for bucket deployment and S3 notifications.
  - Added an end-to-end example demonstrating object uploads and event delivery.

- **Documentation**
  - Documented deployment requirements, supported configurations, limitations, and failure behavior.

### DIFF
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -575,6 +575,21 @@ A CDK BucketDeployment can copy files from synthesized asset output into the sim
 the Bucket is configured for website hosting, those files can be served through Yulin's local
 server.
 
+## CDK S3 Bucket notifications
+
+`bucket.addEventNotification(...)` synthesizes a `Custom::S3BucketNotifications` resource rather than
+a Bucket property. Sim CloudFormation applies the configuration it carries through the ordinary
+`PutBucketNotificationConfiguration` path, so an Object put into the deployed Bucket reaches the
+deployed function.
+
+Deploy into an Account and Region matching the ones the CDK app synthesized for. The `SourceAccount`
+on the `AWS::Lambda::Permission` CDK writes beside the notification is a synth-time literal, so a
+stack deployed into another Account leaves S3 unable to validate the destination, and the stack
+fails.
+
+See [Event notifications](../s3/README.md#event-notifications) in the S3 docs for the configuration
+itself and what it refuses.
+
 ## CloudFront resources from CDK
 
 Sim CloudFormation can create CloudFront Distributions from CloudFormation or CDK templates.
@@ -1090,7 +1105,7 @@ The resource types it creates are:
 - `AWS::SecretsManager::Secret`
 - `AWS::SQS::Queue`
 - `AWS::SSM::Parameter`
-- selected CDK custom resources, including CDK S3 BucketDeployment
+- selected CDK custom resources: `Custom::CDKBucketDeployment` and `Custom::S3BucketNotifications`
 
 Each service's own docs describe what its resource types support.
 
--- a/docs/services/s3/README.md
+++ b/docs/services/s3/README.md
@@ -504,8 +504,8 @@ writes do not match it. Without that, the simulation stops after a thousand deli
 - `Managed: false` on a `Custom::S3BucketNotifications` resource is refused rather than approximated,
   and a queue, topic or EventBridge destination in one is refused by name as it is for an SDK caller.
 - A CDK `BucketDeployment` and `mountBucketFilesystem(...)` both replace the whole storage backend
-  rather than putting Objects, so neither raises anything, where real CDK `BucketDeployment` fires one
-  `ObjectCreated:Put` per file.
+  rather than putting Objects, so neither raises anything, whereas real CDK `BucketDeployment` fires
+  one `ObjectCreated:Put` per file.
 - A function ARN naming a version or an alias is refused, since simulated Lambda has neither.
 - `s3:TestEvent` is not sent. It is an SQS and SNS mechanism.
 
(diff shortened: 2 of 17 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #354, opened 2026-08-03
feat: config sim S3 event notifications AWS::S3::Bucket template

### LINKED ISSUE #333: Configure simulated S3 event notifications from an AWS::S3::Bucket template
## Problem

`SimCfnS3BucketCreator` reads three properties off `AWS::S3::Bucket`: `BucketName`,
`WebsiteConfiguration` and `PublicAccessBlockConfiguration`. Everything else is dropped without
comment, so a template declaring `NotificationConfiguration` deploys a Bucket with no notifications
and no diagnostic.

CDK routes around this property, but plain CloudFormation, SAM, the Serverless Framework, Terraform
and CDK's own `CfnBucket` escape hatch all express notifications through it, so it is not a dead
declaration path.

## Desired behaviour

An `AWS::S3::Bucket` carrying `NotificationConfiguration` deploys the configuration through the
ordinary command path, so a template and an SDK caller get identical validation. A property the
simulator has no behaviour for fails the Stack rather than being dropped.

## Example

```yaml
UploadsBucket:
  Type: AWS::S3::Bucket
  Properties:
    BucketName: uploads
    NotificationConfiguration:
      LambdaConfigurations:
        - Event: "s3:ObjectCreated:*"
          Function: !GetAtt Thumbnailer.Arn
          Filter:
            S3Key:
              Rules:
                - Name: prefix
                  Value: raw/
```

## Proposed scope

A reader translating the CloudFormation shape into the SDK shape the model already uses. The two
shapes differ in four places, and each is a silent failure if missed:

| CloudFormation | SDK |
| --- | --- |
| `LambdaConfigurations` | `LambdaFunctionConfigurations` |
| `Event`, a single string | `Events`, an array |
| `Function` / `Queue` / `Topic` | `LambdaFunctionArn` / `QueueArn` / `TopicArn` |
| `Filter.S3Key.Rules` | `Filter.Key.FilterRules` |

Routing through `putBucketNotificationConfiguration`, as `SimCfnS3BucketPolicyCreator` routes
through `putBucketPolicy`.

An unsimulated-property rule set for `AWS::S3::Bucket`, following
`src/service/sqs/cfn/queue/sim-cfn-sqs-queue-property-rules.ts`, so a property the simulator does
not implement fails by name instead of vanishing.

## Acceptance criteria

- [ ] A template declaring a Lambda notification deploys it, and putting an Object into the deployed
      Bucket invokes the function.
- [ ] A `Filter.S3Key.Rules` prefix rule is applied, and a test asserts that spelling specifically.
- [ ] The singular `Event` string becomes a one-element `Events` array.
- [ ] A configuration the command would refuse from an SDK caller also fails the Stack, with the
      same reason.
- [ ] An `AWS::S3::Bucket` property the simulator has no behaviour for fails the Stack by name,
      rather than being dropped.

There are three distinct wire spellings for a key filter, not two. The SDK and CDK's custom resource
use `Filter.Key.FilterRules`, the REST XML uses `Filter.S3Key.FilterRule`, and CloudFormation uses
`Filter.S3Key.Rules`. A reader written against the wrong one reads `undefined`, deploys an
unfiltered configuration, and the Stack succeeds.

Refusals must be worded `Invalid AWS::S3::Bucket NotificationConfiguration in Resource
<logicalId>: ...`. Anything containing both `Unsupported sim` and `CloudFormation` is downgraded to
a skip, which for a Bucket that does get created is the dangerous case: it looks right and behaves
differently.

Real CloudFormation has a circular dependency here that it documents itself. The Bucket needs the
function ARN and the function's permission needs the Bucket ARN, so templates hardcode `BucketName`
and add `DependsOn`. Simulated CloudFormation would otherwise surface it as an unhelpful dependency
resolution failure, so it is worth naming in the docs.

Adding the unsimulated-property rule set is a behaviour change for existing templates: a Bucket
property that deploys today by being ignored will start failing. That is the intended direction, but
it belongs in the pull request description.

## Out of scope

`Custom::S3BucketNotifications`, which is
https://github.com/KensioSoftware/yulin/issues/332. Queue and topic destinations, which the command
still refuses.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/331.

hughgrigg:
The NotificationConfiguration property deploys through the ordinary PutBucketNotificationConfiguration path, so a template and an SDK caller get identical validation. The four names CloudFormation spells differently from the SDK are translated, and the SDK and REST XML spellings of the key filter are refused rather than read as unfiltered.

An unsimulated-property rule set fails an AWS::S3::Bucket property by name. BucketEncryption and Tags are the exceptions: nothing simulated can tell them apart, and CDK puts both on almost every Bucket it synthesizes.

Resolves https://github.com/KensioSoftware/yulin/issues/333

## Summary by CodeRabbit

- **New Features**
  - Added CloudFormation support for configuring S3 bucket event notifications, including Lambda destinations and key-prefix filters.
  - Added S3 bucket website hosting and public-access configuration support.
  - Added validation for supported, ignored, unsupported, and invalid bucket properties.

- **Bug Fixes**
  - Improved deployment errors and resource handling for invalid notification configurations and unsupported properties.

- **Documentation**
  - Added CloudFormation configuration guidance, limitations, deployment requirements, and working examples.

- **Tests**
  - Added coverage for notification delivery, filtering, website configuration, validation, refusals, and failed deployments.


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #355, opened 2026-08-04
feat: send a simulated S3 event notification to a simulated SQS queue

### LINKED ISSUE #334: Send a simulated S3 event notification to a simulated SQS queue
## Problem

`QueueConfigurations` is refused by simulated S3, so the second most common S3 notification shape
cannot be tested. It is also the shape AWS itself recommends for anything high volume, because the
queue absorbs the burst that would otherwise hit a function's concurrency limit.

The chain it unlocks is the more valuable half: S3 to SQS to Lambda. A handler consuming that has to
reach through a double envelope, parsing the S3 event document out of an SQS message body, and
getting the shape of that wrong is exactly the kind of thing a test should catch rather than a
deployment.

## Desired behaviour

Configure a simulated S3 Bucket to notify a simulated SQS queue, put an Object, and the event
document arrives on the queue as a message body. A Lambda event source mapping on that queue then
delivers it to a function with no further wiring.

## Example

```typescript
await simAws.s3().putBucketNotificationConfiguration(
  new PutBucketNotificationConfigurationCommand({
    Bucket: "uploads",
    NotificationConfiguration: {
      QueueConfigurations: [
        { Events: ["s3:ObjectCreated:*"], QueueArn: uploadsQueueArn },
      ],
    },
  }),
);

await simAws.s3().putObject(
  new PutObjectCommand({ Bucket: "uploads", Key: "raw/cat.jpg", Body: bytes }),
);

// In the consuming function:
//   const event = JSON.parse(record.body);
//   event.Records[0].s3.object.key === "raw/cat.jpg"
```

## Proposed scope

- A queue destination class plus one entry in the destination ARN service-segment map.
- The whole `Records` document sent as the message body through the ordinary `SendMessage` path, as
  the `s3.amazonaws.com` service principal, one message per record and no message attributes.
- `sqs:SendMessage` validated against the queue policy when the configuration is applied, with
  `aws:SourceArn` and `aws:SourceAccount` supplied, and re-checked at delivery.
- The queue ARN to URL step moved into `src/service/sqs/queue/sim-sqs-queue-arn.ts` and exported,
  rather than the URL format being written a third time after `SimSqsQueueArn` and
  `SimLambdaSqsEventSourceArn`.

Nothing in SQS or Lambda needs changing for the chain to work: `SimSqsQueue.add` already announces
to watchers and `SimLambdaEventSourcePoller` already turns that into a scheduled poll.

## Acceptance criteria

- [ ] Putting an Object into a configured Bucket puts the event document on the queue, and
      `JSON.parse(body).Records[0].s3.object.key` is the key.
- [ ] A queue with no policy admitting `s3.amazonaws.com` from the Bucket causes
      `PutBucketNotificationConfiguration` to be refused, and nothing is stored.
- [ ] A queue policy admitting only a different source ARN is refused the same way.
- [ ] A queue in another Region is refused.
- [ ] An event source mapping on the same queue delivers the event to a function, so the chain works
      end to end after one `await simAws.backgroundTasksComplete()`.
- [ ] A FIFO queue is refused by name, as real S3 refuses one as a destination.

A cross-Account queue should be allowed, evaluated against the queue-owning Account's IAM. Real S3
requires only that the queue is in the same Region. AWS's own documented destination queue policy
carries a `StringEquals aws:SourceAccount` guard, which only has a job when the queue and the Bucket
are in different Accounts, so refusing cross-Account would be stricter than AWS and inconsistent
with cross-Account Lambda destinations being allowed.

Note that CDK's `SqsDestination` emits a queue policy with `ArnLike aws:SourceArn` and no
`aws:SourceAccount`, while its `LambdaDestination` always sets `SourceAccount`. Both need to be
satisfiable.

The headline test and docs example should be the full S3 to SQS to Lambda chain, extending
`test/lambda/event-source-fixture.ts` rather than adding a new fixture. The negative test, that the
configuration is refused when the queue has no policy, is what proves the feature fails closed and
should not be left out.

`s3:TestEvent` stays omitted. It is an SQS and SNS mechanism, and sending it would make the simplest
possible test two messages long and hand the consuming function a body with no `Records` on the
first poll, breaking the example that teaches the double envelope. The validation it exists for on
real AWS is simulated directly. Its flat `{Service, Event, Time, Bucket, RequestId, HostId}` shape
belongs in Limitations so the omission is stated rather than implied.

## Out of scope

SNS and EventBridge destinations, still refused. The KMS key policy statement CDK's `SqsDestination`
emits for an encrypted queue.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/331 for the notification machinery, and on
https://github.com/KensioSoftware/yulin/issues/330 for queue policies. Without the latter a service
principal owns no identity policies and every delivery is an implicit deny, so the only way to make
it work would be to skip authorization, which is the fail-open case this simulator rules out.

hughgrigg:
A QueueConfigurations entry names a queue by ARN, and the Records document arrives as one message body, so an event source mapping on the queue carries it to a function with no further wiring.

The queue policy is the whole decision, checked when the configuration is applied and again at delivery, with aws:SourceArn and aws:SourceAccount both supplied. A cross-Account queue is allowed and a cross-Region one is not, as on real S3. A FIFO queue is refused by name.

Resolves https://github.com/KensioSoftware/yulin/issues/334

## Summary by CodeRabbit

* **New Features**
  * Added S3 event notifications to simulated SQS queues.
  * Added support for Lambda processing of S3 events delivered through SQS.
  * Added CloudFormation and CDK configuration support for queue destinations.
  * Added queue policy validation, region checks, cross-account authorization, and source-account conditions.
  * Added detailed S3-to-SQS documentation and executable examples.

* **Bug Fixes**
  * Improved validation and error messages for unsupported, missing, FIFO, cross-region, and unauthorized queue destinations.
  * Applied notification ID and filter-overlap checks across all destination types.

### DIFF
--- a/docs/services/s3/README.md
+++ b/docs/services/s3/README.md
@@ -280,12 +280,14 @@ failures come back.
 
 ## Event notifications
 
-A simulated S3 Bucket can notify a simulated Lambda function when an Object is created or removed.
-The configuration is applied with `PutBucketNotificationConfigurationCommand` and read back with
+A simulated S3 Bucket can notify a simulated Lambda function or a simulated SQS queue when an Object
+is created or removed. The configuration is applied with
+`PutBucketNotificationConfigurationCommand` and read back with
 `GetBucketNotificationConfigurationCommand`.
 
-The function's resource policy decides whether S3 may invoke it. That is checked when the
-configuration is applied, and again for every event, as real S3 does.
+The destination's own policy decides whether S3 may reach it: the function's resource policy, or the
+queue's `Policy` attribute. That is checked when the configuration is applied, and again for every
+event, as real S3 does.
 
 ```typescript sim-s3-event-notifications
 /**
@@ -374,6 +376,8 @@ A configuration can filter on an object key prefix, a suffix, or both. Two confi
 an event type and whose filters could both match the same key are refused with `InvalidArgument`, as
 real S3 refuses them. Overlapping prefixes are fine when the suffixes do not overlap, so one function
 can take the `.jpg` files under a prefix while another takes the `.png` files under the same one.
+The rule applies across the destination groups, so a function and a queue that both want the same
+event are refused as readily as two functions.
 
 `PutBucketNotificationConfigurationCommand` replaces the whole configuration rather than adding to
 it. `GetBucketNotificationConfigurationCommand` answers an empty configuration for a Bucket that has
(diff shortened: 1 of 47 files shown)


hughgrigg:
@coderabbitai full review

PR #363, opened 2026-08-04
feat: simulated CloudFormation template conditions Fn::If

### LINKED ISSUE #359: Deploy a simulated CloudFormation template with Conditions and Fn::If
## Problem

The simulated CloudFormation template engine has no `Conditions` section. Nothing under `src/service/cloudformation/` reads it, and `Fn::If` fails to parse with `Unsupported Sim CloudFormation intrinsic function Fn::If`.

Conditions are how one template covers more than one environment, so a template that deploys against real CloudFormation cannot be deployed against the simulator without being rewritten. CDK emits them too, for example from `CfnCondition` and from constructs that vary a property by environment.

## Desired behaviour

A template's `Conditions` section is evaluated once per deployment, from parameter values and pseudo parameters, using `Fn::Equals`, `Fn::And`, `Fn::Or` and `Fn::Not`.

`Fn::If [conditionName, valueIfTrue, valueIfFalse]` then resolves in resource properties and outputs, and the unused branch is not resolved, so it can name a resource that does not exist.

A resource carrying a `Condition` attribute whose condition is false is not created at all. It does not appear in `stack.resources`, and `Ref` or `Fn::GetAtt` to it fails. This is different from an unsupported resource type, which is recorded as a skipped resource and answers with stand-in values.

## Example

```json
{
  "Parameters": { "EnvName": { "Type": "String" } },
  "Conditions": {
    "IsProd": { "Fn::Equals": [{ "Ref": "EnvName" }, "prod"] }
  },
  "Resources": {
    "Backups": { "Type": "AWS::S3::Bucket", "Condition": "IsProd" },
    "Site": {
      "Type": "AWS::S3::Bucket",
      "Properties": {
        "BucketName": { "Fn::If": ["IsProd", "site", { "Fn::Sub": "site-${EnvName}" }] }
      }
    }
  }
}
```

Deployed with `EnvName: dev`, the stack has one bucket named `site-dev`.

## Proposed scope

- Parsing and evaluating the `Conditions` section, with `Fn::Equals`, `Fn::And`, `Fn::Or` and `Fn::Not`.
- `Fn::If` in resource properties and outputs.
- The resource-level `Condition` attribute.
- Docs for conditions on the CloudFormation page.

## Acceptance criteria

- [ ] A condition built from `Fn::Equals` over a parameter evaluates to the expected boolean.
- [ ] `Fn::And`, `Fn::Or` and `Fn::Not` compose conditions, including references to other conditions by name.
- [ ] `Fn::If` returns the branch the condition selects, in a resource property and in an output.
- [ ] The unselected `Fn::If` branch is not resolved, so a `Ref` inside it to a non-existent resource does not fail the deployment.
- [ ] A resource with a false `Condition` is not created and is absent from `stack.resources`.
- [ ] A `Condition` attribute naming a condition the template does not define fails with an error naming the resource and the condition.
- [ ] A condition that cannot be evaluated, for example one referencing a resource attribute, fails rather than defaulting to false.

## Out of scope

Conditions on outputs' `Export`, and `Fn::If` inside the `Conditions` section itself.

## Implementation notes

Condition evaluation depends only on parameters and pseudo parameters, so it can happen once in the template phase, before any resource is created, rather than per resource. The resolve context (`SimCfnResolveContext`) already carries `mappings` from the template in the same way, which is the pattern to follow for the evaluated conditions.

Skipping a conditioned-out resource has to be distinct from the existing skipped-resource path used for unsupported resource types. A skipped resource stays in `stack.resources` and answers `Ref` and `Fn::GetAtt` with stand-in values, which would be wrong here: on AWS a conditioned-out resource does not exist, and a `Ref` to it is a template validation error.

The dependency graph is built from `referencedNames()`, so an `Fn::If` needs to be careful about which branches contribute dependencies. Contributing both is the safe reading, at the cost of ordering against resources that are never created.

hughgrigg:
Resolves #359

## Summary by CodeRabbit

* **New Features**
  * Added support for CloudFormation conditions, including `Fn::Equals`, `Fn::And`, `Fn::Or`, `Fn::Not`, and named condition references.
  * Added `Fn::If` support for conditional resource properties and outputs.
  * Resources with false conditions are excluded, with validation for references to omitted resources.
* **Documentation**
  * Added CloudFormation condition guidance, limitations, evaluation behavior, and an example deployment.
* **Bug Fixes**
  * Added validation and clear errors for malformed, undefined, cyclic, or unsupported conditions.

### DIFF
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -516,6 +516,102 @@ The value a lookup returns does not have to be a string. A list value is returne
 used in resource properties and in `Outputs`. A map name or key that is not in `Mappings` fails the
 deployment with an error naming the path that could not be found.
 
+## Conditions
+
+A template `Conditions` section names boolean expressions over the stack's parameter values. A
+condition decides whether a resource is created, and which value `Fn::If` gives a property or an
+output.
+
+```typescript sim-cloudformation-conditions
+/**
+ * Choosing resources and property values by condition in a simulated CFN template.
+ */
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+
+const stack = await simAws.cloudFormation().deployTemplate({
+  stackName: "conditions-stack",
+  template: {
+    Parameters: {
+      EnvName: { Type: "String" },
+    },
+    Conditions: {
+      IsProd: { "Fn::Equals": [{ Ref: "EnvName" }, "prod"] },
+    },
+    Resources: {
+      Backups: {
+        Type: "AWS::S3::Bucket",
+        Condition: "IsProd",
+        Properties: { BucketName: "site-backups" },
+      },
+      Site: {
+        Type: "AWS::S3::Bucket",
+        Properties: {
+          BucketName: {
+            // eslint-disable-next-line no-template-curly-in-string
+            "Fn::If": ["IsProd", "site", { "Fn::Sub": "site-${EnvName}" }],
+          },
+        },
+      },
+    },
+  },
+  parameters: { EnvName: "dev" },
+});
+
+await stack.waitForDeployComplete();
+
+// site-dev
+console.log(simAws.s3().getSimBucketByName("site-dev")?.bucketName);
+
+// false, because IsProd is false
+console.log(stack.resources.has("Backups"));
+```
+
+### Writing a condition
+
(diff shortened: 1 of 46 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #365, opened 2026-08-04
feat: resolve Fn::Select and Fn::Split in simulated CFN templates

### LINKED ISSUE #358: Resolve Fn::Select and Fn::Split in simulated CloudFormation templates
## Problem

The simulated CloudFormation template engine parses `Ref`, `Fn::GetAtt`, `Fn::Join`, `Fn::Sub` and `Fn::FindInMap`. A template using `Fn::Select` or `Fn::Split` fails to parse, with `Unsupported Sim CloudFormation intrinsic function <name>` from `SimCfnFunctionParser`.

This blocks CDK-synthesised templates that pull a hostname out of a URL, which is a common shape. CDK's `FunctionUrlOrigin` sets a CloudFront Origin's domain name to `Fn::Select [2, Fn::Split ["/", <function url>]]`, and the same pattern appears in user code deriving a host from `api.apiEndpoint`.

## Desired behaviour

Both functions parse and resolve during template deployment, with the same value-resolution phases as the existing intrinsics, so their arguments can themselves be `Ref`, `Fn::GetAtt` or another nested function.

- `Fn::Split [delimiter, string]` returns a list.
- `Fn::Select [index, list]` returns the element at a zero-based index.

An out-of-range index, a non-integer index, or a non-list second argument fails with an error naming the resource and property, as AWS rejects the template.

## Example

```json
"DomainName": {
  "Fn::Select": [2, { "Fn::Split": ["/", { "Fn::GetAtt": ["Url", "FunctionUrl"] }] }]
}
```

`https://abc123.lambda-url.eu-west-2.on.aws/` resolves to `abc123.lambda-url.eu-west-2.on.aws`.

## Acceptance criteria

- [ ] `Fn::Split` splits a resolved string on a literal delimiter and yields a list.
- [ ] `Fn::Select` picks an element by zero-based index, including from a `Fn::Split` result and from a plain list literal.
- [ ] Both accept nested intrinsics in their arguments, including `Fn::GetAtt`.
- [ ] An out-of-range or non-integer `Fn::Select` index fails with a diagnostic naming the resource and property.
- [ ] The supported-intrinsic lists in `src/service/cloudformation/README.md` and the CloudFormation docs page are updated.

## Out of scope

Other unsupported intrinsics (`Fn::If`, `Fn::ImportValue`, `Fn::Cidr` and the rest). Each is its own decision.

## Implementation notes

Parsing lives in `src/service/cloudformation/template/parse/fn/`, one directory per function alongside `join/` and `sub/`, dispatched from `SimCfnFunctionParser`. `Fn::Select` returning a list element is the first intrinsic whose intermediate value is a list rather than a string, so the value model may need to carry that.

hughgrigg:
Fn::Split cuts a resolved string into a list on a literal delimiter, and Fn::Select reads one entry out of a list by a zero-based index. Both defer to the resource creation pass when an argument is still an unresolved expression, so the Fn::Select over Fn::Split over Fn::GetAtt shape CDK writes for a Function URL origin resolves.

A resolution failure now names the resource and the property path the value sat at, which applies to the existing intrinsics too.

Resolves https://github.com/KensioSoftware/yulin/issues/358

### DIFF
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -516,6 +516,94 @@ The value a lookup returns does not have to be a string. A list value is returne
 used in resource properties and in `Outputs`. A map name or key that is not in `Mappings` fails the
 deployment with an error naming the path that could not be found.
 
+### `Fn::Split` and `Fn::Select`
+
+`Fn::Split` cuts a string into a list on a delimiter. `Fn::Select` reads one value out of a list by
+its zero-based index. They are usually written together, to pull one part out of a string another
+resource gave.
+
+```typescript sim-cloudformation-fn-select-split
+/**
+ * Naming a bucket after part of another bucket's domain name.
+ */
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+
+const stack = await simAws.cloudFormation().deployTemplate({
+  stackName: "select-split-stack",
+  template: {
+    Resources: {
+      SiteBucket: {
+        Type: "AWS::S3::Bucket",
+        Properties: { BucketName: "site-bucket" },
+      },
+      LogsBucket: {
+        Type: "AWS::S3::Bucket",
+        Properties: {
+          BucketName: {
+            "Fn::Join": [
+              "-",
+              [
+                {
+                  "Fn::Select": [
+                    0,
+                    {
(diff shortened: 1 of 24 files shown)


hughgrigg:
@coderabbitai rate limit

PR #366, opened 2026-08-04
feat: capture item changes on a simulated DynamoDB table stream

### LINKED ISSUE #339: Capture item changes on a simulated DynamoDB table stream
## Problem

Simulated DynamoDB has no streams. `StreamSpecification` is refused by name on `CreateTable`
(`src/service/dynamodb/command/table/sim-dynamodb-unsimulated-table-input.ts:37-45`) and on
`UpdateTable` (`sim-dynamodb-unsimulated-update-input.ts:19-25`).

Change data capture is one of the main reasons people reach for DynamoDB: an aggregation counter, a
search index kept in step, an audit log, a transactional outbox. None of those can be tested against
the simulator at all, and the table is where the change actually happens, so nothing downstream can
be built until the table captures it.

## Desired behaviour

A table created or updated with a `StreamSpecification` captures every item change as a stream
record, with the images its `StreamViewType` selects. `DescribeTable` reports the specification, the
latest stream ARN and the latest stream label.

## Example

```typescript
  new CreateTableCommand({
    TableName: "orders",
    KeySchema: [{ AttributeName: "orderId", KeyType: "HASH" }],
    AttributeDefinitions: [{ AttributeName: "orderId", AttributeType: "S" }],
    StreamSpecification: {
      StreamEnabled: true,
      StreamViewType: "NEW_AND_OLD_IMAGES",
    },
  }),
);
```

## Proposed scope

**Where capture goes.** `SimDynamoDbTableItems.put` (`table/sim-dynamodb-table-items.ts:17`) and
`.remove` (`:49`), plus a new `expire(key)` for the time to live caller. That pair is the only place
all nine item-mutation call sites pass through exactly once, and both already read the old value
before mutating, so the old image is available for nothing.

Not `SimDynamoDbTable.putItem`/`deleteItem`, despite `src/service/dynamodb/README.md:1053` calling
`putItem` "the single hook". That is true for writes and false for deletes:
`time-to-live/sim-dynamodb-table-time-to-live.ts:141` calls `this.items.remove` directly. A time to
live deletion is the one record carrying `userIdentity`, and combining time to live with a stream is
a whole architecture in its own right, so missing it would not be a small gap.

**The model.** A new `src/service/dynamodb/stream/` holding the table stream (hung off the table as
one public field the way `timeToLive` is), the stream, its single shard, the record, the view type
matrix, the key projection, the record size, the sequence numbers, the stream ARN, the checked
`StreamViewType` union and reader, the stream activity twin of `SimSqsQueueActivity`, and the stream
store.

**The lifecycle.** Accept `StreamSpecification` on `CreateTable` and `UpdateTable`, enforcing the
AWS rules: `StreamViewType` required when `StreamEnabled` is true, `ValidationException` on enabling
a stream that already exists or disabling one that does not, no in-place `StreamViewType` change,
and a fresh label and ARN on every enable. Report `StreamSpecification`, `LatestStreamArn` and
`LatestStreamLabel` from `table/sim-dynamodb-table-description.ts`, which covers `CreateTable`,
`DescribeTable`, `UpdateTable` and `DeleteTable` in one edit.

## Acceptance criteria

- [ ] A `PutItem` into a streamed table captures an `INSERT` record, and a second `PutItem` for the
      same key captures a `MODIFY` with both images.
- [ ] A `DeleteItem` captures a `REMOVE` carrying the old image and no new image.
- [ ] A time to live expiry captures a `REMOVE` carrying
      `userIdentity: { type: "Service", principalId: "dynamodb.amazonaws.com" }`, and an application
      delete carries none.
- [ ] All four `StreamViewType` values select the right images, including the degenerate shapes:
      `REMOVE` under `NEW_IMAGE` and `INSERT` under `OLD_IMAGE` are keys only.
- [ ] Nothing is captured for: a table with no stream, a write that does not change the item, a
      failed conditional write, a cancelled transaction, a delete of an absent key, or `DeleteTable`.
- [ ] Enabling a stream that already exists, disabling one that does not, and changing
      `StreamViewType` in place each raise `ValidationException`.

**Capture is eager, and that is deliberately against the house style.**
`src/service/dynamodb/README.md:196-199` records that secondary indexes are derived lazily at read
time following `SimSqsQueue.applyLifecycle`, and that reads as the pattern to follow. It cannot work
here: put A, then B, then A leaves state A but owes two `MODIFY` records, and an insert followed by
a delete leaves nothing in the table at all yet owes an `INSERT` and a `REMOVE`. Nothing readable
afterwards reconstructs either. An index is a function of current state; a stream is a log of
transitions. Comment it at the choke point and add a paragraph next to the lazy-index rationale, or
a later reader will helpfully change it back. Laziness applies only to the 24 hour retention trim,
which arrives with the reader.

**`SizeBytes` is not `SimDynamoDbItem.sizeInBytes()`.** That method implements the 400 KB item size
rule, where a number costs `ceil(significantDigits / 2) + 1` and lists and maps carry a 3 byte
overhead. The stream rule is the raw text length of the value, summed over `Keys` plus every image
the record carries together, not one image. The two agree on AWS's published sample only because
"101" is 3 either way; an eight digit number gives 5 against a correct 8. Give it its own
collaborator with a test reproducing the published AWS samples exactly, and leave `simDynamoDbValueSize`
alone, since it is correct for what it does.

**Sequence numbers should be a counter, not a clock.** Tests write several items inside one
millisecond, and `BackgroundTasks` defaults to a real clock. A monotonic per-stream counter rendered
at a fixed width, offset so there is never a leading zero, keeps lexicographic order equal to
numeric order. Real AWS varies the width between 21 and 40 digits, so this is a divergence in the
user's favour and belongs in Limitations. The clock is for `ApproximateCreationDateTime` only.

**`UpdateItem` with no `UpdateExpression`** returns the same object, so the choke point sees the old
and new images as identical by reference. Real DynamoDB does emit a record for that write, so
emitting one is probably right, but it should be a commented decision rather than an accident.

**One shard per stream, never split.** AWS documents each open shard as corresponding to one table
partition, and a simulated table is always one partition, so this is accurate rather than a
shortcut. The resulting total order across all keys is stronger than AWS's per-key guarantee and
belongs in Limitations.

## Out of scope

The DynamoDB Streams SDK operations, which are
https://github.com/KensioSoftware/yulin/issues/340. Any Lambda involvement. CloudFormation. Shard
splitting, parent and child lineage. The 24 hour retention trim, which arrives with the reader.
`KinesisStreamSpecification`, which stays refused.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/337 for the line budget and the
constructor ordering, and on https://github.com/KensioSoftware/yulin/issues/336, without which a
partially applied transaction starts emitting records for a transaction real AWS rejects atomically.

Enable and disable land together rather than as two issues. A release where `CreateTable` accepts a
`StreamSpecification` while `UpdateTable` still throws "DynamoDB streams are not simulated" is an
intermediate state nobody could explain, and the interesting rules are all on the `UpdateTable` side.

Both refusal files use streams as their worked example in the file header comment, so both headers
need rewording rather than just the branch removing.

Deliberately over the usual sizing steer. The negative cases are most of the test weight and should
not be trimmed: they are what makes the feature trustworthy.

hughgrigg:
A table created or updated with a StreamSpecification captures every item change as a stream record, with the images its StreamViewType selects. DescribeTable reports the specification, the latest stream ARN and label.

Capture is eager and lives in SimDynamoDbTableItems put, remove and a new expire, the one place every item mutation passes through exactly once. A stream is a log of transitions rather than a function of current state, so the lazy derivation the secondary indexes follow cannot work here.

Resolves #339

## Summary by CodeRabbit

* **New Features**
  * Added simulated DynamoDB Streams support for table creation, updates, descriptions, lifecycle management, and record capture.
  * Captures INSERT, MODIFY, REMOVE, and TTL expiration events with configurable item images, keys, sequence numbers, sizes, and metadata.
  * Added stream activity notifications and stream lookup capabilities.
* **Documentation**
  * Documented stream configuration, limitations, retention, ordering, CloudFormation behavior, and TTL interactions.
* **Tests**
  * Added comprehensive coverage for stream lifecycle, records, view types, TTL events, sizing, and notifications.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -2476,6 +2476,109 @@ inert while it was off. A removal already scheduled is checked again when it com
 overwritten with a later timestamp, or one on a table whose time to live has since been switched
 off, stays where it is.
 
+## Capturing changes with a stream
+
+A `StreamSpecification` on `CreateTable` gives a table a stream, and every change to an item is
+captured on it as a record: an `INSERT` for the first write of an item, a `MODIFY` for a write over
+one that was there, and a `REMOVE` for a deletion. `DescribeTable` reports the specification back
+along with `LatestStreamArn` and `LatestStreamLabel`.
+
+Which images a record carries is what `StreamViewType` chooses, and every record carries the keys of
+the item that changed whichever one it is:
+
+| `StreamViewType`     | `INSERT`        | `MODIFY`          | `REMOVE`        |
+| -------------------- | --------------- | ----------------- | --------------- |
+| `KEYS_ONLY`          | keys            | keys              | keys            |
+| `NEW_IMAGE`          | keys, new image | keys, new image   | keys            |
+| `OLD_IMAGE`          | keys            | keys, old image   | keys, old image |
+| `NEW_AND_OLD_IMAGES` | keys, new image | keys, both images | keys, old image |
+
+A `REMOVE` under `NEW_IMAGE` and an `INSERT` under `OLD_IMAGE` are keys and nothing else, because
+neither has the image the view type asks for. The record is still written: it is how a reader learns
+that the change happened at all.
+
+```typescript sim-dynamodb-stream-specification
+/**
+ * A table capturing its item changes on a stream.
+ */
+
+import {
+  CreateTableCommand,
+  DescribeTableCommand,
+  PutItemCommand,
+  UpdateTableCommand,
+} from "@aws-sdk/client-dynamodb";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
+
+await dynamoDb.createTable(
+  new CreateTableCommand({
+    TableName: "OrdersTable",
+    KeySchema: [{ AttributeName: "orderId", KeyType: "HASH" }],
+    AttributeDefinitions: [{ AttributeName: "orderId", AttributeType: "S" }],
+    BillingMode: "PAY_PER_REQUEST",
+    StreamSpecification: {
+      StreamEnabled: true,
+      StreamViewType: "NEW_AND_OLD_IMAGES",
+    },
+  }),
+);
+await simAws.backgroundTasksComplete();
+
+const described = await dynamoDb.describeTable(
+  new DescribeTableCommand({ TableName: "OrdersTable" }),
+);
+
+console.log(described.Table?.StreamSpecification?.StreamViewType); // "NEW_AND_OLD_IMAGES"
+console.log(described.Table?.LatestStreamArn?.includes("/stream/")); // true
+
+// Every write from here is captured on the stream.
+await dynamoDb.putItem(
+  new PutItemCommand({
+    TableName: "OrdersTable",
+    Item: { orderId: { S: "order-1" }, total: { N: "101" } },
+  }),
+);
+
+// Switching the stream off keeps what it captured, and keeps naming it.
+await dynamoDb.updateTable(
+  new UpdateTableCommand({
+    TableName: "OrdersTable",
+    StreamSpecification: { StreamEnabled: false },
+  }),
+);
+await simAws.backgroundTasksComplete();
+
+const withoutStream = await dynamoDb.describeTable(
+  new DescribeTableCommand({ TableName: "OrdersTable" }),
+);
+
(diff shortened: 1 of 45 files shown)


hughgrigg:
@coderabbitai full review

PR #369, opened 2026-08-04
feat: read simulated DynamoDB stream records with the Streams API

### LINKED ISSUE #340: Read simulated DynamoDB stream records with the DynamoDB Streams API
## Problem

Once a simulated table captures stream records, nothing can read them. There is no way for a test to
see what a table captured, and no read path for a Lambda event source mapping to poll.

## Desired behaviour

The four DynamoDB Streams operations work against a simulated table's stream:
`ListStreams`, `DescribeStream`, `GetShardIterator` and `GetRecords`.

## Example

```typescript
const { Streams } = await simAws
  .dynamoDbStreams()
  .listStreams(new ListStreamsCommand({ TableName: "orders" }));

const { Records } = await simAws.dynamoDbStreams().getRecords(
  new GetRecordsCommand({ ShardIterator: iterator }),
);

console.log(Records?.[0]?.dynamodb?.NewImage);
```

## Proposed scope

- `src/service/dynamodb/command/stream/` with one directory per operation, each holding local
  structural types, a handler and a colocated test.
- All four shard iterator types: `TRIM_HORIZON`, `LATEST`, `AT_SEQUENCE_NUMBER` and
  `AFTER_SEQUENCE_NUMBER`.
- `GetRecords` capping at 1000 records. An empty `Records` array with a live `NextShardIterator` is
  normal and must not be smoothed away. `NextShardIterator` is absent only when the shard is closed
  and drained.
- The 24 hour retention trim, applied lazily on read following `SimSqsQueue.applyLifecycle`, with
  `TrimmedDataAccessException` past the trim point. Retention is the one place the lazy pattern
  genuinely fits, because trimming is a pure function of the current time.
- `authorizeStream(action, streamArn, caller)` beside the existing `authorizeTable` and
  `authorizeAnyTable` in `command/authorize/`, with a stream ARN builder beside the table ARN
  builder. `ListStreams` names no stream and routes through `authorizeAnyTable`.
- `@aws-sdk/client-dynamodb-streams` as a devDependency at the same pin as the other clients, and a
  second router registration so SDK interception reaches it.

## Acceptance criteria

- [ ] `ListStreams` reports a streamed table's stream, and reports nothing for a table with no
      stream.
- [ ] `DescribeStream` reports the shard, the view type and the stream status.
- [ ] Each of the four iterator types returns the records it should, and
      `AFTER_SEQUENCE_NUMBER` excludes the record it names.
- [ ] Repeated `GetRecords` on successive `NextShardIterator` values walks the whole log and then
      returns empty without ending.
- [ ] A disabled stream closes its shard, so `GetRecords` eventually drops `NextShardIterator`, and
      the stream stays listable and readable within its retention window.
- [ ] Reading past the trim point raises `TrimmedDataAccessException`.
- [ ] A caller without the stream action on the stream ARN is denied.

**A captured record has two renderings, not one.** The Streams API declares `Identity` as
`PrincipalId`/`Type` and `ApproximateCreationDateTime` as a timestamp, while the Lambda event uses
`principalId`/`type` and epoch seconds. Keep the stored record a domain object with an explicit
Streams API renderer, so the Lambda event builder later adds its own projection rather than
reshaping a payload. Sharing one structural payload between the two surfaces makes one of them
silently wrong.

**No simulated IAM change is needed.** Statement matching is wildcard matching on the raw ARN, so a
statement naming `table/Orders` correctly fails to match `table/Orders/stream/<label>`, and the ARN
matcher already treats everything past the fifth colon as one segment.

## Out of scope

The 15 minute shard iterator expiry, which interacts with a poller's held checkpoint and is better
added alongside the poller that has to re-acquire. Shard splitting, lineage and `ShardFilter`.
Anything on the Lambda side. The two-readers-per-shard throttle and the `DescribeStream` rate limit,
which are throughput protections a single-process simulation cannot produce honestly.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/339.

Worth stating why this is not skipped in favour of a Yulin-native accessor, since nobody writes
`GetRecords` loops in Node application code. The reason is the repo's own port precedent:
`SimLambdaEventSourceQueueService` is SDK-command-shaped and carries a `SimAwsCaller`, so every poll
authorizes as the execution role through the same path a user's own call takes. A bespoke internal
read interface would need the same caller and the same authorization, so it would be `GetRecords`
under another name, plus a permanent accessor users have to learn that AWS does not have. It is also
the only way a failing delivery test can tell "nothing was captured" from "the poller did not
deliver".

hughgrigg:
ListStreams, DescribeStream, GetShardIterator and GetRecords over a table's captured changes, reached through simAws.dynamoDbStreams() or an intercepted DynamoDBStreamsClient. All four shard iterator types, an opaque self-describing iterator, and the 24 hour retention trim applied on read with TrimmedDataAccessException past the trim point.

Resolves #340

## Summary by CodeRabbit

* **New Features**
  * Added simulated DynamoDB Streams support with stream discovery, description, shard iterators, and record retrieval.
  * Added support for stream read positions, record rendering, IAM authorization, shard closure, and 24-hour retention behavior.
  * Added access through the SDK and standard simulation entry points.
* **Documentation**
  * Added comprehensive DynamoDB Streams usage documentation and an example workflow.
* **Tests**
  * Added coverage for stream operations, validation, authorization, retention, pagination, and record formats.

### DIFF
--- a/docs/sdk/README.md
+++ b/docs/sdk/README.md
@@ -243,8 +243,8 @@ which descriptors is in
 ## Supported services and Commands
 
 All simulated services support SDK interception: ACM, API Gateway v2, CloudFormation, CloudFront,
-Cognito, DynamoDB, IAM, KMS, Lambda, Route53, S3, Secrets Manager, SQS, SSM and STS. Each service's
-own docs under
+Cognito, DynamoDB, DynamoDB Streams, IAM, KMS, Lambda, Route53, S3, Secrets Manager, SQS, SSM and
+STS. Each service's own docs under
 [docs/services](../services/) list the Commands it simulates.
 
 Both kinds of gap are refused on send rather than misbehaving silently, with a different error each:
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -2576,8 +2576,129 @@ cancelled transaction, a delete of a key holding nothing, or a request the table
 the table takes its items with it rather than removing them one at a time, so nothing is captured
 for that either.
 
-Reading the records back is the DynamoDB Streams API, which is not simulated yet, so for now a
-stream is captured rather than consumed.
+## Reading a stream's records
+
+Reading the records back is the DynamoDB Streams API, which AWS puts behind a client of its own.
+`simAws.dynamoDbStreams()` is that API here, with `ListStreams`, `DescribeStream`,
+`GetShardIterator` and `GetRecords`.
+
+Reading a stream takes four calls the first time. `ListStreams` finds the stream ARN for a table,
+`DescribeStream` reports the shard, `GetShardIterator` says where on that shard to start, and
+`GetRecords` reads from there and hands back the iterator to carry on with.
+
+```typescript sim-dynamodb-stream-records
+/**
+ * Reading a table's captured changes back off its stream.
+ */
+
+import { CreateTableCommand, PutItemCommand } from "@aws-sdk/client-dynamodb";
+import {
+  DescribeStreamCommand,
+  GetRecordsCommand,
+  GetShardIteratorCommand,
+  ListStreamsCommand,
+} from "@aws-sdk/client-dynamodb-streams";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const dynamoDb = simAws.dynamoDb();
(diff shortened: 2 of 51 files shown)


hughgrigg:
@coderabbitai full review

PR #370, opened 2026-08-04
feat: deliver simulated DynamoDB stream records to a simulated Lambda…

### LINKED ISSUE #341: Deliver simulated DynamoDB stream records to a simulated Lambda function
## Problem

The point of a DynamoDB stream, for most applications, is that a Lambda function runs when the table
changes. A simulated event source mapping only accepts an SQS queue, so the whole architecture is
untestable: an aggregation counter, a search index kept in step, an audit log.

## Desired behaviour

Create an event source mapping from a simulated table's stream to a simulated Lambda function, write
an item, and the function is invoked with the DynamoDB stream event.

## Example

```typescript
await simAws.lambda().createEventSourceMapping(
  new CreateEventSourceMappingCommand({
    EventSourceArn: streamArn,
    FunctionName: "order-projector",
    StartingPosition: "TRIM_HORIZON",
  }),
);

  new PutItemCommand({ TableName: "orders", Item: marshall(order) }),
);

```

## Proposed scope

- `src/service/lambda/event-source/stream/` with the stream port and a no-streams implementation
  refusing with guidance for a standalone `SimLambda`, mirroring the queue port. The service and
  activity interfaces are declared Lambda-side and satisfied structurally by simulated DynamoDB, so
  neither service imports the other.
- A sibling poller rather than a generalisation of the SQS one. Every step differs, and
  `pollAgainAfter` actively inverts: the SQS poller reschedules when a batch came back, a stream
  reschedules when a batch failed. A template method there would have every hook overridden.
- A checkpoint, a retry backoff and the stream event builder as their own small files.
- `StartingPosition` required for a stream and refused for a queue, accepting `TRIM_HORIZON` and
  `LATEST`.
- Batch size defaulting to 100 for a stream.
- The role permission check covering `dynamodb:DescribeStream`, `GetRecords` and `GetShardIterator`
  on the stream ARN plus `dynamodb:ListStreams` on `*`, which is what both the AWS managed policy
  and CDK's own grant do.
- Bounded retry: on exhaustion, discard the batch, advance the checkpoint and continue.

## Acceptance criteria

- [ ] A `PutItem` into a streamed table with a mapping invokes the function with the stream event
      after one `await simAws.backgroundTasksComplete()`.
- [ ] The event's key casing is exactly right: `Records` Pascal, `eventID` with a capital `ID`,
      `dynamodb` all lower case, `eventSourceARN` with a capital `ARN`, the inner block Pascal, and
      `userIdentity` camel.
- [ ] Several items written in one `Promise.all` are each delivered exactly once.
- [ ] A record behind a failing batch is not delivered until that batch succeeds or is discarded,
      and is delivered afterwards.
- [ ] `AT_TIMESTAMP` is refused by name.
- [ ] A mapping whose execution role lacks the stream actions is refused at creation.

**Two ways this hangs the test run rather than failing it, and both must be closed before the
implementation.** A stream batch that fails retries from an unmoved checkpoint, unlike an SQS batch
that goes back on a queue. `SimLambdaEventSourcePollSchedule.at()` falls back to `now()` for any
instant at or before the present, and `afterSeconds(0)` schedules at the current instant. Either
makes `simAws.clock().advanceBy` never return, because `runDueTasksUpTo` deliberately keeps
dispatching work that falls due inside the interval. Routing a retry through `schedule.now()`
instead goes to `background.schedule()`, which `backgroundTasksComplete()` waits on, so that never
returns either. Both surface as vitest's ten second timeout pointing at nothing.

The retry delay must therefore be strictly positive and increasing, in its own collaborator, and the
two guard tests are written first: an hour's clock advance against a permanently throwing handler
returns, and `backgroundTasksComplete()` returns while a mapping is mid-backoff. Present the
cadence in the docs as a simulator constraint, not as AWS behaviour: AWS documents no backoff for
stream mapping retries.

**A dropped wake-up.** `SimLambdaEventSourcePollSchedule.now()` clears its scheduled flag at the
start of the task, not the end, so a write landing while a poll is in flight successfully schedules
a second poll. If that second poll returns early on a one-poll-in-flight guard and nothing
reschedules, those records sit undelivered until an unrelated later write happens to wake the
poller. The guard has to remember it was asked and reschedule from the `finally` block. The existing
table factories already write in one `Promise.all`, so this would show up as a flake rather than a
clean failure.

**Do not implement the documented `BatchSize` above 10 batching-window rule.** It is in the API
reference, but for a DynamoDB source the same docs give `BatchSize` 100 and
`MaximumBatchingWindowInSeconds` 0 as simultaneous defaults, and CDK hardcodes a batch size of 100
with no window on every `DynamoEventSource`. Enforcing it would refuse the default CDK stack that
deploys fine on AWS. Note the non-enforcement in Limitations.

**The self-feeding write cascade.** A handler writing an aggregate back into its own streamed table
is the most common DynamoDB Streams architecture and would spin `backgroundTasksComplete()` forever.
Guard it with a named error carrying the function, the stream and the table, and make the primary
docs example write into a **second** table so the published code cannot demonstrate the trap.

**On exhaustion, discard rather than park.** AWS is explicit: "If the error handling measures fail,
Lambda discards the records and continues processing batches from the stream." A shard that parks
permanently is a simulator invention, and a test asserting it would be asserting an artifact.

## Out of scope

`ReportBatchItemFailures` for a stream, which is
https://github.com/KensioSoftware/yulin/issues/342 because the rule genuinely differs from the SQS
one. `BisectBatchOnFunctionError`, `DestinationConfig`, `ParallelizationFactor`,
`TumblingWindowInSeconds` and `FilterCriteria`, all of which stay refused with reasons naming the
stream case. The 15 minute iterator expiry. CloudFormation.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/340 for the read path and on
https://github.com/KensioSoftware/yulin/issues/338 for the per-source split.

The stream event source ARN parser must anchor on `table/<name>/stream/<label>` carefully: the ISO
timestamp label contains two colons of its own, so splitting on colons naively gets it wrong.

Extend `test/lambda/event-source-fixture.ts` rather than writing a new fixture; its recording
handler is hard-typed to the SQS event and needs a type parameter.

hughgrigg:
… function

An event source mapping from a table's stream to a function invokes it with the DynamoDB stream event when the table changes. A failing batch blocks its shard until it is through or discarded, and a function writing back into its own source table is refused rather than delivered its own writes forever.

Resolves https://github.com/KensioSoftware/yulin/issues/341

## Summary by CodeRabbit

* **New Features**
  * Added support for delivering DynamoDB Streams records to Lambda through event source mappings.
  * Added configurable starting positions, including `TRIM_HORIZON` and `LATEST`.
  * Added AWS-shaped DynamoDB stream event payloads and attribute image types.
  * Added stream batching, retries, checkpointing, shard ordering, and lifecycle handling.
  * Added safeguards against self-triggering write cascades.

* **Documentation**
  * Added setup guidance, permissions, limitations, event formats, retry behavior, and a complete usage example.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -2704,6 +2704,13 @@ it means the oldest record still there, whatever has gone.
 A stream stays listable and readable after its table switches it off, and after everything on it has
 been trimmed. A trimmed stream reads as empty rather than as missing.
 
+### Delivering a stream to a Lambda function
+
+Nobody writes `GetRecords` loops in application code. What most applications do with a stream is
+have a Lambda function run on it, which is a
+[Lambda event source mapping](../lambda/#triggering-a-function-from-a-dynamodb-stream "Simulated Lambda event source mapping docs"):
+create the mapping, write to the table, and the function is invoked with the changes.
+
 ## Numbers
 
 A DynamoDB number carries up to 38 significant digits, where a JavaScript number carries about 15.
@@ -3329,7 +3336,8 @@ nothing is written.
   elapsing. An item whose window goes by while a running-mode clock tracks the host stays where it
   is until something moves the clock. A simulated DynamoDB constructed standalone as
   `new SimDynamoDb()` has no clock control at all, so nothing there ever expires.
-- A Lambda event source mapping cannot read a stream yet. A test polls one with `GetRecords` itself.
+- A Lambda event source mapping delivers a stream to a function, but nothing else consumes one. A
+  Kinesis Data Streams destination is not simulated.
 - A `StreamSpecification` in a CloudFormation template still skips the table, so a streamed table
   has to be created through `CreateTable`.
 - A shard iterator never expires. Real DynamoDB gives one 15 minutes and then answers
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -2706,10 +2706,11 @@ been trimmed. A trimmed stream reads as empty rather than as missing.
 
 ### Delivering a stream to a Lambda function
 
-Nobody writes `GetRecords` loops in application code. What most applications do with a stream is
-have a Lambda function run on it, which is a
+Most applications consume a stream by having a Lambda function run on it rather than by polling it
+themselves, which is a
 [Lambda event source mapping](../lambda/#triggering-a-function-from-a-dynamodb-stream "Simulated Lambda event source mapping docs"):
-create the mapping, write to the table, and the function is invoked with the changes.
+create the mapping, write to the table, and the function is invoked with the changes. The
+`GetRecords` loop above is still there for a consumer that wants to read a stream directly.
 
 ## Numbers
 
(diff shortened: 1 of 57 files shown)


hughgrigg:
@coderabbitai full review

PR #371, opened 2026-08-04
feat: report individual record failures from a simulated DynamoDB str…

### LINKED ISSUE #342: Report individual record failures from a simulated DynamoDB stream batch
## Problem

A function consuming a DynamoDB stream can report which records in a batch it failed on, through
`FunctionResponseTypes: ["ReportBatchItemFailures"]` and a `batchItemFailures` array. Simulated
Lambda supports that for SQS and not for a stream.

Reusing the SQS implementation would be silently wrong. `SimLambdaSqsBatchResponse` partitions the
batch into the entries the function named and the entries it did not, which is correct for a queue
where each message is deleted or returned independently. The stream rule is different, and AWS
states it plainly:

> If the batchItemFailures array contains multiple items, Lambda uses the record with the lowest
> sequence number as the checkpoint. Lambda then retries all records starting from that checkpoint.

That is a checkpoint rewind, not a set partition. Under the SQS rule, a report naming an early
record would checkpoint past later records the function never reached, losing them with no error.

## Desired behaviour

A stream mapping created with `ReportBatchItemFailures` rewinds to the lowest reported sequence
number and redelivers everything from there, including records that succeeded.

## Proposed scope

A sibling batch response for streams, answering with one checkpoint sequence number rather than two
lists, built over the shared report reader extracted earlier. Keep the existing defence that a
report naming something outside the batch fails the whole batch, which is correct for both sources.

## Acceptance criteria

- [ ] A report naming only the last record redelivers only that record.
- [ ] A report naming an early record redelivers it and everything after it, including records that
      succeeded.
- [ ] An empty or absent `batchItemFailures` is a complete success.
- [ ] An empty string or null `itemIdentifier` is a complete failure.
- [ ] A report naming a sequence number not in the batch fails the whole batch.
- [ ] A response is ignored entirely unless `ReportBatchItemFailures` is in `FunctionResponseTypes`.
- [ ] SQS batch response behaviour is unchanged.

Redelivering records that already succeeded is real AWS behaviour, not a simulation artifact, and it
is the reason stream consumers have to be idempotent. The docs should say so plainly rather than
leaving a reader to assume the simulator is being imprecise.

## Out of scope

`BisectBatchOnFunctionError`, which interacts with the retry quota (splitting does not consume it)
and needs its own issue. Any change to SQS batch response behaviour.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/341.

Kept separate from the delivery work both because the rule deserves its own review and because it
keeps that issue at a reviewable size.

hughgrigg:
…eam batch

A stream mapping created with ReportBatchItemFailures now rewinds to the lowest sequence number the function reported and delivers everything from there again, rather than partitioning the batch the way a queue mapping does. Splitting a stream batch would checkpoint past records the function never reached.

Resolves https://github.com/KensioSoftware/yulin/issues/342

## Summary by CodeRabbit

* **New Features**
  * Added support for partial batch failure reporting in DynamoDB stream event sources.
  * Failed records can now trigger retries from the earliest reported sequence number, with subsequent records redelivered.
  * Retry behavior respects existing backoff, retry limits, and eventual record discard rules.
  * Added support for configuring and testing batch failure responses.

* **Documentation**
  * Updated guidance for stream handling, validation, availability, and retry behavior.

### DIFF
--- a/docs/services/lambda/README.md
+++ b/docs/services/lambda/README.md
@@ -539,9 +539,10 @@ the sender sees is the message coming back.
 
 ### Reporting individual message failures
 
-A mapping created with `FunctionResponseTypes: ["ReportBatchItemFailures"]` takes the
+A queue mapping created with `FunctionResponseTypes: ["ReportBatchItemFailures"]` takes the
 `batchItemFailures` list the handler returns: the message ids named in it go back to the queue, and
-the rest of the batch is deleted.
+the rest of the batch is deleted. A stream mapping takes the same list and does something else with
+it, which is [reporting individual record failures](#reporting-individual-record-failures).
 
 ```typescript
 await simAws.lambda().createEventSourceMapping(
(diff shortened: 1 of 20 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #372, opened 2026-08-04
feat: create a simulated DynamoDB table with a stream from CloudForma…

### LINKED ISSUE #343: Create a simulated DynamoDB table with a stream from CloudFormation
## Problem

`StreamSpecification` is on the unsimulated property list for `AWS::DynamoDB::Table`
(`src/service/dynamodb/cfn/table/sim-cfn-dynamodb-table-property-rules.ts:44`), so a template
declaring a streamed table skips the table entirely and the rest of the stack deploys without it.
That is the right behaviour while streams are unsimulated and the wrong one once they are.

## Desired behaviour

An `AWS::DynamoDB::Table` carrying a `StreamSpecification` deploys a table with a stream, through
the ordinary `CreateTable` command path so a template and an SDK caller get identical validation.
`Fn::GetAtt StreamArn` returns the stream ARN.

## Example

```yaml
OrdersTable:
  Type: AWS::DynamoDB::Table
  Properties:
    TableName: orders
    KeySchema:
      - AttributeName: orderId
        KeyType: HASH
    AttributeDefinitions:
      - AttributeName: orderId
        AttributeType: S
    StreamSpecification:
      StreamViewType: NEW_AND_OLD_IMAGES
```

## Proposed scope

- Move `StreamSpecification` from the unsimulated property names to the simulated ones, and reword
  the file header comment, which uses streams as its worked example.
- Teach the properties reader to read it. **CloudFormation's `StreamSpecification` has no
  `StreamEnabled` field at all**: it is exactly `{ResourcePolicy?, StreamViewType}` with
  `StreamViewType` required. Passing the block straight through to `CreateTable` would produce
  `StreamEnabled: undefined` and a silently stream-less table. Synthesize `StreamEnabled: true` from
  the property's presence.
- Refuse the nested `StreamSpecification.ResourcePolicy`, as the table-level `ResourcePolicy`
  already is. It is a different property and would otherwise be silently ignored.
- Make `Fn::GetAtt StreamArn` conditional in the value adapter: the stream ARN when a stream is on,
  and a refusal naming the table and the missing `StreamSpecification` when it is not.

## Acceptance criteria

- [ ] A template declaring a streamed table deploys it, and a write to the deployed table captures a
      stream record.
- [ ] `Fn::GetAtt StreamArn` resolves to the stream ARN.
- [ ] `Fn::GetAtt StreamArn` on a table with no `StreamSpecification` is refused, naming the table.
- [ ] `KinesisStreamSpecification` still skips the resource, with a test asserting it.
- [ ] `StreamSpecification.ResourcePolicy` fails the stack by name.

**The `StreamArn` refusal is a deliberate departure from the stand-in fallback**, so it is worth
stating why. The general rule is that a skipped Resource answers `Fn::GetAtt` with a stand-in, which
is what stops one unsupported Resource cascading into every Resource that references it
(https://github.com/KensioSoftware/yulin/issues/335). That reasoning does not apply here: the table
is supported and did deploy, so there is no cascade to protect against, and the attribute is
genuinely absent rather than unsimulated. Real CloudFormation rejects the template outright. The
divergence is in timing, since AWS refuses at template validation and this refuses when the
attribute is asked for, and that belongs in the docs.

If that turns out to break templates in practice, falling back to the stand-in is the alternative
and this criterion is the one to revisit.

**`docs/services/dynamodb/README.md:2792-2793` currently claims `Fn::GetAtt … StreamArn` is refused
by name**, which is not what happens today. Issue #335 corrects it to describe the stand-in; this
issue then rewrites it again to describe the conditional refusal. Worth doing in this order rather
than trying to skip a step, because the two issues can land far apart.

**Moving a property name between the two lists changes stack outcomes.** Whether a resource is
skipped or the stack fails is decided by substring-matching the error message
(`"Unsupported sim"` and `"CloudFormation"`). Move only the name and do not touch the error
factories.

CloudFormation replaces the table when `StreamViewType` changes, whereas `UpdateTable` refuses the
change in place. Worth a docs sentence, since it is a real difference between the two paths.

## Out of scope

`AWS::Lambda::EventSourceMapping`, which is
https://github.com/KensioSoftware/yulin/issues/344. `KinesisStreamSpecification`, which stays
refused because Kinesis Data Streams is not simulated.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/339.

No CloudFormation ordering work is needed: `Fn::GetAtt` already contributes an implicit dependency,
so a mapping referencing the table's stream ARN waits for the table.

hughgrigg:
…tion

An AWS::DynamoDB::Table carrying a StreamSpecification now deploys a table with a stream through CreateTable, rather than being skipped. CloudFormation's StreamSpecification has no StreamEnabled field, so it is synthesized from the property being declared. Fn::GetAtt StreamArn gives the stream ARN, and is refused naming the table when the table has no stream.

Resolves https://github.com/KensioSoftware/yulin/issues/343

## Summary by CodeRabbit

* **New Features**
  * Added CloudFormation support for creating DynamoDB tables with streams.
  * Stream records can be captured and read after table updates.
  * `StreamArn` can now be retrieved for tables with configured streams.
  * Added documentation and an example workflow for deploying streamed tables and reading `INSERT` events.

* **Bug Fixes**
  * Invalid or incomplete stream configurations now produce clear validation errors.
  * Unsupported stream options are handled consistently.
  * Changes to the stream view type are rejected instead of unexpectedly replacing the table.

### DIFF
--- a/docs/services/dynamodb/README.md
+++ b/docs/services/dynamodb/README.md
@@ -3012,9 +3012,9 @@ console.log(stack.outputs.get("OrdersTableArn")?.value);
 
 The properties that are read are `TableName`, `KeySchema`, `AttributeDefinitions`, `BillingMode`,
 `ProvisionedThroughput`, `TableClass`, `DeletionProtectionEnabled`, `Tags`,
-`GlobalSecondaryIndexes`, `LocalSecondaryIndexes` and `TimeToLiveSpecification`. All but the last are
-passed to `CreateTable` rather than applied here, so a value the template gets wrong fails the same
-way it would for an SDK caller.
+`GlobalSecondaryIndexes`, `LocalSecondaryIndexes`, `StreamSpecification` and
+`TimeToLiveSpecification`. All but the last are passed to `CreateTable` rather than applied here, so
+a value the template gets wrong fails the same way it would for an SDK caller.
 
 `TimeToLiveSpecification` is applied after the table is created, through `UpdateTimeToLive`. Real
 `CreateTable` has no parameter for it either, so real CloudFormation makes the table and then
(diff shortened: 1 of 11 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #373, opened 2026-08-04
feat: create a simulated Lambda event source mapping for a DynamoDB s…

### LINKED ISSUE #344: Create a simulated Lambda event source mapping for a DynamoDB stream from CloudFormation
## Problem

CDK's `fn.addEventSource(new DynamoEventSource(table, { startingPosition }))` emits an
`AWS::Lambda::EventSourceMapping` carrying `StartingPosition`, which is on the unsimulated property
list (`src/service/lambda/cfn/event-source-mapping/sim-cfn-lambda-event-source-mapping-property-rules.ts:40-41`).
So the stack that wires a table's stream to a function is the one thing that still does not deploy,
even once every other piece is in place.

`StartingPosition` is the **only** property a default `DynamoEventSource` emits that the simulator
refuses. `BatchSize` 100, `EventSourceArn` and `FunctionName` all already pass.

## Desired behaviour

A CDK stack with a streamed table and a `DynamoEventSource` deploys, and writing an item into the
deployed table invokes the deployed function.

## Example

```typescript
const table = new dynamodb.Table(this, "Orders", {
  partitionKey: { name: "orderId", type: dynamodb.AttributeType.STRING },
  stream: dynamodb.StreamViewType.NEW_AND_OLD_IMAGES,
});

projector.addEventSource(
  new DynamoEventSource(table, {
    startingPosition: lambda.StartingPosition.TRIM_HORIZON,
  }),
);
```

## Proposed scope

- Remove `StartingPosition` and `StartingPositionTimestamp` from the flat unsimulated set and pass
  them through to the command, which now decides per source. This matches the stated design that the
  CloudFormation properties class validates nothing and the command owns the rules.
- Reorder the property assertion so `EventSourceArn` is read first. It currently runs before the ARN
  is known, which is the same ordering problem the command itself had.
- Teach the properties reader to read `StartingPosition`.

## Acceptance criteria

- [ ] A hand-written template with a streamed table and an event source mapping deploys, and a write
      to the table reaches the function.
- [ ] A local test synthesising a CDK stack with `DynamoEventSource` deploys and delivers.
- [ ] `StartingPosition` on an SQS mapping is still refused, and its absence on a stream mapping is
      refused.
- [ ] `StartingPositionTimestamp` is refused by name.

The CDK test is worth having as a local test rather than only an isolated one, because it exercises
CDK's own inline grant against the role permission check: `dynamodb:ListStreams` on `*` as one
statement and the other three actions on the stream ARN as another. A single-ARN permission check
would refuse a correctly granted role, so this is the test that proves the split ARN handling works.

Worth a Limitations note: simulated IAM does not model AWS-managed policy ARNs, so a hand-written
template or a SAM application that attaches `AWSLambdaDynamoDBExecutionRole` gets a role with no
stream permissions and the mapping is refused at creation. CDK is unaffected because it writes an
inline policy instead. That is a real gap between two declaration paths and users will hit it.

## Out of scope

`FilterCriteria`, `ParallelizationFactor`, `BisectBatchOnFunctionError`, `TumblingWindowInSeconds`
and `DestinationConfig`, all of which a non-default `DynamoEventSource` emits and which stay refused
by name.

## Implementation notes

Depends on https://github.com/KensioSoftware/yulin/issues/341 for delivery and on
https://github.com/KensioSoftware/yulin/issues/343 for the table side.

hughgrigg:
…tream from CloudFormation

StartingPosition and StartingPositionTimestamp are read off AWS::Lambda::EventSourceMapping and passed to CreateEventSourceMapping rather than refused by name, so the command decides per source: a stream mapping needs a position and a queue mapping is refused for naming one. The event source ARN is read before anything else about the Resource is judged, as the command reads it first.

A CDK stack with a streamed table and a DynamoEventSource now deploys as it is synthesised, and a write to the table reaches the function.

Resolves https://github.com/KensioSoftware/yulin/issues/344

### DIFF
--- a/docs/services/lambda/README.md
+++ b/docs/services/lambda/README.md
@@ -570,7 +570,9 @@ or an empty `batchItemFailures` list, has handled the whole batch.
 `AWS::Lambda::EventSourceMapping` deploys the same thing, which is what CDK's
 `fn.addEventSource(new SqsEventSource(queue))` emits. `EventSourceArn` and `FunctionName` accept the
 `Fn::GetAtt` and `Ref` values a template gives them, `Ref` on the mapping returns its UUID, and
-`Fn::GetAtt` exposes `Id` and `EventSourceMappingArn`.
+`Fn::GetAtt` exposes `Id` and `EventSourceMappingArn`. A queue mapping has no `StartingPosition`, and
+a template naming one is refused. The same Resource deploys a
+[stream mapping](#stream-mappings-in-templates), which has to have one.
 
 ## Triggering a function from a DynamoDB stream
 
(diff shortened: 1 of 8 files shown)


hughgrigg:
@coderabbitai rate limit

PR #375, opened 2026-08-04
feat: control simulated Rekognition results by image name and content…

### LINKED ISSUE #317: Control simulated Rekognition results by image name and content hash
## Problem

Yulin has no simulated Rekognition. There is no `src/service/rekognition/`, and `resolveSimSdkCommandRouter` has no case for the Rekognition client, so intercepting a `RekognitionClient` throws `SimSdkUnknownServiceError`. An application that moderates user uploads cannot be tested against Yulin at all.

The rest of that pipeline is simulated now. A simulated S3 Bucket notifies a simulated Lambda function when an Object is created, so the upload and the handler that runs on it can both be exercised for real. Rekognition is the part in the middle that is missing.

Rekognition is also a service where the interesting behaviour is not the call, it is what the call returns. A test needs to say "this image fails moderation and this one does not" without any real image analysis happening, in the same way simulated ACM issues certificates without producing real TLS certificates.

## Desired behaviour

A `SimRekognition` facade reachable as `simAws.rekognition()`, scoped to an account and region, answering `DetectModerationLabels` for images supplied either as `Image.Bytes` or as `Image.S3Object` pointing at a simulated S3 object.

Results come from an ordered rule list. A rule matches an exact S3 object name, an exact content hash, or anything at all. Yulin registers a default rule so the service is useful before anything is configured, and the user overrides it or adds per-image rules. There is no pattern syntax: names and hashes match exactly, which keeps precedence to a single sentence.

Rules are grouped by operation rather than hung off the service facade, so `SimRekognition` stays thin as operations are added.

## Example

```typescript
const moderation = simAws.rekognition().moderation();

// Every image is clean unless a more specific rule matches.
moderation.byDefault({ labels: [] });

// By S3 object name.
moderation.onName("uploads/nsfw.jpg", { labels: ["Explicit Nudity"] });

// By content hash, for systems that generate their own object keys.
moderation.onHash("9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08", {
  labels: ["Violence"],
});

const { ModerationLabels } = await simAws.rekognition().detectModerationLabels(
  new DetectModerationLabelsCommand({
    Image: { S3Object: { Bucket: "uploads", Name: "nsfw.jpg" } },
    MinConfidence: 80,
  }),
);
```

## Proposed scope

- `DetectModerationLabels` only. `Image` and `MinConfidence` are simulated. `HumanLoopConfig` and `ProjectVersion` are refused by name, following the `SimApiGatewayV2UnsimulatedInput` allowlist pattern, so the response never carries `HumanLoopActivationOutput` or a response-level `ProjectVersion`. Silently ignoring `ProjectVersion` would make a custom moderation adapter appear applied here and unapplied in production.
- Rule precedence, in one sentence: an exact hash rule wins, then an exact name rule, then the default. An image supplied as `Image.Bytes` has no name, so it consults hash rules and then the default.
- The content hash is sha256 of the image bytes as received, lowercase hex, with no normalisation. Export a helper so a user with a fixture file can obtain the digest without hand-rolling `node:crypto`, and document that re-encoding an image between upload and detection changes it.
- `MinConfidence` defaults to 50 and compares inclusively. Read it with an explicit `undefined` check so an explicit `0` is not silently promoted to 50, and reject values outside 0 to 100 with `InvalidParameterException`.
- A declared label name expands to its full ancestor chain from a static taxonomy, so declaring `"Explicit Nudity"` returns the correct ancestors, `ParentName` values and `TaxonomyLevel`. Top-level labels carry `ParentName` of `""`, present rather than omitted. Every label in one chain shares that chain's confidence, and `MinConfidence` filters whole chains, so a surviving label never names an absent parent. Build the taxonomy from AWS's published label list rather than transcribing it from prose.
- `ModerationModelVersion` is a constant pinned in the same file as the taxonomy it describes, so the two cannot drift apart.
- Reading `Image.S3Object` goes through the existing `SimS3GlobalRegistry`, so a bucket in another account resolves, as real Rekognition reads across accounts given the right policy.
- Order of operations: validate the input, sequence background work, authorize the Rekognition action, read from S3 as the caller, then resolve the rule. Authorizing first means a caller lacking `rekognition:DetectModerationLabels` sees a denial rather than an S3 error, and goes looking at the right policy.
- A missing or unreadable S3 object becomes `InvalidS3ObjectException`, as on AWS, with the underlying simulator error attached as `cause` so a missing `s3:GetObject` grant is still diagnosable.
- Image format comes from PNG and JPEG magic bytes. A content type given to `PutObject` is kept as a lowercase `content-type` metadata key and is absent unless the uploader supplied one, so the stored content type is deliberately not consulted. Record this in Limitations, along with the fact that an existing test storing a placeholder string body will now get `InvalidImageFormatException`.
- IAM authorization on `rekognition:DetectModerationLabels` against `*`, through the shared `SimIamInterServiceAuthZ`. Denials throw a Rekognition-specific `AccessDeniedException` with HTTP status 400, matching real Rekognition, following the `SimS3AccessDenied` precedent rather than leaking the shared IAM error name.
- Wiring: the `simAws.rekognition()` accessor in `sim-aws-service-accessors.ts`, account and region scope registration, the SDK command router case, the `@aws-sdk/client-rekognition` devDependency, the `./rekognition` package export, the `docs/tsconfig.json` path entry, `src/service/rekognition/README.md`, `docs/services/rekognition/README.md` with labelled examples, and entries in `docs/README.md`, the root `README.md` and `docs/sdk/README.md`.
- A docs section and test for the flow this is aimed at, which simulated S3 event notifications now make assemblable end to end: an upload into a simulated Bucket whose notification configuration invokes a simulated Lambda function, which calls `DetectModerationLabels` on the object the event names. Rules registered on `simAws.rekognition()` are found only when the account and region match the ones the call resolves to, which for a Lambda is the function's own scope.

## Out of scope

`DetectLabels`, `DetectFaces`, face collections, text detection, video jobs, CloudFormation resource types, and serving Rekognition over `serveSimAws`. Face collections are a natural later addition once this baseline exists.

## Implementation notes

Rekognition's S3 key field is `Name`, not `Key`. The response has exactly three members here: `ModerationLabels`, `ModerationModelVersion` and `ContentTypes`.

The docs example is the real pipeline rather than a substitute: `PutBucketNotificationConfigurationCommand` with a `LambdaFunctionConfigurations` entry for `s3:ObjectCreated:*`, an `AddPermissionCommand` letting `s3.amazonaws.com` invoke the function, `PutObject`, then `await simAws.backgroundTasksComplete()`. A handler that writes a moderated or resized copy back into the Bucket that triggered it notifies itself forever, so any example doing that filters the configuration by prefix or suffix.

hughgrigg:
… hash

Add simulated Rekognition with DetectModerationLabels, answering from results declared against an image by exact S3 object name, exact content hash, or a default. A declared label expands to its chain in the version 7.0 moderation taxonomy, and MinConfidence filters whole chains.

Resolves https://github.com/KensioSoftware/yulin/issues/317

## Summary by CodeRabbit

* **New Features**
  * Added simulated Amazon Rekognition support through the SDK and service accessors.
  * Added moderation label detection for image bytes and S3 objects.
  * Added configurable default, object-name, and image-hash moderation rules.
  * Added confidence filtering, taxonomy expansion, image validation, IAM authorization, and account/Region scoping.
  * Added Lambda/S3 upload-pipeline support for automated clean or flagged image handling.

* **Documentation**
  * Added comprehensive Rekognition service documentation and executable usage examples.

* **Tests**
  * Added coverage for moderation, validation, authorization, SDK routing, taxonomy, and upload pipelines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ npm i -D @kensio/yulin
 - [IAM](./docs/services/iam "Simulated IAM docs")
 - [KMS](./docs/services/kms "Simulated KMS docs")
 - [Lambda](./docs/services/lambda "Simulated Lambda docs")
+- [Rekognition](./docs/services/rekognition "Simulated Rekognition docs")
 - [Route53](./docs/services/route53 "Simulated Route53 docs")
 - [S3](./docs/services/s3 "Simulated S3 docs")
 - [Secrets Manager](./docs/services/secretsmanager "Simulated Secrets Manager docs")
--- a/docs/README.md
+++ b/docs/README.md
@@ -14,6 +14,7 @@ behaviour and includes example code that can be copied into tests or local devel
 - [IAM](./services/iam/ "Simulated IAM usage docs")
 - [KMS](./services/kms/ "Simulated KMS usage docs")
 - [Lambda](./services/lambda/ "Simulated Lambda usage docs")
+- [Rekognition](./services/rekognition/ "Simulated Rekognition usage docs")
 - [Route53](./services/route53/ "Simulated Route53 usage docs")
 - [S3](./services/s3/ "Simulated S3 usage docs")
 - [Secrets Manager](./services/secretsmanager/ "Simulated Secrets Manager usage docs")
--- a/docs/sdk/README.md
+++ b/docs/sdk/README.md
@@ -243,8 +243,8 @@ which descriptors is in
 ## Supported services and Commands
 
 All simulated services support SDK interception: ACM, API Gateway v2, CloudFormation, CloudFront,
-Cognito, DynamoDB, DynamoDB Streams, IAM, KMS, Lambda, Route53, S3, Secrets Manager, SQS, SSM and
-STS. Each service's own docs under
+Cognito, DynamoDB, DynamoDB Streams, IAM, KMS, Lambda, Rekognition, Route53, S3, Secrets Manager,
+SQS, SSM and STS. Each service's own docs under
 [docs/services](../services/) list the Commands it simulates.
 
 Both kinds of gap are refused on send rather than misbehaving silently, with a different error each:
--- a/docs/services/rekognition/README.md
+++ b/docs/services/rekognition/README.md
@@ -75,8 +75,6 @@ from, and each rule matches an exact S3 object name, an exact content hash, or a
  * The three kinds of rule, and which one wins.
  */
 
-import { readFileSync } from "node:fs";
-
 import { SimAws } from "@kensio/yulin";
 import { simRekognitionImageHash } from "@kensio/yulin/rekognition";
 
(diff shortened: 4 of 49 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #376, opened 2026-08-04
feat: detect labels in simulated Rekognition images

### LINKED ISSUE #318: Simulate detecting labels in simulated Rekognition images
## Problem

https://github.com/KensioSoftware/yulin/issues/317 adds the rule mechanism and `DetectModerationLabels`. Auto-tagging, alt-text and search-indexing pipelines use `DetectLabels` instead, and it is refused by name until this lands.

## Desired behaviour

`DetectLabels` resolves through the same rule list as moderation, configured under `simAws.rekognition().labels()`, and honours `MinConfidence` and `MaxLabels`.

## Example

```typescript
const labels = simAws.rekognition().labels();

labels.onName("uploads/dog.jpg", {
  labels: [{ name: "Dog", confidence: 98.2, parents: ["Animal", "Pet", "Canine"] }],
});

const { Labels } = await simAws.rekognition().detectLabels(
  new DetectLabelsCommand({
    Image: { S3Object: { Bucket: "uploads", Name: "dog.jpg" } },
    MaxLabels: 10,
  }),
);
```

## Proposed scope

- `MinConfidence` defaults to 55 here, not the 50 that moderation uses. `MaxLabels` applies after confidence filtering. Both need an explicit `undefined` check so an explicit `0` is not treated as unset.
- `Features` gates which response members appear. `GENERAL_LABELS` and `IMAGE_PROPERTIES` beyond a declared result are refused rather than returned empty.
- `Parents`, `Categories`, `Aliases` and `Instances` are declarable per label and default to empty.
- Yulin ships no general label ontology. AWS's is thousands of entries with no published enumerable table, so any curated subset would be a Yulin invention that could never be regenerated from an upstream source. State this in Limitations in one blunt line.
- No validation of declared label names against a curated list. Rejecting a valid AWS label such as `Pizza` because it is missing from a Yulin subset would be failing closed against Yulin's own incompleteness rather than against an unsimulated feature.
- The built-in default result is copied from the `DetectLabels` example response in the AWS API reference, so it is a real Rekognition response rather than an invented one. Note in the docs that which labels come back by default is a simulator convention, not something AWS would return for an arbitrary image.
- `LabelModelVersion` as a pinned constant, next to the default it belongs with.
- IAM authorization on `rekognition:DetectLabels` against `*`, with the same ordering and the same `AccessDeniedException` shape as moderation.

## Out of scope

`DetectFaces`, face collections, `DetectText`, and image properties beyond what a declared result carries.

Depends on https://github.com/KensioSoftware/yulin/issues/317

hughgrigg:
DetectLabels answers from the same rule mechanism as moderation, under simAws.rekognition().labels(), honouring MinConfidence, MaxLabels and Features. Declared labels carry their own parents, aliases, categories and instances, since Yulin ships no general label ontology.

Resolves https://github.com/KensioSoftware/yulin/issues/318

### DIFF
--- a/docs/services/rekognition/README.md
+++ b/docs/services/rekognition/README.md
@@ -1,8 +1,8 @@
 # Simulated Rekognition
 
 Simulated Rekognition answers detection calls from results declared against images, so a test can
-say which image fails moderation without any image analysis happening. No recognition of any kind is
-performed on the bytes.
+say which image fails moderation or holds a dog without any image analysis happening. No recognition
+of any kind is performed on the bytes.
 
 Rekognition-specific types are imported from the `@kensio/yulin/rekognition` subpath.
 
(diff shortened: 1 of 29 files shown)


hughgrigg:
@coderabbitai rate limit

Issue #389, opened 2026-08-05
Simulated CloudFront ignores CustomErrorResponses, so 404s never reach the error page


hughgrigg:
## Problem

A CloudFront distribution commonly maps origin errors to a single error page: ChineseBoost maps both 403 and 404 to `/404.html` so a missing URL renders the site's own error page. Under simulated CloudFront that mapping is dropped. A request for a key that is not in the S3 origin comes back as simulated S3's own `text/plain` body, `Object <key> not found in sim S3 Bucket <bucket>`, with status 404. There is no way to test that a site's error page is wired up correctly, and serving the site on localhost shows the wrong page.

## Current behaviour

`SimCloudFrontDistributionConfig` has no `CustomErrorResponses` field and nothing in the request pipeline reads one. `SimCloudFrontDistributionConfigNormalizer` spreads the incoming config, so the property does survive into the stored `distributionConfig` and `GetDistributionCommand` reads it back unchanged. That is the case simulated S3's docs say should not happen: a resource that "would look configured and behave as though it were not".

## Desired behaviour

After the origin fetch, if the response status matches a configured `ErrorCode`, simulated CloudFront requests `ResponsePagePath` from the origin and returns that body to the viewer with `ResponseCode` as the status. Without `ResponseCode` the origin's status is kept, as in real CloudFront. If the response page itself is missing from the origin, the original error response is returned rather than a second error. `ErrorCachingMinTTL` is accepted and ignored, since nothing here caches.

## Example

```ts
CustomErrorResponses: {
  Quantity: 2,
  Items: [
    { ErrorCode: 404, ResponsePagePath: "/404.html", ResponseCode: "404" },
    { ErrorCode: 403, ResponsePagePath: "/404.html", ResponseCode: "404" },
  ],
},
```

## Acceptance criteria

- [ ] `CustomErrorResponses` is part of the structural `DistributionConfig` type, accepted from `CreateDistributionCommand` and from `AWS::CloudFront::Distribution`, in both `{Quantity, Items}` and bare-array form.
- [ ] A missing S3 object with a 404 rule returns the body of `/404.html` from the origin.
- [ ] `ResponseCode` sets the status seen by the viewer; omitting it keeps the origin status.
- [ ] A status with no matching rule is passed through untouched.
- [ ] A `ResponsePagePath` that is itself missing from the origin returns the original error response.
- [ ] The CloudFront docs describe the property and what is ignored.

## Edge cases and compatibility considerations

[...]

PR #399, opened 2026-08-05
fix: apply CloudFront DefaultRootObject and CustomErrorResponses

### LINKED ISSUE #388: Simulated CloudFront ignores DefaultRootObject, so a request for / misses index.html
## Problem

`DefaultRootObject: "index.html"` is on almost every static-site distribution, and simulated CloudFront drops it. A request for `/` reaches the S3 origin as an empty object key, which is not in the bucket, so the response is simulated S3's `text/plain` 404 rather than the site's home page. Any test that fetches the root of a distribution fails, and a site served on localhost has no working front page.

## Current behaviour

`SimCloudFrontDistributionConfig` has no `DefaultRootObject` field and nothing in the request pipeline reads one. The config normalizer spreads the incoming object, so the value is stored on the Distribution and `GetDistributionCommand` reads it straight back. The distribution looks configured through the API while behaving as though it were not, which the simulated S3 docs call out as the thing to avoid.

## Desired behaviour

When a request path is exactly `/` and the Distribution has a `DefaultRootObject`, simulated CloudFront requests that object from the origin instead. As in real CloudFront this applies only to the root of the distribution, so `/blog/` is still forwarded as `/blog/` and does not become `/blog/index.html`.

## Example

```ts
await simAws.cloudFront().createDistribution(
  new CreateDistributionCommand({
    DistributionConfig: {
      CallerReference: "site",
      DefaultRootObject: "index.html",
      // origins and default cache behaviour as usual
    },
  }),
);

// GET https://distro123.cloudfront.net/ returns the body of the
// index.html object in the backing sim S3 Bucket.
```

## Acceptance criteria

- [ ] `DefaultRootObject` is part of the structural `DistributionConfig` type and is accepted from `CreateDistributionCommand` and from `AWS::CloudFront::Distribution`.
- [ ] A request for `/` on a Distribution with `DefaultRootObject` returns that object from the origin.
- [ ] A request for a sub-path ending in `/` is unchanged.
- [ ] A Distribution without the property keeps today's behaviour for `/`.
- [ ] It applies to a custom origin as well as an S3 origin, since the rewrite happens before the origin fetch.
- [ ] The CloudFront docs mention it, including the root-only limit.

Real CloudFront rejects a `DefaultRootObject` beginning with `/`. Worth rejecting here too so a template that real CloudFront would refuse is refused first, as sim CloudFront already does for viewer certificates.

The rewrite has to happen before the `viewer-request` CloudFront Function runs, or a function that inspects `request.uri` sees a different path than it would in production.

## Out of scope

`CustomErrorResponses`, which is #389. Any general directory-index behaviour, which real CloudFront does not do either.

hughgrigg:
Resolves #388 and #389.

`DefaultRootObject` and `CustomErrorResponses` were accepted from `CreateDistributionCommand` and from `AWS::CloudFront::Distribution`, stored on the Distribution and read back by `GetDistributionCommand`, while nothing in request handling looked at either. A Distribution therefore looked configured through the API and behaved as though it were not, which is what simulated S3's docs say a resource should never do.

## Default root object

A request for the Distribution root is rewritten to the named object before Behavior resolution, so a Cache Behavior path pattern and a `viewer-request` CloudFront Function both act on the object being served. Only the exact root is substituted, never a subdirectory, which is where CloudFront differs from an S3 website index document. It applies to every method the Distribution allows, and to a custom Origin as well as an S3 one.

A value beginning with a forward slash is refused at create time as `InvalidDefaultRootObject`, since in real CloudFront it answers every root request with a 403. An empty value means none, as it does in AWS.

## Custom error responses

An Origin response whose status matches an `ErrorCode` is replaced by fetching `ResponsePagePath` as a request in its own right, so the Behavior matching that path decides which Origin the page comes from and error pages can live somewhere other than the content that failed. `ResponseCode` is the status the viewer sees, which covers a single page app serving its shell with a 200. Where the response page is itself missing, the viewer gets the status from fetching it, as CloudFront documents. This runs before the `viewer-response` CloudFront Function, so a function sees the response the viewer is about to get.

Configuration CloudFront refuses is refused here: an `ErrorCode` outside the eleven codes it supports (`InvalidErrorCode`), `ResponsePagePath` and `ResponseCode` given without each other (`InvalidArgument`), and a `ResponseCode` that is not an HTTP status (`InvalidResponseCode`). A rule carrying only `ErrorCode` and `ErrorCachingMinTTL` configures error caching, which nothing here caches, so it is accepted and ignored.

## Notes

- Both properties arrive through CloudFormation as well, in CDK's bare-array form with an integer `ResponseCode` as well as the SDK's `{Quantity, Items}` with a string.
- The shared test fixture sits under `test/cloudfront/` rather than beside the tests, for the reason given in `test/sqs/queue-fixture.ts`: eslint refuses an AWS SDK import from `src/` outside a test file.
- Refusing a `ResponsePagePath` without a leading slash is a judgement call rather than a documented API constraint. CDK rejects one, and the value cannot resolve as a Distribution path, so it looked more like a mistake worth catching than a value to accept.

## Summary by CodeRabbit

* **New Features**
  * Added support for CloudFront default root objects for serving homepage content.
  * Added custom error responses with configurable error pages and response status codes.
  * Added support for static-site routing, including custom 403 and 404 pages.
  * Added validation for supported error codes, response codes, and page paths.
  * Viewer-response functions now run on custom error responses.

* **Documentation**
  * Added configuration guidance and a static-site usage example, including local serving behavior.

### DIFF
--- a/docs/services/cloudfront/README.md
+++ b/docs/services/cloudfront/README.md
@@ -63,6 +63,122 @@ const distributionCreation = await simCloudFront.createDistribution(
 console.log(distributionCreation.Distribution?.DomainName);
 ```
 
+## Static sites: default root object and error pages
+
+A static site behind CloudFront usually leans on two Distribution settings: `DefaultRootObject`, so
+a request for the site root returns the home page, and `CustomErrorResponses`, so a URL that matches
+no object returns the site's own error page rather than the Origin's. Sim CloudFront applies both,
+so a test can assert what a visitor would actually see.
+
+```typescript sim-cloudfront-static-site
+/**
+ * Serving a static site with a default root object and a custom error page.
+ */
+
+import { CreateDistributionCommand } from "@aws-sdk/client-cloudfront";
+import { CreateBucketCommand, PutObjectCommand } from "@aws-sdk/client-s3";
+
+import { SimAws } from "@kensio/yulin";
+import { serveSimAws } from "@kensio/yulin/serve";
+
+const simAws = new SimAws();
+const srv = await serveSimAws({ simAws });
+
+try {
+  const simS3 = simAws.s3();
+
+  await simS3.createBucket(new CreateBucketCommand({ Bucket: "site-bucket" }));
+
+  const pages = {
+    "index.html": "<h1>Home</h1>",
+    "404.html": "<h1>Page not found</h1>",
+  };
+
+  for (const [key, body] of Object.entries(pages)) {
+    await simS3.putObject(
+      new PutObjectCommand({
+        Bucket: "site-bucket",
+        Key: key,
+        ContentType: "text/html",
+        Body: body,
+      }),
+    );
+  }
+
+  const distributionCreation = await simAws.cloudFront().createDistribution(
+    new CreateDistributionCommand({
+      DistributionConfig: {
+        CallerReference: "static-site",
+        Comment: "Static site",
+        Enabled: true,
+        DefaultRootObject: "index.html",
+        CustomErrorResponses: {
+          Quantity: 2,
+          Items: [
+            {
+              ErrorCode: 404,
+              ResponsePagePath: "/404.html",
+              ResponseCode: "404",
+            },
+            {
+              ErrorCode: 403,
+              ResponsePagePath: "/404.html",
+              ResponseCode: "404",
+            },
+          ],
+        },
+        Origins: {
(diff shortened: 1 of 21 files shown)


hughgrigg:
@coderabbitai full review

Issue #406, opened 2026-08-05
Delete a CloudFormation resource through the service that created it


hughgrigg:
## Problem

Simulated CloudFormation knows how to create every `AWS::<Service>::*` resource type it supports, because each service owns a factory under its own `cfn/` directory and the engine resolves it by service name. There is no matching path for deletion. `SimCfnServiceResourceFactory` has a `create` method and nothing else, so the engine has no way to ask a service to remove something it made.

That is the missing half of the design rule the rest of CloudFormation follows: CloudFormation orchestrates, services create. It should also be services delete.

## Desired behaviour

`SimCfnServiceResourceFactory` gains a `delete` method alongside `create`, and every service implements it for the resource types it can create, calling its own SDK-shaped delete command rather than reaching into another service's state. A resource type with no delete path is reported the same way an unsupported type is on create, so a Stack deletion records it as skipped instead of failing silently.

The engine gains the resource-level side of it: a delete lifecycle on `SimCfnResource` mirroring the create one, and deletion in reverse dependency order, so a Bucket policy goes before its Bucket and a Role's policies go before the Role.

## Proposed scope

The factory interface change, the per-service deleters, and the engine's resource delete lifecycle and ordering. Stack-level status and the `DeleteStack` command itself belong to #397, which this unblocks.

## Dependencies

Needs the service delete commands to exist first. Those are split across their own issues by service.

## Implementation notes

Deletion ordering can reuse the dependency graph `SimCfnResource.dependencies()` already exposes for creation, reversed. `SimCfnStackResourceCreator` and its batch creator are the shape to mirror.

Issue #405, opened 2026-08-05
Delete simulated HTTP API routes, integrations and stages


hughgrigg:
## Problem

Simulated API Gateway has `DeleteApi` and `DeleteAuthorizer`, but nothing for the other three resource types an HTTP API is built from. A Route, an Integration or a Stage can be created and never removed, so a test cannot check that a route stops matching once it is gone, and a CloudFormation Stack deleting an `AWS::ApiGatewayV2::Route` has nothing to call.

## Desired behaviour

`DeleteRoute`, `DeleteIntegration` and `DeleteStage` exist, each removing only its own resource and leaving the API and the others in place. A request that matched a deleted Route no longer matches, and answers the way a simulated HTTP API answers an unmatched request.

## Acceptance criteria

- [ ] The three commands are routed and authorized as their own `apigateway:DELETE` actions.
- [ ] Deleting a Route stops it matching, without disturbing the other Routes on the API.
- [ ] Deleting a Stage removes it from the API's stages, and a request addressed to that stage no longer resolves.
- [ ] Each answers `NotFoundException` for an id the API does not have.

## Edge cases and compatibility considerations

Real API Gateway refuses to delete an Integration a Route still points at, answering `BadRequestException`. That refusal is worth keeping, because it is the ordering constraint a Stack deletion has to respect.

## Out of scope

REST APIs, which this simulator does not model, and the CloudFormation teardown that would call these.

Issue #404, opened 2026-08-05
Delete a simulated CloudFront distribution and CloudFront Function


hughgrigg:
## Problem

Simulated CloudFront can create a distribution and a CloudFront Function, but neither can be removed. A test cannot check what happens to a request once a distribution is gone, and a CloudFormation Stack holding an `AWS::CloudFront::Distribution` has nothing to call when it is deleted.

## Desired behaviour

`DeleteDistribution` and `DeleteFunction` exist and behave as the real ones do.

## Preconditions: decided

Real CloudFront makes `DeleteDistribution` awkward on purpose, in two ways this simulation does not model:

- The distribution must already be disabled. Deleting an enabled one fails with `DistributionNotDisabled`, so the real sequence is `UpdateDistribution` to `Enabled: false`, wait for deployment, then delete.
- Both operations take an `IfMatch` ETag from the preceding read, and answer `PreconditionFailed` or `InvalidIfMatchVersion` when it is wrong or missing.

The decision is to model the first and not the second. Enabled and disabled state is the precondition that actually catches people out, and it is what makes a simulated Stack deletion fail the way a real one does. ETags would touch every existing CloudFront command to introduce a versioning concept nothing else here needs, for a failure mode that is a retry rather than a design mistake.

So `IfMatch` is accepted and not checked, and that is recorded in the CloudFront docs as a divergence.

## Acceptance criteria

- [ ] `DeleteDistributionCommand` and `DeleteFunctionCommand` are routed and authorized as `cloudfront:DeleteDistribution` and `cloudfront:DeleteFunction`.
- [ ] A deleted distribution stops serving, so a request to its domain no longer resolves to it.
- [ ] Deleting something that is not there answers CloudFront's `NoSuchDistribution` and `NoSuchFunctionExists`.
- [ ] A distribution carries Enabled state, `UpdateDistribution` can change it, and deleting an enabled distribution answers `DistributionNotDisabled`.
- [ ] A disabled distribution deletes.
- [ ] The unchecked `IfMatch` is documented as a divergence in the CloudFront docs.

## Out of scope

CloudFront cache behaviour and invalidation, and the CloudFormation teardown that would call these.

Issue #403, opened 2026-08-05
Delete simulated KMS keys, Route53 hosted zones and ACM certificates


hughgrigg:
## Problem

Simulated KMS, Route53 and ACM can each create their main resource and never remove it. There is no way to delete a key, an alias, a hosted zone or a certificate, so a test cannot check what happens once one is gone, and a CloudFormation Stack holding any of them has nothing to call when it is deleted.

Grouped into one issue because each is a single small command on a service that needs nothing else, unlike the CloudFront and API Gateway deletions.

## Desired behaviour

- `KMS ScheduleKeyDeletion`, which is how a KMS key is deleted. Real KMS does not delete a key on request: it moves it to `PendingDeletion` with a waiting period of 7 to 30 days, and the key stops working for cryptographic operations at once. `CancelKeyDeletion` puts it back. The pending state is the interesting part to model, because a Stack deletion leaves the key behind in a way a test can observe.
- `KMS DeleteAlias`, which does delete on request, and leaves the key it pointed at alone.
- `Route53 DeleteHostedZone`, refusing a zone that still holds records other than its own NS and SOA, which is real Route53's `HostedZoneNotEmpty`.
- `ACM DeleteCertificate`, refusing a certificate that is still in use with `ResourceInUseException`.

## Acceptance criteria

- [ ] Each command is routed and authorized as its own IAM action.
- [ ] A key with deletion scheduled reports `PendingDeletion` and refuses encrypt and decrypt, rather than disappearing.
- [ ] `CancelKeyDeletion` returns a pending key to `Enabled`.
- [ ] A hosted zone with records left in it is refused, and one holding only its NS and SOA records is deleted.
- [ ] Deleting a hosted zone stops its records resolving through simulated DNS.
- [ ] Each answers the service's own not-found error for something that is not there.

## Out of scope

The CloudFormation teardown that would call these, and KMS key rotation.

PR #407, opened 2026-08-05
feat: delete simulated Buckets, functions, Roles

### LINKED ISSUE #402: Delete simulated S3 Buckets, Lambda functions and IAM Roles
## Problem

Most simulated services can create their main resource but never delete it. There is no `DeleteBucket`, no `DeleteFunction`, and nothing on IAM for removing a Role or a policy, so a test that wants to check clean-up behaviour cannot, and a long-running local development process accumulates resources it can never release.

This also blocks stack deletion: #397 needs a real delete operation to call for each resource type a template created.

## Desired behaviour

The AWS delete operations exist for S3 Buckets, Lambda functions, and IAM Roles and policies, and behave the way the real ones do, including the refusals. Real AWS will not delete a Bucket with Objects in it, or a Role that still has policies on it, and those refusals are exactly why a CloudFormation Stack deletion fails partway. Simulating the success and not the refusal would make a passing test mean less than it should.

## Proposed scope

- `S3 DeleteBucket`, refusing a Bucket that still holds Objects with `BucketNotEmpty`, and releasing the Bucket name globally so another Account or Region can claim it.
- `Lambda DeleteFunction`, taking the function's Function URL with it. Event source mappings are left alone, as real Lambda leaves a mapping pointing at a function that is gone.
- `IAM DeleteRole`, refusing with `DeleteConflict` while the Role carries inline or attached policies.
- `IAM DetachRolePolicy` and `IAM DeleteRolePolicy`, which are what a caller uses to get a Role into a deletable state.
- `IAM DeletePolicy`, refusing with `DeleteConflict` while any Role or User still carries the ARN.

Each is authorized as its own IAM action and routed for SDK client interception.

## Out of scope

The delete operations for the other simulated services, which are their own issues, and the CloudFormation teardown that calls them.

hughgrigg:
Adds the AWS delete operations for simulated S3 Buckets, Lambda functions, and IAM Roles and policies. These are the first of the delete commands stack deletion needs, and are useful on their own: until now most services could create their main resource and never remove it.

## What is here

| Command | Behaviour |
| --- | --- |
| `S3 DeleteBucket` | Refuses a Bucket that still holds Objects with `BucketNotEmpty`, and releases the Bucket name globally so another Account or Region can claim it |
| `Lambda DeleteFunction` | Takes the function's Function URL with it, and leaves event source mappings alone as real Lambda does |
| `IAM DeleteRole` | Refuses with `DeleteConflict` while the Role carries inline or attached policies |
| `IAM DetachRolePolicy` | Removes the attachment `AttachRolePolicy` recorded, leaving the managed policy itself |
| `IAM DeleteRolePolicy` | Removes the inline policy `PutRolePolicy` stored |
| `IAM DeletePolicy` | Refuses with `DeleteConflict` while any Role or User still carries the ARN |

Each is authorized as its own IAM action, routed for SDK client interception, and covered by colocated isolated tests including the routing path.

The refusals matter as much as the deletions. Real AWS will not delete a Bucket with Objects in it or a Role that still has policies on it, and those refusals are exactly why a CloudFormation Stack deletion fails partway. Simulating only the success would make a passing test mean less than it should.

## Refactors

Two collaborators were extracted because the service facades sat exactly on the 300 line limit:

- `SimS3BucketAccess` takes the Bucket lookup, endpoint URL and filesystem mount conveniences off `SimS3`. They are test and local development conveniences rather than S3 operations, so they read better together and away from the command delegations.
- IAM gained the `command/sim-iam-command.types.ts` barrel that S3 and Lambda already had, which replaced 60 lines of per-command imports in `SimIam` with one.

## Not here

Docs. They are held back until the rest of the delete commands land, so the S3, Lambda and IAM pages can be updated once rather than three times.

The delete operations for the other services are #403, #404 and #405. The CloudFormation side is #406, which unblocks #397 and then #398.

Resolves https://github.com/KensioSoftware/yulin/issues/402

### DIFF
--- a/src/service/iam/command/policy/delete-policy/delete-policy.command.ts
+++ b/src/service/iam/command/policy/delete-policy/delete-policy.command.ts
@@ -1,0 +1,9 @@
+export interface SimDeletePolicyCommandInput {
+  readonly PolicyArn?: string | undefined;
+}
+
+export interface SimDeletePolicyCommand {
+  readonly input: SimDeletePolicyCommandInput;
+}
+
+export type SimDeletePolicyCommandOutput = Record<string, never>;
--- a/src/service/iam/command/policy/delete-policy/delete-policy.handler.ts
+++ b/src/service/iam/command/policy/delete-policy/delete-policy.handler.ts
@@ -1,0 +1,99 @@
+import type { CommandHandler } from "../../../../../command/command-handler.js";
+import {
+  type BackgroundScheduler,
+  BackgroundTasks,
+} from "../../../../../util/background/background.js";
+import type { SimArn } from "../../../../aws/arn.js";
+import {
+  SimIamDeleteConflict,
+  SimIamNoSuchEntity,
+} from "../../../error/sim-iam.error.js";
+import type { SimIamManagedPolicy } from "../../../policy/sim-iam-policy.js";
+import type { SimIamRole, SimIamRoleName } from "../../../role/sim-iam-role.js";
+import type { SimIamUser, SimIamUsername } from "../../../user/sim-iam-user.js";
+import type {
+  SimDeletePolicyCommand,
+  SimDeletePolicyCommandOutput,
+} from "./delete-policy.command.js";
+
+interface DeletePolicyCommandHandlerProperties {
+  readonly policies: Map<SimArn, SimIamManagedPolicy>;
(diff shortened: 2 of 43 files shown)


hughgrigg:
@coderabbitai rate limit

PR #408, opened 2026-08-05
feat: create Resources without unsupported props

### LINKED ISSUE #391: Unsupported properties fail a simulated CloudFormation stack instead of deploying best effort
## Problem

Simulated CloudFormation deploys as much of a template as it can. A resource type it does not simulate is skipped, the stack carries on, and `stack.skippedResources` says what was missed. That gets a lot of real templates deployed. An unsupported *property* on a resource gets one of two harsher answers, and neither is best effort.

## Current behaviour

An `AWS::S3::Bucket` carrying `VersioningConfiguration` throws `Invalid AWS::S3::Bucket Resource ...` from `SimCfnS3BucketPropertyRules`. That fails the resource, and `SimCfnStackDeploymentLifecycle` moves the whole stack to `CREATE_FAILED`. Eighteen other real bucket properties do the same, as does any property name the rules do not recognise. Cognito is stricter still: `SimCfnCognitoPropertyParser` runs an allow-list, so any property off it, at the top level or nested, fails the same way.

DynamoDB, Lambda, SQS, SSM, Secrets Manager and API Gateway v2 word their property refusals as `Unsupported sim ... CloudFormation`, which `SimCfnResourceCreateOperation` downgrades to a skip. Better, but the whole table is then absent over one setting on one index.

Simulated S3 already does the right thing for two properties: `BucketEncryption` and `Tags` are read and ignored, because nothing this simulator models can tell an encrypted bucket from an unencrypted one. That case is currently a hardcoded exception rather than the default.

## Desired behaviour

Three answers to a property, with ignore as the default:

1. **Act on it**, where it is simulated.
2. **Ignore it and record it** against the resource. The resource is created, the deployment continues, and the ignored property is visible to whoever is reading the results.
3. **Refuse** only where there is no way to deploy anything coherent, such as a `BucketName` that is not a string.

Skipping a whole resource stays available for the case where the resource cannot meaningfully exist at all, like a Lambda function declaring a Python runtime.

Ignored properties need a reporting surface next to `stack.skippedResources`: `resource.ignoredProperties` and a stack-level roll-up, each naming the property path and why it was ignored. An unrecognised property name is ignored and recorded too, so a template using a property added to AWS after this simulator read the docs still deploys.

This trades safety for reach and should be stated plainly in the docs: an ignored property means the simulated resource behaves differently to AWS, and the recorded list is where a test author checks whether that matters to them.

## Example

```typescript
  stackName: "site",
  template: {
    Resources: {
      SiteBucket: {
        Type: "AWS::S3::Bucket",
        Properties: {
          BucketName: "site-bucket",
          VersioningConfiguration: { Status: "Enabled" },
        },
      },
    },
  },
});

await stack.waitForDeployComplete();

// The bucket exists and is usable, unversioned.
console.log(stack.getResource("SiteBucket")?.simResource !== undefined); // true

for (const ignored of stack.ignoredProperties) {
  console.log(ignored.logicalId, ignored.path, ignored.reason);
  // "SiteBucket VersioningConfiguration simulated S3 does not simulate object versions"
}
```

## Proposed scope

The mechanism and the policy in simulated CloudFormation core, plus simulated S3 converted as the reference implementation, since it already has the three-set structure to build on. Documentation of the rule in `docs/services/cloudformation/README.md` alongside the existing skipped-resource section.

## Acceptance criteria

- [ ] A resource can record an ignored property with its path and reason, and creation continues.
- [ ] `stack.ignoredProperties` lists every ignored property across the stack, and each resource exposes its own.
- [ ] An `AWS::S3::Bucket` with any of the currently refused properties deploys, is usable, and appears in `stack.ignoredProperties`.
- [ ] A property name simulated S3 does not recognise is ignored and recorded rather than failing the resource.
- [ ] A `BucketName` that is not a string still fails the resource, since there is no coherent fallback.
- [ ] The docs state the trade-off: an ignored property means simulated behaviour diverges from AWS, and say where to check.

## Out of scope

Reclassifying the other services, which is a follow-up issue. Actually simulating any of the properties that become ignored.

## Implementation notes

The skip path is triggered by a string match on the error message (`SimCfnResourceCreateOperation.isUnsupportedResourceError`). The ignore path should not be a second message-wording convention layered on top of that one: it needs to be something a property parser calls, not something it phrases. Whether the existing skip trigger is worth replacing at the same time is an open question.

### LINKED ISSUE #392: Stop skipping whole resources over one unsupported property in simulated DynamoDB, SQS and Cognito
## Problem

Simulated CloudFormation refuses or skips a resource over a single property it does not simulate, so a template that is 95% deployable produces a stack missing tables, queues and user pools. Once the ignore-and-record mechanism from #391 exists, each service's property rules need reclassifying against it so outright refusal becomes rare.

## Desired behaviour

Each service that refuses CloudFormation properties today sorts them into: simulated, ignored and recorded, or refused. Refused is for the cases where there is no way to deploy anything coherent, and should be a short list per service.

Services to work through, with the file that decides today:

- `AWS::Cognito::UserPool` and friends, `sim-cfn-cognito-property-parser.ts`, currently a strict allow-list including nested keys.
- `AWS::DynamoDB::Table` and `AWS::DynamoDB::GlobalTable`, `sim-cfn-dynamodb-property-error.ts` and `sim-cfn-dynamodb-global-table-replicas.ts`.
- `AWS::SQS::Queue`, `AWS::SSM::Parameter`, `AWS::SecretsManager::Secret`, `AWS::ApiGatewayV2::*`, `AWS::KMS::Key`.

The Lambda runtime skip is a genuine whole-resource case and should stay a skip: a Python function cannot run in a Node.js vm whatever is done with its other properties.

## Proposed scope

Per-service reclassification only, using the mechanism and policy landed in #391. Each service keeps its own list, since only the service knows whether a property's absence is observable through its simulated commands.

## Acceptance criteria

- [ ] Each listed service records ignored properties instead of skipping the resource, except where the resource cannot exist at all.
- [ ] A DynamoDB table setting an unsimulated property on one index is created, with the property recorded.
- [ ] A Cognito user pool carrying a property off the current allow-list is created, with the property recorded.
- [ ] Existing tests asserting on `skippedReason` for property-level cases are updated to assert on the recorded ignore instead.
- [ ] Each service README states which properties it ignores and what that means for the simulated behaviour.

## Out of scope

Simulating any of the properties that become ignored. Changing which resource *types* are simulated.

Depends on #391.

hughgrigg:
Simulated CloudFormation already skipped a Resource type it could not create and carried on with the rest of the stack. A *property* it could not act on got one of two harsher answers: a hard failure that took the whole stack to `CREATE_FAILED` (S3, Cognito), or a skip that removed the Resource entirely over one setting (DynamoDB, SQS, API Gateway v2). This applies the same best-effort answer one level down.

Resolves #391
Resolves #392

## The mechanism

A service property parser calls `resource.ignoreProperty(path, reason)` instead of throwing. The Resource is created without the property, and the omission is on `resource.ignoredProperties` and on `stack.ignoredProperties`:

```typescript
  stackName: "uploads-stack",
  template: {
    Resources: {
      UploadsBucket: {
        Type: "AWS::S3::Bucket",
        Properties: {
          BucketName: "uploads",
          VersioningConfiguration: { Status: "Enabled" },
        },
      },
    },
  },
});
await stack.waitForDeployComplete();

// The Bucket exists and is usable, unversioned.
for (const ignored of stack.ignoredProperties) {
  console.log(ignored.logicalId, ignored.path, ignored.reason);
}
```

The path is the whole way down, so a setting on one entry of a list says which entry it was on, such as `GlobalSecondaryIndexes.1.WarmThroughput`. The stack getter reads from the Resources rather than collecting during deployment, so the two can never disagree.

This trades safety for reach, and the docs say so plainly: an ignored property means the simulated Resource behaves differently to the one the template describes, and the record is where a test checks whether that matters.

## Services

S3, DynamoDB (table and global table, at all six levels a global table nests), Cognito, API Gateway v2, SQS and KMS now record rather than refuse or skip. A property name AWS has no such thing as is recorded too, rather than a stack failing over a typo or a property AWS added since the list was written.

Properties nothing simulated could tell apart stay unrecorded. `BucketEncryption` and `Tags` on an S3 Bucket are on almost every Bucket CDK synthesizes, and listing a difference no test could observe would only bury the ones that matter.

## What is still refused

Three cases, each because the code disagreed with the plan in the issues:

- **A template that leaves nothing coherent to create.** A `BucketName` that is not a string, or a global table whose replica list omits the region the stack is deploying into. Real CloudFormation refuses these too.
- **A value the simulated service itself refuses an SDK caller.** `AWS::SQS::Queue` with `FifoQueue: true` is the one: a FIFO queue is named `<name>.fifo`, which simulated SQS refuses, so there is no queue to create under the name the template gave it. Issue #392 assumed this could be recorded; it cannot.
- **A Resource that cannot exist at all**, such as a Lambda function declaring a Python runtime. These stay whole-Resource skips.

## Two fallbacks

- A `AWS::DynamoDB::GlobalTable` naming several replicas is now created as an ordinary table in the stack's own region, with `Replicas` recorded, rather than skipped.
- Autoscaled capacity needed a fallback rather than only a record. Once `WriteCapacityAutoScalingSettings` was ignored, a `PROVISIONED` table had no capacity at all and `CreateTable` refused it. It now falls back to the `MinCapacity` those settings name, which is where autoscaling starts the table on AWS.

## Notes for review

- One API Gateway test was only passing because the allow-list refused it before `CreateAuthorizer` saw it; its authorizer template was incomplete and now needs a valid one.
- An LSI carrying `ProvisionedThroughput` still fails the stack. The index entry is handed to `CreateTable` as the template wrote it, and real DynamoDB refuses it, so that is the AWS-matching outcome rather than a gap.
- `docs/services/cloudformation/README.md` gains a section on the new behaviour, and the S3, DynamoDB, SQS, KMS, Cognito and API Gateway READMEs are updated where they described the old refusals.

`pnpm check` passes: 753 test files, 4686 tests, 99.9% statements.

## Summary by CodeRabbit

* **New Features**
  * CloudFormation stacks can now deploy resources when they contain unsupported or unknown properties.
  * Omitted properties are reported through `stack.ignoredProperties`, including resource, property path, and reason.
  * API Gateway, Cognito, DynamoDB, KMS, S3, and SQS simulations now support this behavior.
  * DynamoDB global tables can deploy locally when replication is unsupported; replication settings are recorded as ignored.
* **Bug Fixes**
  * Unsupported properties no longer incorrectly cause resources to be skipped or deployments to fail.
  * Existing validation errors remain for invalid resource configurations and unsupported resource types.

### DIFF
--- a/docs/services/apigatewayv2/README.md
+++ b/docs/services/apigatewayv2/README.md
@@ -1890,10 +1890,10 @@ Current documented limitations:
 - An authorizer function that throws is a 500 rather than a 401. Real Lambda turns a thrown error
   into a payload carrying `errorMessage`, and simulated Lambda rejects with the error itself, so
   returning `{ "errorMessage": "Unauthorized" }` is how an authorizer asks for a 401 here.
-- `AuthorizerCredentialsArn` is refused, as is `AuthorizerCredentialsArn` on an
-  `AWS::ApiGatewayV2::Authorizer` and `authorizerCredentials` in an imported document. It names a
-  Role API Gateway assumes to invoke the authorizer, and the function's own resource policy is the
-  whole decision here.
+- `AuthorizerCredentialsArn` is refused on `CreateAuthorizer`, as is `authorizerCredentials` in an
+  imported document. On an `AWS::ApiGatewayV2::Authorizer` it is recorded instead, and the authorizer
+  is created without it. It names a Role API Gateway assumes to invoke the authorizer, and the
+  function's own resource policy is the whole decision here.
 - `AuthorizerResultTtlInSeconds` is accepted between 0 and 3600, which is the range AWS accepts, and
   is refused on an authorizer with no `IdentitySource`, since there would be nothing to key the held
   decision on.
@@ -1987,8 +1987,9 @@ Current documented limitations:
   only the name `GetApi` reports.
 - A terminal `{proxy+}` in an imported path reaches `CreateRoute` unchanged. Greedy segments are
   established for route keys rather than for OpenAPI path templating.
-- `AWS::ApiGatewayV2::Api` refuses `BodyS3Location` by name. Reading a document out of a simulated S3
-  bucket adds a fetch path and nothing about OpenAPI.
+- `AWS::ApiGatewayV2::Api` records `BodyS3Location` rather than reading it, so the API is created
+  with no routes at all. Reading a document out of a simulated S3 bucket adds a fetch path and
+  nothing about OpenAPI.
 - `AWS::ApiGatewayV2::Api` refuses `Policy` with a message of its own saying an HTTP API has no
   resource policy. There is no such property on the real Resource type, so a template carrying one
   was written for a REST API rather than hitting a gap here.
--- a/docs/services/apigatewayv2/README.md
+++ b/docs/services/apigatewayv2/README.md
@@ -1675,8 +1675,11 @@ console.log(await response.text());
 srv.close();
 ```
 
-Every property outside the simulated set is refused by name, and the refusal fails the stack. The
-simulated properties are:
+Every property outside the simulated set is left out of what is created and recorded in
+[`stack.ignoredProperties`](../cloudformation/README.md#properties-a-resource-was-created-without),
+naming the Resource type, the logical id and the ones this can act on instead. The API, authorizer,
+integration, route or stage is created either way, so the stack deploys and the record says which of
+its parts behaves differently to the template. The simulated properties are:
 
 - `Api`: `Name`, `ProtocolType`, `Description`, `DisableExecuteApiEndpoint`, `Body`,
   `FailOnWarnings`
@@ -1756,6 +1759,7 @@ Api:
 `ProtocolType` that is present has to be `HTTP`, and a `Name` that is present names the API instead
 of the document's `info.title`. `Description` and `DisableExecuteApiEndpoint` are refused alongside a
 `Body`, because `ImportApi` does not take them and nothing here changes an API after it is created.
+Each is recorded rather than applied, so the API is created from the document without them.
 
 A template combining an `Api` with a `Body` and a separate `Route`, `Integration` or `Authorizer`
 Resource for that same API fails the stack, naming both logical IDs. The document already declares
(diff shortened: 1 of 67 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #411, opened 2026-08-05
feat: delete and update Distributions, delete CFFs

### LINKED ISSUE #404: Delete a simulated CloudFront distribution and CloudFront Function

hughgrigg:
Adds CloudFront `DeleteDistribution`, `UpdateDistribution` and `DeleteFunction`, the third of the delete-operation PRs a simulated Stack teardown needs. Resolves #404.

## Acceptance criteria

- [x] `DeleteDistributionCommand` and `DeleteFunctionCommand` are routed and authorized as `cloudfront:DeleteDistribution` and `cloudfront:DeleteFunction`. Both have a dedicated authorizer running before the state map is read, so an unauthorized caller cannot learn whether the resource exists.
- [x] A deleted Distribution stops serving. Deletion deregisters it from `SimCloudFrontRegistry`, which is what request routing reads, and the alternate domain names it answered on go with it. Covered by tests that serve a request to the CloudFront domain and to an alternate domain before and after the deletion.
- [x] Deleting something that is not there answers `NoSuchDistribution` and `NoSuchFunctionExists`.
- [x] A Distribution carries Enabled state, `UpdateDistribution` can change it, and deleting an enabled Distribution answers `DistributionNotDisabled`. The rule lives on the resource model as `assertDeletable`, alongside `SimRoute53HostedZone`.
- [x] A disabled Distribution deletes.
- [x] The unchecked `IfMatch` is documented as a divergence, in a new Limitations section on the CloudFront docs page.

## UpdateDistribution

The issue asks for enough of `UpdateDistribution` to flip Enabled. It is implemented as a real update rather than an Enabled-only special case, because CloudFront takes a whole `DistributionConfig` rather than a patch: `replaceConfiguration` on the Distribution drops everything derived from the previous config, and the same configurators creation uses then apply the new one. Anything left out of the new config is therefore dropped, which is what AWS does. The alternate domain names are resynchronized in the registry around that, so a name an update drops is free for another Distribution.

## Divergences

All four are recorded in the new Limitations section of `docs/services/cloudfront/README.md`:

- `IfMatch` is accepted and not checked on all three commands, so neither `PreconditionFailed` nor `InvalidIfMatchVersion` is ever returned. This is the decision recorded on the issue.
- A deletion does not wait for the disable to deploy. Real CloudFront needs the disabled Distribution to reach `Deployed` first; here `Enabled: false` is enough.
- A disabled Distribution still serves requests. Real CloudFront answers a disabled Distribution with a 403. Only deleting one stops it serving here.
- `DeleteFunction` never answers `FunctionInUse`, because nothing tells a CloudFront Function that a cache Behavior has taken it up.

## Also in here

- `SimCloudFront` gained the command parts class S3, Lambda, IAM and KMS already have. The facade was at 224 lines and three more commands would not fit under the 300 line limit.
- `simCloudFrontDistributionView` replaces the Distribution output shape that create, get and update would otherwise each build.
- The update handler tripped the FTA threshold at 58, so the reconfiguration moved onto `SimCloudFrontDistributionReconfigurer` and the configurator wiring onto a shared factory.
- The SDK router test's "rejects a Command simulated CloudFront does not support" case used `DeleteDistributionCommand`, which is now supported, so it uses `CreateInvalidationCommand` instead.

One thing left alone: `GetDistribution` still answers a missing Distribution with `ResourceNotFoundException` rather than `NoSuchDistribution`, which is what real CloudFront returns. Changing it is out of scope here but worth a follow-up.

`pnpm run check` passes.

## Summary by CodeRabbit

* **New Features**
  * Added support for updating and deleting CloudFront distributions.
  * Added support for deleting CloudFront functions.
  * Enforced disabling distributions before deletion.
  * Distribution updates now replace the complete configuration and refresh alternate-domain mappings.
  * Added IAM authorization for update and delete operations.

* **Documentation**
  * Expanded CloudFront documentation with usage examples, deletion guidance, and service limitations.

* **Bug Fixes**
  * Improved cleanup of distribution and alternate-domain registrations after deletion or reconfiguration.

### DIFF
--- a/docs/services/cloudfront/README.md
+++ b/docs/services/cloudfront/README.md
@@ -496,6 +496,107 @@ CDK app works without changes.
 A Distribution using `CloudFrontDefaultCertificate` needs no ACM certificate and is not checked. A
 standalone `new SimCloudFront()` has no sim ACM to check against, so it does not check either.
 
+## Disabling and deleting a Distribution
+
+`DeleteDistributionCommand` removes a Distribution. CloudFront will not delete one that is still
+serving, so the sequence is `UpdateDistributionCommand` with `Enabled: false` first, then the
+deletion. Deleting an enabled Distribution answers `DistributionNotDisabled`, as it does in AWS.
+
+`UpdateDistributionCommand` takes a whole `DistributionConfig` rather than a patch, so the update is
+applied as a replacement. Anything left out of the new config is dropped, including alternate domain
+names and the default root object. Read the Distribution first, change the field you want, and send
+the config back.
+
+Once the Distribution is deleted, a request to its CloudFront domain or any of its alternate domain
+names no longer resolves to it, and those alternate domain names are free for another Distribution.
+
+```typescript sim-cloudfront-delete-distribution
+/**
+ * Disabling a simulated CloudFront Distribution and then deleting it.
+ */
+
+import {
+  CreateDistributionCommand,
+  DeleteDistributionCommand,
+  type DistributionConfig,
+  GetDistributionCommand,
+  UpdateDistributionCommand,
+} from "@aws-sdk/client-cloudfront";
+import { CreateBucketCommand } from "@aws-sdk/client-s3";
+
+import { SimAws } from "@kensio/yulin";
+
+const simAws = new SimAws();
+const simCloudFront = simAws.cloudFront();
+
+await simAws
+  .s3()
+  .createBucket(new CreateBucketCommand({ Bucket: "site-bucket" }));
+
+const distributionConfig: DistributionConfig = {
+  CallerReference: "site-distribution",
+  Comment: "Site distribution",
+  Enabled: true,
+  Origins: {
+    Quantity: 1,
+    Items: [
+      {
+        Id: "site-origin",
+        DomainName: "site-bucket.s3.amazonaws.com",
+        S3OriginConfig: { OriginAccessIdentity: "" },
+      },
+    ],
+  },
+  DefaultCacheBehavior: {
+    TargetOriginId: "site-origin",
+    ViewerProtocolPolicy: "allow-all",
+  },
+};
+
+const created = await simCloudFront.createDistribution(
+  new CreateDistributionCommand({ DistributionConfig: distributionConfig }),
+);
+await simAws.backgroundTasksComplete();
+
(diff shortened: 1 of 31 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

Issue #397, opened 2026-08-05
Delete a simulated CloudFormation stack and the resources it created


hughgrigg:
## Problem

Simulated CloudFormation can create a stack but never delete one, so a stack is permanent for the lifetime of the process. In a test that is fine, but during local development against a long-running simulated AWS it means any change to the infrastructure is a full process restart. There is no way to tear a stack down and put it back.

## Current behaviour

`SimCloudFormationSdkCommandRouter` routes only `CreateStackCommand` and `DescribeStacksCommand`; sending a `DeleteStackCommand` through an intercepted SDK client throws an error naming the two supported commands. `SimCloudFormationStackStatus` covers only `REVIEW_IN_PROGRESS`, `CREATE_IN_PROGRESS`, `CREATE_COMPLETE` and `CREATE_FAILED`. Nothing removes an entry from the stack map, and `CreateStack` rejects a name already in use with `AlreadyExistsException`, so the same stack name cannot be deployed twice in one process.

## Desired behaviour

`DeleteStack` moves the stack to `DELETE_IN_PROGRESS`, deletes its resources in reverse dependency order, and finishes at `DELETE_COMPLETE`. The stack name becomes free, so a following `CreateStack` with the same name succeeds. `DescribeStacks` for a deleted stack name fails the way real CloudFormation does, with a `ValidationError` saying the stack does not exist.

## Proposed scope

The blocker is that most simulated services cannot delete their own primary resource. The only delete operations that exist today are `DeleteObject`, `DeleteObjects`, `DeleteBucketPolicy`, `DeletePublicAccessBlock`, `DeleteFunctionUrlConfig` and `RemovePermission`. There is no `DeleteBucket`, `DeleteTable`, `DeleteFunction`, `DeleteSecret`, `DeleteKey`, `DeleteQueue`, `DeleteParameter` or `DeleteHostedZone`.

So this issue has to add a teardown path per service alongside the existing creation factories. Whether that path is an SDK-shaped delete command on each service or an internal teardown hook on the resource factory is an open decision, and worth settling first: the SDK commands are useful in their own right, but they are a much larger piece of work than stack deletion needs.

## Acceptance criteria

- [ ] `DeleteStackCommand` is routed and authorized as `cloudformation:DeleteStack` against the stack ARN.
- [ ] `SimCloudFormationStackStatus` gains the delete statuses, and `DescribeStacks` reports them.
- [ ] Resources are deleted in reverse dependency order, mirroring the creation ordering.
[...]

PR #413, opened 2026-08-05
feat: delete Stack Resources through their services

hughgrigg:
The CloudFormation side of Resource deletion, which unblocks #397.

## Acceptance criteria

- **`SimCfnServiceResourceFactory` gains `delete`, implemented by every service for the Resource types it can create.** Met. ACM, API Gateway v2, CloudFormation, CloudFront, Cognito, DynamoDB, IAM, KMS, Lambda, Route53, S3, Secrets Manager, SQS and SSM, plus the two CDK custom Resources. Each calls its own SDK-shaped command; the engine never reaches into a service's state.
- **A delete lifecycle on `SimCfnResource` mirroring the create one.** Met. `SimCloudFormationResourceStatus` gains `DELETE_IN_PROGRESS`, `DELETE_COMPLETE` and `DELETE_FAILED`, held in a `SimCfnResourceDeletionState` beside the creation state, driven by `SimCfnResourceDeleteOperation` and `SimCfnResourceDeleter`.
- **Deletion in reverse dependency order.** Met. `simCfnResourceDependents` inverts the graph `dependencies()` already exposes, and `SimCfnStackResourceDeleter` / `SimCfnStackPendingDeletions` / `SimCfnStackResourceBatchDeleter` mirror the creation trio batch for batch.
- **A Resource type with no delete path is recorded rather than failing.** Met. A service raises the same unsupported-Resource error creation raises; the operation marks the deletion skipped (`DELETE_COMPLETE` with a reason) and the Stack reports it on `stack.skippedResourceDeletions`, read off the Resources the way `ignoredProperties` is.
- **Stack-level statuses and `DeleteStack` left to #397.** Met. `SimCfnStack.teardown()` is the internal entry point this PR tests through; nothing stack-level was added.

## Per-Resource behaviour

Where calling the delete command blindly would give the wrong answer, the deleter does what real CloudFormation does:

- **CloudFront Distribution** is disabled with `UpdateDistribution` before `DeleteDistribution`, which refuses an enabled one.
- **IAM Role** has its attached and inline policies removed before `DeleteRole`.
- **SQS QueuePolicy** is removed by `SetQueueAttributes` with an empty `Policy`, since there is no `DeleteQueuePolicy`.
- **Route53 RecordSet** is removed by a `DELETE` change.
- **S3 Bucket** holding Objects fails the teardown with `BucketNotEmpty`, deliberately not emptied first.
- **KMS Key** is scheduled for deletion and left in `PendingDeletion`.
- **WaitConditionHandle** and `Custom::CDKBucketDeployment` have nothing to delete. `Custom::S3BucketNotifications` puts an empty configuration back, as the CDK provider does on its Delete event.

## Divergences worth flagging

- **`SimCfnResource` was split.** It was at the 300-line limit, so what a Resource *is* moved to a new `SimCfnResourceRecord` base class and what has *happened* to it stayed on `SimCfnResource`. No public member moved or changed name.
- **Sim SQS now accepts an empty `Policy` attribute as a removal.** `SetQueueAttributes` rejected `Policy: ""`, so the queue policy had no removal path at all. Real SQS treats an empty value as clearing the attribute, which is the only way back to a queue without a policy. One line in `SimSqsQueueJsonAttribute`, with an SDK-level test beside the other attribute commands.
- **User-facing docs are unchanged.** `docs/services/cloudformation/README.md` still says stack deletes are not supported, which stays true until #397 puts a command on top of this. The implementation README carries the design.
- **The PR is larger than the usual steer**, at roughly 3,900 lines. It is 14 services plus the engine, and a little over half is tests.

## Testing

`pnpm run check` passes. Each service has a teardown test that deploys a template and asserts through that service's own API that the resources are gone, including the CloudFront disable-first and S3 `BucketNotEmpty` cases, which a naive implementation gets wrong while looking right.

## Summary by CodeRabbit

- **New Features**
  - Added CloudFormation stack teardown with reverse dependency-order deletion.
  - Added deletion lifecycle statuses, skip reporting, failure handling, and cyclic-dependency detection.
  - Added teardown support across ACM, API Gateway, CloudFront, Cognito, DynamoDB, IAM, KMS, Lambda, Route 53, S3, Secrets Manager, SQS, SSM, and custom resources.
  - Added cleanup for policies, notifications, integrations, routes, stages, and related dependencies.

- **Tests**
  - Added comprehensive teardown integration tests covering ordering, skipped resources, failures, and service-specific cleanup.

- **Documentation**
  - Expanded CloudFormation lifecycle and teardown documentation.

### DIFF
--- a/src/service/acm/cfn/sim-cfn-acm-resource-factory.ts
+++ b/src/service/acm/cfn/sim-cfn-acm-resource-factory.ts
@@ -5,6 +5,8 @@ import type {
 import type { SimCfnServiceResourceFactory } from "../../cloudformation/resource/factory/sim-cfn-resource-factory.type.js";
 import { SimAcm } from "../sim-acm.js";
 import { SimCfnAcmCertificateCreator } from "./certificate/sim-cfn-acm-cert-creator.js";
+import type { SimAcmCertificate } from "../certificate/sim-acm-certificate.js";
+import { assertDefined } from "../../../util/type-guard/defined.js";
 
 interface SimAcmCfnResourceFactoryProperties {
   readonly acm?: SimAcm | undefined;
@@ -14,11 +16,13 @@ interface SimAcmCfnResourceFactoryProperties {
  * CloudFormation Resource factory for simulated ACM resources.
  */
 export class SimAcmCfnResourceFactory implements SimCfnServiceResourceFactory {
+  private readonly acm: SimAcm;
   private readonly certificateCreator: SimCfnAcmCertificateCreator;
 
   constructor(properties: SimAcmCfnResourceFactoryProperties = {}) {
     const { acm = new SimAcm() } = properties;
 
+    this.acm = acm;
     this.certificateCreator = new SimCfnAcmCertificateCreator({ acm });
   }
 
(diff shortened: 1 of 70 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #414, opened 2026-08-05
feat: delete a simulated CloudFormation Stack

hughgrigg:
Adds the Stack-level delete lifecycle on top of the Resource teardown #413 landed.

`DeleteStackCommand` moves the Stack to `DELETE_IN_PROGRESS`, runs the reverse-dependency-ordered teardown in the background, and releases the Stack name when it completes. That name release is the point of the issue: a long-running local process can replace a Stack rather than restart.

## Acceptance criteria

- [x] `DeleteStackCommand` is routed and authorized as `cloudformation:DeleteStack` against the Stack ARN.
- [x] `SimCloudFormationStackStatus` gains `DELETE_IN_PROGRESS`, `DELETE_COMPLETE` and `DELETE_FAILED`, and `DescribeStacks` reports them.
- [x] Resources are deleted in reverse dependency order. This is `SimCfnStack.teardown()` from #413, unchanged.
- [x] Every Resource type a Stack can create can also be deleted, or is recorded as skipped. Also #413.
- [x] After deletion, `CreateStack` with the same Stack name succeeds instead of throwing `AlreadyExistsException`.
- [x] A test deploys a Stack, deletes it, and asserts the underlying resources are gone through their own service APIs (S3, Lambda and IAM in `delete-stack.iso.test.ts`, plus an intercepted SDK client in the router test).

## Design

`SimCfnStackDeletionLifecycle` is separate from `SimCfnStackDeploymentLifecycle` for the same reason `SimCfnResourceDeletionState` is separate from `SimCfnResourceCreationState`: a Stack that was never asked to delete has no deletion status at all. `SimCfnStack.status` reads the deletion status first and falls back to the deployment status, so it says the last thing CloudFormation did to the Stack. The scheduler both lifecycles share moves up from `stack/deploy/` to `stack/` as `SimCfnStackOperationScheduler`.

Releasing the name belongs to whoever holds the Stack map, so the lifecycle only says when the name is free and `DeleteStackCommandHandler` frees it.

## Divergences and behaviour changes

- **`DescribeStacks` now refuses an unknown Stack name.** It threw nothing before and returned an empty `Stacks` list. Real CloudFormation returns a `ValidationError`, and that is also how it answers a deleted Stack name, so the two cases are the same code path here. The existing test asserting the empty list is changed rather than kept.
- **A deleted Stack cannot be described at all.** Real CloudFormation keeps a deleted Stack readable by its unique Stack ID; the simulator identifies a Stack by name alone. Recorded under Limitations.
- **`DeletionPolicy: Retain` is now honoured** rather than ignored. The Resource is left in simulated AWS and reported as `DELETE_SKIPPED`, which is the status CloudFormation reports for one it stepped over, and the Stack still deletes around it. `RetainExceptOnCreate` is treated the same way, since the two differ only in what a rolled back creation does. `Snapshot` is treated as `Delete`, because no simulated service takes snapshots; that one is recorded under Limitations. `Retain` is common in CDK output, so deleting those resources would have been a quiet wrong answer rather than a missing feature.
- **A failed teardown does not report success.** An S3 Bucket still holding Objects fails, as it does in AWS. The Stack stays in `DELETE_FAILED` with the error on it and the name in use, `waitForStackDeleteComplete(...)` rethrows, and `DescribeStacks` reports it as `StackStatusReason`. Asking again after the cause is dealt with deletes the Stack. Covered in `delete-stack-failures.iso.test.ts`.
- **Deleting a Stack name that is not there succeeds.** Checked against the `DeleteStack` API reference, which lists only `TokenAlreadyExists` as an error.
- `DeleteStack` reads only `StackName`. `RetainResources`, `DeletionMode`, `RoleARN` and `ClientRequestToken` are not read, so a Stack in `DELETE_FAILED` cannot be forced through the way `FORCE_DELETE_STACK` forces it. Recorded under Limitations.

## Other changes

- `error/sim-cloudfront.error.ts` is renamed to `error/sim-cloudformation.error.ts`. It holds the CloudFormation errors, and now holds `ValidationError` as well.
- The SDK router test used `DeleteStackCommand` as its example of an unsupported Command. It now uses `ListStacksCommand`, and there is a test that deletes a Stack through an intercepted client.
- `SimCfnDescribedStacks` is extracted from `DescribeStacksCommandHandler`, which was one line short of the FTA threshold before this change.

`pnpm run check` passes.

## Summary by CodeRabbit

* **New Features**
  * Added simulated CloudFormation stack deletion with asynchronous completion tracking.
  * Added support for deletion statuses, failure handling, retries, and stack-name reuse.
  * Added `DeletionPolicy` retention behavior and reporting for retained resources.
  * Added delete-stack SDK routing and completion waiting.
* **Bug Fixes**
  * Corrected CloudFormation error references and improved missing-stack validation responses.
* **Documentation**
  * Expanded CloudFormation deletion guidance, limitations, and executable examples.


hughgrigg:
@coderabbitai full review

PR #415, opened 2026-08-05
feat: update a simulated CloudFormation Stack

### LINKED ISSUE #398: Update a simulated CloudFormation stack from a changed template
## Problem

Simulated CloudFormation has no `UpdateStack`, so a template change cannot be applied to a stack that already exists. Every deployment is a create. During local development that means restarting the process to pick up an edit to the infrastructure code, and losing everything in the simulated resources along with it.

The assumption is already baked into service code rather than being merely absent: `SimCfnSsmParameterCreator` documents that "Sim CloudFormation has no UpdateStack, so every deployment is a create", and writes parameters as creates that fail on a name already in use for exactly that reason.

## Desired behaviour

`UpdateStack` with a new `TemplateBody` for an existing stack compares the new template against the deployed one and reconciles the difference: resources only in the new template are created, resources only in the old one are deleted, and resources whose resolved properties changed are updated. The stack moves through `UPDATE_IN_PROGRESS` to `UPDATE_COMPLETE`, and `DescribeStacks` reports the new template body and re-resolved outputs.

An update with no differences fails the way real CloudFormation does, with a `ValidationError` reading `No updates are to be performed.`

## Example

```ts
await cfn.send(new CreateStackCommand({ StackName: "app", TemplateBody: v1 }));
await cfn.send(new UpdateStackCommand({ StackName: "app", TemplateBody: v2 }));
// A bucket added in v2 exists; a queue dropped from v2 is gone.
```

## Proposed scope

Full-template updates only, applied directly. No change sets, no rollback on a failed update, no drift detection.

For a changed resource, the simple first cut is replacement: delete it and create it again from the new properties. That is correct for immutable properties but diverges from real CloudFormation, which updates most properties in place and keeps the resource's data. It is a deliberate simplification worth stating in the docs, and in-place update is the obvious follow-up. Yulin has no users, so changing this later is fine.

## Acceptance criteria

- [ ] `UpdateStackCommand` is routed and authorized as `cloudformation:UpdateStack` against the stack ARN.
- [ ] `SimCloudFormationStackStatus` gains the update statuses, and `DescribeStacks` reports them.
- [ ] Added, removed and changed resources are each reconciled, in dependency order.
- [ ] Stack outputs are re-resolved against the new template.
- [ ] An update with an identical template throws `ValidationError` with `No updates are to be performed.`
- [ ] `UpdateStack` for a stack name that does not exist throws the same not-found `ValidationError` as `DescribeStacks`.

## Dependencies

Depends on #397, and through it on the resource teardown in #406 and the service delete commands in #402, #403, #404 and #405. Removing a resource and replacing a changed one both need that teardown path, so this is the last piece of the chain to land.

## Out of scope

- `CreateChangeSet` and `ExecuteChangeSet`, which are a distinct usage mode.
- Rollback to the previous template when an update fails partway.
- In-place property mutation where AWS would not replace the resource, per the note above.
- A file watch mode that calls `UpdateStack` on template changes.

hughgrigg:
Adds `UpdateStackCommand` to simulated CloudFormation, so a template change can be applied to a stack that is already deployed rather than needing the process restarted and the simulated state thrown away. This is the last piece of the chain that started from that report: #402, #403, #410 and #411 added the service deletes, #406 the resource teardown, and #397 `DeleteStack`.

## Acceptance criteria

- [x] `UpdateStackCommand` is routed and authorized as `cloudformation:UpdateStack` against the stack ARN.
- [x] `SimCloudFormationStackStatus` gains `UPDATE_IN_PROGRESS`, `UPDATE_COMPLETE` and `UPDATE_FAILED`, and `DescribeStacks` reports them, with the reason beside `UPDATE_FAILED` the way it is beside `DELETE_FAILED`.
- [x] Added, removed and changed resources are each reconciled, in dependency order: removals in reverse dependency order first, then additions in forward order.
- [x] Stack outputs are re-resolved against the new template, which `DescribeStacks` reports along with the new template body.
- [x] An update with an identical template throws `ValidationError` with `No updates are to be performed.`
- [x] `UpdateStack` for a stack name the service does not hold throws the same not-found `ValidationError` as `DescribeStacks`.

## The engine change behind it

Create and delete were whole-stack operations, so `SimCfnStackResourceCreator.createAll()` and `SimCfnStackResourceDeleter.deleteAll()` worked over the whole resource map. Both now take the resources to work on, and read dependencies across every resource in the stack rather than only those, so a resource an update leaves alone still counts as a satisfied dependency and still holds up the resource it names. `createAll()` and `deleteAll()` remain as the whole-map calls.

`SimCfnStackResourceOperations` now holds the simulated AWS scope a stack creates and deletes in, because deploying, updating and tearing down all need the same one. `SimCfnStack` gained an update lifecycle beside the other two, and lost the same amount of code again to `SimCfnStackOperationStatus` (which of the three operations the status and error come from) and `SimCfnStackResourceReport` (skipped, retained and ignored, all read off the resources). `stack.skippedResources` is now read off the resources like the others rather than collected during deployment, so it reports the resources the stack holds after an update.

## Divergences

Three, all documented under [changed resources are replaced](https://github.com/KensioSoftware/yulin/blob/feat/hg/update-stack/docs/services/cloudformation/README.md#changed-resources-are-replaced):

- **Replacement rather than in-place update.** A resource whose resolved template entry changed is deleted and created again. Real CloudFormation updates most properties in place and keeps the resource's data, so a bucket that gains a property loses its objects here. The issue records this as a deliberate first cut; in-place update is the follow-up.
- **Replacement cascades to dependents.** A resource naming a replaced resource is replaced too, all the way up the chain. Real CloudFormation hands the dependent the new physical name instead. Without this the dependent would be left pointing at a resource that has gone, and the deletion could not proceed past it.
- **`UpdateReplacePolicy` is not read.** Honouring `Retain` would leave the old resource holding the name its replacement needs, and CDK marks buckets and tables `Retain` as a matter of course, so honouring it would fail every such update. The old resource is deleted whatever the policy says.

Resources are compared as they resolve rather than as they are written, so a changed parameter value is a changed resource and a reordered template is not a change at all. Outputs are compared the same way, so an update that only changes an output is an update rather than a refusal.

Out of scope as the issue says: change sets, rollback of a failed update, and drift detection. A failed update leaves the stack holding whatever it managed.

## Docs

Extends the existing CloudFormation page with an "Updating a stack" section and a runnable example, and updates its functionality and limitations lists. The read-through turned up four stale claims elsewhere that the delete work and this one made wrong: API Gateway v2, DynamoDB, KMS and SSM each said stack updates or deletes were not simulated. Those are corrected, including a KMS limitation still saying `DeleteAlias` is not supported.

`pnpm run check` passes.

Resolves https://github.com/KensioSoftware/yulin/issues/398

## Summary by CodeRabbit

* **New Features**
  * Added support for updating CloudFormation stacks, including resource creation, deletion, replacement, parameter changes, output-only changes, and asynchronous status tracking.
  * Added update failure handling, validation, authorization checks, rollback behavior, and completion waiting.
  * Preserved unchanged resources and dependencies during updates.

* **Documentation**
  * Expanded CloudFormation update guidance and examples.
  * Clarified replacement behavior and limitations for API Gateway, DynamoDB, KMS, and SSM resources.

* **Tests**
  * Added comprehensive coverage for successful updates, failures, replacements, deletions, no-op changes, and SDK integration.


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #421, opened 2026-08-05
feat: reload the browser when serving on localhost

### LINKED ISSUE #417: Reload the browser automatically when serving simulated AWS on localhost
## Problem

When a simulated environment serves HTML to a browser, a CloudFront distribution over a filesystem-backed S3 bucket, a Lambda Function URL, or an HTTP API, every change means a manual refresh. Yulin is the only thing in the response path, so it is the only thing that can tell the browser to reload.

## Desired behaviour

Opt in with `serveSimAws({ port: 8787, liveReload: true })`. The local server then serves a reload channel on a reserved path, injects a small client script into HTML responses to browser requests, and exposes `server.reload()` so a watcher can trigger a reload.

Reload has to survive a full process restart, since a changed setup script or Lambda handler needs a fresh module graph. Use Server-Sent Events rather than a WebSocket: `EventSource` reconnects on its own, so a dropped and restored connection is the signal, with no coordination between the outgoing and incoming process. The server sends a boot id on connect and the client reloads only when the id it reconnects to differs from the first one it saw, so a blip on a still-running process does nothing. A `reloading` event sent before the process goes down lets the page show that a reload is coming rather than a dead connection. `server.reload()` sends an explicit event for a reload that does not involve a restart.

This has to work with no supervisor process involved, so that a dev script run directly from an IDE debugger gets browser reload with the debugger attached throughout.

## Injecting into the response body

`SimAwsResponseHints` states the existing rule: simulator hints go in headers, never the body, so a response keeps the shape the real service returns. Injection breaks that rule deliberately, so it is off by default, reported on startup and in a response header, and never applied to a response an SDK client will parse.

Skip injection when the response is not `text/html`, when the request does not `accept` `text/html` or carries a SigV4 `authorization` or any `x-amz-*` header, when `content-encoding` is set, or when the status is 206. An SDK `GetObject` for an HTML object must come back byte-exact. Injected responses need `content-length` recomputed and `etag` and `last-modified` dropped, since the bytes no longer match the object.

## Acceptance criteria

- [ ] `liveReload` is off by default and changes no byte of any response when off.
- [ ] With it on, an HTML response to a browser-shaped request carries the client script, with `content-length` recomputed and `etag` and `last-modified` dropped.
- [ ] A signed SDK request for an HTML S3 object gets the object bytes unchanged.
- [ ] `server.reload()` reloads a connected client.
- [ ] A client reconnecting to a server with a different boot id reloads; one reconnecting to the same boot id does not.
- [ ] The serving docs cover the opt-in, the reserved path and the fidelity divergence.

## Out of scope

Watching any file. An error overlay for a failed reload; the terminal is enough for now.

## Implementation notes

Intercept the reload path and apply injection in `SimAwsLocalRequestHandler`, either side of `SimAwsHttp.handleRequest`. `SimAwsHttp` is also the SDK interception path and the in-process test path, so keeping the change out of it means only a real browser over the local server ever sees an injected byte. Intercepting the reserved path before Route53 host resolution makes it answer on any served hostname, at the cost of shadowing that path on every simulated distribution, which needs documenting.

Browser per-origin connection limits are worth knowing about: an open EventSource holds one of the six HTTP/1.1 connections per origin per tab.

Depends on #416. An open reload connection otherwise stops the process exiting at all.

hughgrigg:
Opt in with serveSimAws({ liveReload: true }). The local server serves an SSE reload channel on a reserved path and injects a small script into HTML responses to browser requests. A page reloads when it reconnects to a different boot id, so a process restart reaches it with nothing shared between the outgoing and incoming process, and reload() covers a change that needs no restart.

Injection is off by default, reported on startup and in a response header, and never applied to a response an SDK client will parse.

Resolves https://github.com/KensioSoftware/yulin/issues/417

## Summary by CodeRabbit

* **New Features**
  * Added support for serving simulated AWS environments on localhost.
  * Added optional browser live reload for HTML pages, including automatic reloads after server restarts and manual reload triggers.
  * Added local URL conversion and server lifecycle support, including graceful shutdown.
* **Documentation**
  * Added comprehensive serving and live-reload documentation with practical TypeScript examples.
* **Tests**
  * Added extensive coverage for local serving, live-reload behavior, browser integration, response handling, and shutdown scenarios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ npm i -D @kensio/yulin
 ## Feature specific docs
 
 - [AWS SDK interception](./docs/sdk "Simulated AWS SDK docs")
+- [Serving on localhost](./docs/serve "Serving simulated AWS on localhost docs")
 - [Simulated time](./docs/time "Simulated time docs")
 
 ## Usage
@@ -207,6 +208,21 @@ process.on("SIGTERM", () => {
 A restart usually overlaps the process it is replacing. `listen` waits a couple of seconds for a
 pinned port that is still held, then throws `SimAwsLocalPortInUse` naming the port, which means
 something other than the outgoing process owns it.
+
+#### Reload the browser on a restart
+
+Yulin is the only thing in the response path of a page it serves, so it can tell the browser to
+reload. Turning `liveReload` on puts a small script into the HTML pages served to browsers, and the
+page reloads itself when the process restarts:
+
+```typescript
+const srv = await serveSimAws({ simAws, port: 4599, liveReload: true });
+```
+
+`srv.reload()` reloads connected browsers for a change that needs no restart. Live reload is off by
+default, since an injected page is not byte for byte what the real service returns. See the
+[serving docs](./docs/serve "Serving simulated AWS on localhost docs") for what gets the script and
+what does not.
 
 ### Control simulated time
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -25,4 +25,5 @@ behaviour and includes example code that can be copied into tests or local devel
 ## Feature documentation
 
 - [AWS SDK interception](./sdk/ "Simulated AWS SDK usage docs")
+- [Serving on localhost](./serve/ "Serving simulated AWS on localhost usage docs")
 - [Simulated time](./time/ "Simulated time usage docs")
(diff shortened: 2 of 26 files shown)


hughgrigg:
@coderabbitai full review

PR #423, opened 2026-08-06
feat: restart a dev process when project files change

### LINKED ISSUE #418: Reload a simulated AWS local dev process when project files change
## Problem

Running a local dev script that deploys into simulated AWS and serves it on localhost, every source edit means killing the process and starting it again. Lambda handler source, the setup script, seeding code and CloudFront Function modules all need it. `node --watch` can do the restarting, but it tracks only the module graph, so it misses the synthesized template file and any directory mounted with `mountBucketFilesystem(...)`, and the user hand-maintains a `--watch-path` list of paths Yulin is already holding.

## Desired behaviour

A `yulin watch -- <command>` CLI, Yulin's first `bin` entry. It restarts the process on a change and signals the browser, so the loop is edit, save, refreshed page, with nothing typed.

- Watches the working directory by default, ignoring `node_modules`, `.git`, `dist`, CDK asset directories and editor temp files, plus paths the child reports.
- Coalesces a burst of writes into one restart, so one save is one reload.
- Sends a `reloading` event to any connected browser before going down, so the page knows a reload is coming.
- Reports the reload in the terminal, with the time taken and the path that triggered it.
- Detects a restart loop, where setup writes into a watched path, and stops with an error naming the file rather than spinning.

A restart, rather than swapping code in place, is the deliberate choice. A handler is a JS function reference from the user's own module graph, and a fresh process re-imports it with no module cache to defeat, so a restart is correct by construction where an in-process swap would have to invalidate a transitive import graph that ESM does not expose. It also keeps simulated state identical to what a fresh test run sees, and keeps a debugger stepping through ordinary module loads rather than cache-busted URLs.

Editing a bound handler does not change the synthesized template on disk, so a restart re-reads the same template and rebinds without any `cdk synth` in the loop.

## The child needs no particular shape

The CLI does not import the child, does not look for an exported setup function, and does not care whether the simulation was built from SDK commands, a CloudFormation template, or several `SimAws` instances at once. The only change to a dev script is opting into live reload on `serveSimAws`.

Paths flow back over IPC. The parent sets an environment variable, and Yulin in the child reports mounted bucket directories and resolved template paths as they are registered, which the parent adds to its watch set. Nothing depends on this working, it just saves listing paths Yulin is already holding.

Self-respawn is not an option: the shell's job is the original process, so exiting it returns the prompt while an orphan holds the terminal and Ctrl-C stops behaving. The supervisor has to outlive the restarts.

## Debugging

Stepping through a simulated system in a debugger is a large part of the point, so it has to keep working. A run configuration pointed at the CLI attaches to the supervisor, which does nothing interesting, so children need to be spawned with a pass-through inspector port and attached to separately, with automatic reconnect across respawns.

The JetBrains and VS Code behaviour here should be confirmed before this is implemented rather than after, since the docs will carry a recipe. The inspector port also collides across a restart the same way the HTTP port does.

Users who would rather keep a plain "debug this script" run configuration are not left out: the reload channel (#417) and the template watch both work with no supervisor, so an IDE-launched process gets browser and infra reload with the debugger attached throughout, and a handler edit is a manual restart as it is today.

## Acceptance criteria

- [ ] Editing a watched source file restarts the process and reloads a connected browser, on the same port.
- [ ] A burst of writes produces one restart.
- [ ] A mounted bucket directory and a deployed template path are watched without being named.
- [ ] A restart whose setup throws reports the error and leaves the watcher running, so the next save retries.
- [ ] Setup writing into a watched path is detected and refused rather than looping.
- [ ] A debugger stays usable across a respawn.
- [ ] Watch mode is opt-in and inert when not enabled.

## Out of scope

Swapping Lambda or CloudFront Function code without a restart. A fresh process already picks up an edited handler correctly, and an in-process swap would need the ESM import subgraph invalidated, would add a second way to declare a binding, and would re-import under cache-busted URLs that the inspector reports as distinct scripts, putting breakpoint binding at risk.

Delegating the watching to `node --watch`. It restarts unconditionally on any watched change, so Yulin would never see the change and never get to decide what it means, which forecloses updating a stack in place instead of restarting.

Re-synthing CDK. `cdk watch` is a shortcut for `deploy --watch` against real AWS and there is no synth-only watch, so the user runs their own synth and Yulin watches its output. This boundary should be stated plainly in the docs.

Preserving simulated state across a restart. Seeding belongs in the setup script so it reseeds fresh, which keeps local state the same as what tests and CI see.

## Implementation notes

Depends on #416 for the port, and on #417 for the browser signal.

hughgrigg:
Adds `yulin watch -- <command>`, a supervisor that restarts the command on a change and lets live reload refresh the page. It watches the working directory, plus the paths Yulin holds that a module graph never mentions: a directory mounted into a Bucket, and a deployed template, reported over IPC as they are registered.

A burst of writes is one restart, a setup that throws leaves the watcher up, and a process that writes into its own watched path is refused rather than restarted forever.

Resolves https://github.com/KensioSoftware/yulin/issues/418

## Summary by CodeRabbit

* **New Features**
  * Added the `yulin` command-line entry point.
  * Added `yulin watch` to automatically restart development commands when watched files change.
  * Added debounced restarts, graceful shutdown, failure recovery, restart-loop detection, and debugger support.
  * Added watching for project files and registered external paths, including mounted directories and deployed templates.
  * Added local server live reload notifications for watch-mode restarts.

* **Documentation**
  * Documented watch command usage, behavior, supported paths, limitations, and debugger configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,9 +220,19 @@ const srv = await serveSimAws({ simAws, port: 4599, liveReload: true });
 ```
 
 `srv.reload()` reloads connected browsers for a change that needs no restart. Live reload is off by
-default, since an injected page is not byte for byte what the real service returns. See the
-[serving docs](./docs/serve "Serving simulated AWS on localhost docs") for what gets the script and
-what does not.
+default, since an injected page is not byte for byte what the real service returns.
+
+Run the dev script through `yulin watch` and a save is the whole loop, with the process restarted and
+the page refreshed:
+
+```bash
+yulin watch -- tsx dev.ts
+```
+
+It watches the working directory, plus the paths Yulin is holding that a module graph never mentions:
+a directory mounted into a Bucket, and a synthesized template. See the
+[serving docs](./docs/serve "Serving simulated AWS on localhost docs") for what gets the script, what
+is watched, and what is not.
 
 ### Control simulated time
 
--- a/docs/serve/README.md
+++ b/docs/serve/README.md
@@ -191,6 +191,67 @@ The reload channel is served at `/__sim-aws/live-reload` on every hostname the s
 a page can ask for it relative to wherever it was served from. While live reload is on, no simulated
 service can serve anything at that path.
 
+## Restarting on a file change
+
+Live reload gets a page back on its feet after the process restarts. `yulin watch` is what restarts
+it. Run the dev script through it and a save is the whole loop:
+
+```bash
+yulin watch -- tsx dev.ts
+```
+
+Everything after `--` is the command, run as written and restarted when something changes. The CLI
+does not import it, does not look for an exported setup function, and does not care whether the
+simulation was built from SDK commands, a CloudFormation template, or several `SimAws` instances at
+once. The only change to a dev script is turning `liveReload` on.
+
+A restart, rather than swapping code in place, is the deliberate choice. A Lambda handler is a
+function reference out of your own module graph, and a new process re-imports it with no module cache
+to defeat. It also keeps simulated state the same as what a fresh test run sees, since seeding is
+part of the setup script and runs again.
+
+### What is watched
+
+The working directory, minus what nothing edits by hand: `node_modules`, `.git`, `dist`, `coverage`,
+CDK asset directories, and the working files an editor writes around a save.
+
+On top of that, Yulin names paths it is holding that the module graph never mentions. A directory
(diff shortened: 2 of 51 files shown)


hughgrigg:
@coderabbitai full review

PR #424, opened 2026-08-06
feat: watch a deployed template file for changes

### LINKED ISSUE #419: Update a simulated CloudFormation stack in place when its template file changes
## Problem

A local dev process that deployed `cdk.out/MyStack.template.json` has no way to notice the file changed. Restarting for it discards everything held in simulated S3, DynamoDB and SQS for no reason, since a template file is data rather than code. With CDK the common case is a `cdk synth` rewriting the output template while the dev process's own code has not changed at all.

## Current behaviour

`UpdateStackCommand` already reconciles a changed template against a deployed stack: additions created, removals deleted, changed resources replaced, and everything else left holding its data. The CloudFormation docs describe that as what lets a long-running local process pick up an infrastructure change without restarting. Nothing watches the file, so the caller has to notice and send the command themselves. There is no file watching anywhere in `src/`.

## Desired behaviour

Yulin watches the template file of a stack it deployed from a path, and on a change re-reads it and sends `UpdateStackCommand`. Writes are debounced so one save is one update. A template that changed nothing is already refused with `No updates are to be performed.`, so a touched but unchanged file is a no-op. A failed update logs the reason and leaves the previous resources serving, which a restart cannot do. With live reload on, a completed update signals the browser.

This runs in the process that owns the simulation, with no supervisor involved, so a dev script launched straight from an IDE debugger gets infrastructure reload with the debugger attached throughout. Where watch mode is supervising the process, Yulin reports the watched template paths so the parent excludes them from its restart triggers and the in-place update wins.

## Acceptance criteria

- [ ] Rewriting a deployed template file updates the stack without restarting the process.
- [ ] Data in a resource the template did not change survives the update.
- [ ] A burst of writes produces one update.
- [ ] A connected browser reloads once the update completes, not when the write lands.
- [ ] A failed update leaves the previous resources serving and reports why.
- [ ] Under watch mode, a template change updates in place rather than restarting the process.

## Out of scope

In-place resource update. A changed resource is replaced and loses its contents, which is the divergence already recorded in the CloudFormation docs, and it means a template change touching the bucket being served from still costs its objects.

Re-synthing. Yulin watches the synthesized output template, never the CDK app or a YAML source template.

## Implementation notes

The resolved template path is already held on `SimCdkOutContext`. Depends on #417 for the browser signal, and composes with #418.

hughgrigg:
Resolves #419.

The one thing worth a decision: watching is opted into on `deployTemplateFile({ templatePath, watch })`, rather than a separate watch method or a serve-side option.

## Summary by CodeRabbit

* **New Features**
  * Added support for updating deployed CloudFormation stacks from synthesized template files.
  * Added optional template-file watching with debounced, in-place updates and success/failure callbacks.
  * Preserved unchanged resources and their state during updates; changed resources may be replaced.
  * Added controls for listing and stopping active template watchers.
  * Added live reload support for watched simulated environments.

* **Documentation**
  * Added usage guidance, examples, lifecycle details, limitations, and troubleshooting information for template updates and watching.

* **Bug Fixes**
  * Prevented watched template changes from unnecessarily restarting the running process.


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

PR #465, opened 2026-08-06
feat: serve extra file extensions from a mount

hughgrigg:
A mounted Bucket serves files whose extension is on a cautious list of the web's
own types, so that pointing a Bucket at a directory cannot be talked into
reading whatever else is in it. Anything else is not served, and a `GetObject`
for it comes back as though the file were not there.

That is the right default and the wrong answer for a site with a data file of
its own. ChineseBoost serves a pinyin dictionary as three static files — two of
text, one a binary frequency table named `.freq` — and the frequency table was
the single thing the simulated site could not load, so the page came up and then
reported that its dictionary had failed.

## What changes

A mount can name the extensions it needs:

```typescript
simAws.s3().mountBucketFilesystem("site", directory, {
  additionalFileExtensions: [".freq"],
});
```

**Added to the default list rather than replacing it**, which is the one place
this departs from `allowedDirectoryNames` sitting beside it. That list is short
and worth replacing wholesale. This one is long, general, and the reason the
check earns its keep — so a project wanting one more extension should not have
to restate the other twenty-odd, nor be able to drop `.html` by forgetting it.
A leading dot is optional, since nobody means anything by the difference.

Two things follow:

- An extension the content-type table has never heard of is now reachable, and
  is served as `application/octet-stream` rather than with no type at all. That
  is what S3 reports for an object whose type it was never told, and it retires
  a fallback that was previously marked unreachable.
- The lists move to `s3-filesystem-allowed.ts`, leaving the checks that apply
  them in `s3-filesystem-safety.ts`. One file answers "what is allowed" and the
  other "is this allowed" — which also brings the latter back under the FTA
  threshold it had crossed.

## Why not just add `.freq` to the default set

It is a pinyin dictionary's file extension. Baking that into a general AWS
simulator's default allowlist would be the wrong home for it, and the next
project would arrive with a different extension and the same problem.

## Checks

`pnpm check` passes: format, FTA, build, doc examples and coverage at 99.9%
statements / 97.9% branches. Four tests cover the new option — that a named
extension is served, that a dot is optional, that the defaults survive naming
one, and that everything unnamed is still refused.

The docs page gains a section on it, and its example is extracted and
type-checked by `examples:check` like the rest.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring additional file extensions when mounting filesystem-backed S3 storage.
  * Extension entries are case-insensitive and may include or omit a leading dot.
  * Default web file types remain supported, while custom types can now be served.

* **Bug Fixes**
  * Unsupported file types remain hidden as nonexistent.
  * Unrecognized served file types now receive `application/octet-stream` content metadata.

* **Documentation**
  * Added configuration guidance and an example for serving custom binary files.

### DIFF
--- a/docs/services/s3/README.md
+++ b/docs/services/s3/README.md
@@ -1564,7 +1564,7 @@ Filesystem storage is somewhat restrictive to make it slightly safer:
 - The directory must not be the user's home directory
 - The path must not contain `..`
 - Object keys must not be absolute paths or contain `..`
-- Unsupported file extensions are rejected or ignored
+- Only files whose extension is on a cautious list are served; see below
 - Symlinks are ignored when listing Objects
 - Deletion is refused rather than unlinking a real file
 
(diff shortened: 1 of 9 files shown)


hughgrigg:
Worked through the review. All three code findings were valid and are fixed; the fourth (the `.pem` GET test) was a real coverage gap and is added too.

**Content-type wording in the docs.** The sentence read as though any file gets `application/octet-stream`, when in fact an unlisted extension is not served at all. Now scoped: a *served* file whose extension is not in the table gets octet-stream, and that only arises for an extension a mount named itself.

**Empty extension.** `normaliseExtension("")` returned `"."`, and `path.extname` reports `"."` for a name ending in a dot — so a stray empty entry in a mount's list silently became permission to serve `secret.`. It now throws, which is what the rest of this class does with input a caller got wrong, and it fails at mount time rather than quietly.

**Content type asserted on the `.freq` test.** Fair — the metadata fallback was changed in this PR and nothing pinned it. It now asserts `application/octet-stream`.

**`.pem` on the GET path.** Added as its own test rather than folded into the `putObject` one, since they cover different branches: `putObject` throws, `getObject` returns `undefined` for a file already sitting in the directory. That second path is the one that mattered here — a browser asking for it gets nothing, which is exactly how the ChineseBoost symptom presented.

Also added a test that an empty extension is refused. `pnpm check` passes: 99.9% statements, 97.91% branches.

PR #602, opened 2026-08-15
feat: accept MfaConfiguration on a user pool

### LINKED ISSUE #601: Deploy a user pool with MfaConfiguration OPTIONAL
`ChineseboostUserStack` cannot be deployed into a simulation at all, because its pool sets `MfaConfiguration: OPTIONAL` and `CreateUserPool` refuses anything but `OFF`:

    InvalidParameterException: Sim CloudFormation Resource SiteUserPool4E2CF00C creation failed:
    CreateUserPool MfaConfiguration 'OPTIONAL' is not simulated: multi-factor authentication would
    be ignored here and applied on real AWS. Only 'OFF' is supported.

The refusal is the right instinct: `sim-cognito-user-pool-view.js` says `MfaConfiguration` is always `OFF` because MFA is not simulated, and silently accepting the property would be worse than refusing it. The trouble is that it takes the whole stack down rather than the one feature, and a pool offering optional MFA is an ordinary thing to declare.

## Why it bites harder than it looks

The setting is a **product decision recorded in the consuming repository**, not an accident: the pool offers time-based one-time passwords and deliberately not SMS. Nothing in the sign-up or sign-in flows being tested touches MFA at all. So the effect is that a stack is unsimulatable because of a property none of its tests exercise.

Concretely, on chineseboost.com, this blocks covering server-side sign-up against the pool the stack actually declares. That matters because of the *other* half, below.

## The related gap, which is why the stack template is the only way through

`CreateUserPool` also refuses `Schema`:

    InvalidParameterException: CreateUserPool Schema is not simulated: custom attributes would be
    ignored here and applied on real AWS

and then, predictably:

    InvalidParameterException: User attribute 'custom:userId' is not in the pool's schema: a pool
    here holds the standard attributes only, because CreateUserPool refuses a Schema of its own,
    so custom attributes cannot exist

That pool carries a `custom:userId` attribute because **nothing outside Cognito is ever keyed on `sub`**: a `sub` belongs to the pool that issued it, so keying application data on one welds the data to a pool that cannot be moved. Every sign-up writes that attribute, so a pool without it cannot exercise sign-up at all.

Between the two, there is no route to a simulated pool that sign-up works against: not through the SDK, because custom attributes are refused, and not through CloudFormation, because MFA is refused.

## What would close it

Either of these unblocks the CloudFormation route, and the first is the smaller change:

- **Accept `MfaConfiguration` of `OPTIONAL` and `ON` on `CreateUserPool`**, storing it and reporting it back from `DescribeUserPool`, without simulating a second factor in any auth flow. A pool that records the setting and never challenges on it is honest as long as the auth commands say so; refusing the pool outright is the part that costs.
- Or, if a stored-but-unenforced setting is the wrong shape for this simulator, support `Schema` on `CreateUserPool` so custom attributes can exist on a pool built through the SDK, which would give a route that avoids the template.

Supporting both would be better than either. `SetUserPoolMfaConfig` and `GetUserPoolMfaConfig` are already named in the consuming repository's CloudFormation execution policy, because real CloudFormation sets MFA in a second call after the pool exists, so that shape is worth knowing about if this is implemented.

## Testing it

The case that matters is that a pool declaring optional MFA deploys and behaves as an ordinary pool for everything that is not MFA: `SignUp`, `AdminInitiateAuth` and the rest go on working, and `DescribeUserPool` reports back what was asked for rather than `OFF`. If an auth flow would have to challenge for a second factor and cannot, that is the place to refuse, where it says something true and specific, rather than at pool creation.

hughgrigg:
A pool can now be created with an `MfaConfiguration` of `OPTIONAL` or `ON`, and `DescribeUserPool` reports back what it was asked for. `SetUserPoolMfaConfig` and `GetUserPoolMfaConfig` set and read the factors behind it, `SOFTWARE_TOKEN_MFA` being the one this simulation accepts, and CloudFormation deploys `MfaConfiguration` and `EnabledMfas` in that second call once the pool exists, as real CloudFormation does and as a stack's `cognito-idp:SetUserPoolMfaConfig` permission implies. Nothing challenges for a second factor: a pool configured `OPTIONAL` behaves as a real one does for a user that never registered one, so sign-up and sign-in go on working against the pool a stack actually declares, and the refusal `CreateUserPool` used to make moves to the sign-in a pool configured `ON` would have challenged, where it can say what it was that could not be done.

Resolves https://github.com/KensioSoftware/yulin/issues/601

## Summary by CodeRabbit

* **New Features**
  * Added simulated Cognito MFA configuration with `OFF`, `OPTIONAL`, and `ON` modes.
  * Added software-token MFA support and APIs to set and retrieve pool MFA settings.
  * Added CloudFormation and SDK support for deploying and managing MFA configuration.
  * Added MFA behavior documentation and an end-to-end usage example.

* **Bug Fixes**
  * User-pool updates now preserve separately configured MFA factors.
  * Sign-ins requiring unsupported MFA challenges are rejected with clear errors.
  * Unsupported SMS and email MFA configurations now receive targeted validation.

### DIFF
--- a/docs/services/cognito/README.md
+++ b/docs/services/cognito/README.md
@@ -2950,6 +2950,8 @@ Current documented limitations:
 - `SetUserPoolMfaConfig` accepts `SoftwareTokenMfaConfiguration` alone. `SmsMfaConfiguration` and
   `EmailMfaConfiguration` are refused, because a pool here has neither the `SmsConfiguration` nor
   the `EmailConfiguration` real Cognito wants before it will deliver either message.
+  `WebAuthnConfiguration` is refused because a passkey is presented through the `USER_AUTH` flow,
+  which is refused as a flow of its own.
 - `UpdateUserPool` replaces a pool's settings rather than merging into them, as real Cognito does, so
   a setting the request leaves out goes back to the default `CreateUserPool` would have given it. It
   covers the settings this simulation models: `Policies.PasswordPolicy`, `DeletionProtection`,
--- a/docs/services/cognito/README.md
+++ b/docs/services/cognito/README.md
@@ -2242,12 +2242,15 @@ the secret with `DescribeUserPoolClient`, which reports it here as it does on re
 The properties each type reads are the ones this simulation models:
 
 - `AWS::Cognito::UserPool`: `UserPoolName`, `Policies`, `DeletionProtection`, `LambdaConfig`,
-  `AdminCreateUserConfig`, `AutoVerifiedAttributes`, `MfaConfiguration`, `UserPoolTier`,
-  `AccountRecoverySetting`, `EmailVerificationMessage`, `EmailVerificationSubject`,
+  `AdminCreateUserConfig`, `AutoVerifiedAttributes`, `MfaConfiguration`, `EnabledMfas`,
+  `UserPoolTier`, `AccountRecoverySetting`, `EmailVerificationMessage`, `EmailVerificationSubject`,
   `SmsVerificationMessage` and `VerificationMessageTemplate`. `LambdaConfig` is read a trigger at a
   time, so a template naming a trigger this simulation runs deploys and one naming a trigger it
-  does not fails the stack. The last six are accepted at one value each and refused at any other,
-  as `CreateUserPool` refuses them.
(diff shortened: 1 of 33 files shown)


hughgrigg:
@coderabbitai full review

PR #613, opened 2026-08-16
feat: hold custom attributes on a user pool

### LINKED ISSUE #603: Sign up with custom attributes on a simulated Cognito user pool
## Problem

A simulated Cognito user pool holds the standard attributes and nothing else, so no application that keys its own data on a custom attribute can exercise sign-up against a simulated pool. `CreateUserPool` refuses a `Schema`, and every write of a `custom:` attribute is refused after it.

This is the second half of https://github.com/KensioSoftware/yulin/issues/601, which https://github.com/KensioSoftware/yulin/pull/602 resolved for `MfaConfiguration` alone. A custom attribute is the ordinary way to hold an application's own identifier for a user, because a `sub` belongs to the pool that issued it, so keying application data on one welds that data to a pool that cannot be moved.

## Current behaviour

`SimCognitoUnsimulatedUserPoolIdentity` refuses `Schema` on `CreateUserPool`, saying custom attributes are not simulated. `SimCognitoUserAttributes` then checks every attribute name against the fixed standard set in `sim-cognito-standard-attributes.ts` and refuses anything else, naming that first refusal as the reason. `DescribeUserPool` reports no `SchemaAttributes` at all, which is listed as a divergence in `docs/services/cognito`.

CloudFormation does not fail on it: `Schema` is not in the `AWS::Cognito::UserPool` allow-list, so a template declaring one deploys and the property is recorded in `stack.ignoredProperties`. A CDK `UserPool` with `customAttributes` emits exactly that, so the stack comes up and the sign-up it was built for is then refused.

## Desired behaviour

`CreateUserPool` accepts a `Schema`, the pool holds the custom attributes it declares alongside the standard ones, `SignUp`, `AdminCreateUser` and `AdminUpdateUserAttributes` accept and report a `custom:` attribute the schema holds, and one it does not is still refused. `DescribeUserPool` reports the pool's `SchemaAttributes`. The `AWS::Cognito::UserPool` `Schema` property deploys rather than being recorded as ignored.

Real Cognito prefixes a declared attribute with `custom:`, so a `Schema` naming `userId` is written as `custom:userId`. It also refuses a write to an immutable attribute on an existing user, and refuses a `Schema` change on `UpdateUserPool`, both of which are worth keeping.

## Example

```typescript
await cognito.createUserPool(
  new CreateUserPoolCommand({
    PoolName: "myapp-users",
    Schema: [{ Name: "userId", AttributeDataType: "String", Mutable: false }],
  }),
);

await cognito.signUp(
  new SignUpCommand({
    ClientId: clientId,
    Username: "alice",
    Password: "Sup3rSecret!",
    UserAttributes: [{ Name: "custom:userId", Value: "usr_01H8" }],
  }),
);
```

## Acceptance criteria

- [ ] `CreateUserPool` accepts a `Schema` and the pool holds the attributes it declares, under their `custom:` names
- [ ] A `custom:` attribute the schema holds is accepted by `SignUp`, `AdminCreateUser` and `AdminUpdateUserAttributes`, and reported by `AdminGetUser` and `ListUsers`
- [ ] A `custom:` attribute no schema declares is still refused, in words that say the pool's schema does not hold it
- [ ] `DescribeUserPool` reports `SchemaAttributes`, for the standard attributes as well as the declared ones
- [ ] A `Mutable: false` attribute is refused a second write on a user that already has it, as real Cognito refuses one
- [ ] A CloudFormation template declaring a pool with a `Schema` deploys with no ignored property, and `SignUp` against that pool writes the attribute
- [ ] The attribute constraints Cognito applies are honoured or refused rather than dropped: `AttributeDataType`, `StringAttributeConstraints` and `NumberAttributeConstraints`
- [ ] `docs/services/cognito` covers it, and the divergence saying a pool has the standard attributes only is removed

## Out of scope

App client `ReadAttributes` and `WriteAttributes`, which decide what one client may see and set, and `AdminDeleteUserAttributes`, which is not implemented. An identity provider `AttributeMapping` onto a custom attribute is worth checking against the pool's schema in this issue, since that check already exists for the standard ones.

hughgrigg:
CreateUserPool now takes a Schema, and the pool holds what it declares alongside the standard attributes, under the custom: names Cognito gives them. SignUp, AdminCreateUser and AdminUpdateUserAttributes write one, AdminGetUser and ListUsers report it, and DescribeUserPool reports the whole schema. Every write is held to what the declaration said: its type, its bounds, and whether a user that already has the attribute may be given another value, so an application can key its own data on an immutable identifier of the pool's rather than on a sub that belongs to the pool that issued it. An attribute no schema declares is still refused, in words that say what the pool does hold. A template declaring a Schema deploys rather than recording an ignored property, which is what a CDK pool with customAttributes emits, and an identity provider mapping a claim onto a declared attribute is accepted where one onto an attribute the pool has not got is refused.

Resolves https://github.com/KensioSoftware/yulin/issues/603

## Summary by CodeRabbit

* **New Features**
  * Added user-pool schema support for standard and custom attributes.
  * Added validation for attribute types, requiredness, mutability, names, and value constraints.
  * Custom attributes can be read, updated when mutable, and mapped through identity providers.
  * CloudFormation user pools can declare schemas, including constrained string and numeric attributes.
  * User-pool descriptions now report schema attributes.

* **Bug Fixes**
  * Schemas are preserved across pool updates, while unsupported schema updates are rejected.
  * Immutable attributes can no longer be changed after assignment.


hughgrigg:
@coderabbitai full review

hughgrigg:
@coderabbitai full review

hughgrigg:
@coderabbitai full review

PR #619, opened 2026-08-16
feat: deliver log events to Lambda by subscription filter

### LINKED ISSUE #609: Invoke a simulated Lambda from a CloudWatch Logs subscription filter
## Problem

Subscription filters are how teams get log events out of CloudWatch Logs and into their own processing, usually a Lambda that forwards to an error tracker or a metrics sink. That forwarding code is rarely tested, because doing so needs a real log group delivering to a real function.

## Desired behaviour

A subscription filter on a log group delivers matching events to a Lambda function asynchronously, in the real payload shape: a base64-encoded gzipped `awslogs.data` field containing the owner, log group, log stream and the matched events.

## Example

```typescript
await sim.logs().putSubscriptionFilter({
  logGroupName: "/aws/lambda/orders",
  filterName: "errors-to-tracker",
  filterPattern: "ERROR",
  destinationArn: trackerFunctionArn,
});

await sim.lambda().invoke({ FunctionName: "orders", Payload: badOrder });
// the tracker function receives the ERROR lines the orders handler wrote
```

## Proposed scope

`PutSubscriptionFilter`, `DescribeSubscriptionFilters`, `DeleteSubscriptionFilter`, and delivery to a Lambda destination.

## Acceptance criteria

- [ ] A subscription filter can be put, described and deleted.
- [ ] Events matching `filterPattern` are delivered to the destination function; non-matching ones are not.
- [ ] The payload matches AWS's shape: `awslogs.data` base64 of gzipped JSON with `owner`, `logGroup`, `logStream`, `messageType` and `logEvents`.
- [ ] Delivery is asynchronous, so a failing destination does not fail the `PutLogEvents` call that triggered it.
- [ ] A destination the log group has no permission to invoke fails at `PutSubscriptionFilter`, as AWS does.
- [ ] A log group accepts only one subscription filter unless the account has the two-filter quota, matching AWS's limit.

## Out of scope

Kinesis and Firehose destinations, and cross-account destinations. Lambda is the destination worth having first because the delivery machinery for asynchronous invocation already exists.

## Depends on

#606, and #607 for the case above where the events come from a handler rather than a direct `PutLogEvents`.

hughgrigg:
Brief, concise description of the change:

A subscription filter on a log group now delivers the events its pattern matched to a simulated Lambda function, so the code a team wrote to forward log lines to an error tracker or a metrics sink can be tested against the handler it forwards from. The payload is the real one, an `awslogs.data` field holding the base64 of a gzipped document carrying `messageType`, `owner`, `logGroup`, `logStream`, `subscriptionFilters` and `logEvents`, so a handler written against a real subscription decodes it unchanged. Delivery happens on the background scheduler because real CloudWatch Logs answers a write before anything is delivered, so a destination that throws cannot fail the `PutLogEvents` that triggered it, and every failure is kept in `subscriptionFailures` rather than being lost the way an account loses it. The destination is checked when the filter is put, as an account checks it, so a function that never granted the regional `logs.<region>.amazonaws.com` principal permission fails there rather than leaving a filter that silently drops every event, and the resource policy is consulted again on every delivery so a permission taken away later stops delivery too. What a Lambda function logged for itself is delivered as well, since both write paths hand events to the same fan-out, which is what makes a forwarder testable against a real handler's output. Simulated CloudWatch Logs moves off the self-contained service builder to reach Lambda, resolving the destination function when an event is delivered rather than at construction, because every function records its output here and reaching one while this is being built would be a cycle with no bottom; that is the same lazy lookup simulated SNS uses for the same reason. Kinesis, Firehose and cross-account destinations are refused rather than accepted and never delivered to.

Resolves https://github.com/KensioSoftware/yulin/issues/609

- [x] Rebased off latest main

## Summary by CodeRabbit

* **New Features**
  * Added CloudWatch Logs subscription filters with create, list, update, and delete operations.
  * Added asynchronous delivery of matching log events to Lambda functions using AWS-compatible compressed payloads.
  * Added filter limits, pattern matching, permission validation, and delivery-failure tracking.
  * Added support for service-writer-generated logs and standalone Logs behavior.
* **Documentation**
  * Added usage guides and TypeScript examples covering setup, permissions, payload decoding, and asynchronous delivery.
* **Tests**
  * Added comprehensive coverage for filtering, delivery, permissions, failures, limits, and lifecycle scenarios.

### DIFF
--- a/docs/services/logs/README.md
+++ b/docs/services/logs/README.md
@@ -305,6 +305,129 @@ Two divergences are worth knowing about:
   retention ends up correct, but the events the group held are gone, where a real update to
   `RetentionInDays` keeps them.
 
+## Subscription filters
+
+A subscription filter delivers the events matching its pattern to a Lambda function, so the code a
+team wrote to forward log lines to an error tracker or a metrics sink can be tested against the
+handler it forwards from.
+
+```typescript sim-logs-subscription-filter
+/**
+ * Delivering matched log events to a simulated Lambda function.
+ */
+
+import { gunzipSync } from "node:zlib";
+
+import {
+  CreateLogGroupCommand,
+  CreateLogStreamCommand,
+  PutLogEventsCommand,
+  PutSubscriptionFilterCommand,
+} from "@aws-sdk/client-cloudwatch-logs";
+import {
+  AddPermissionCommand,
+  CreateFunctionCommand,
+} from "@aws-sdk/client-lambda";
+
+import { SimAws } from "@kensio/yulin";
+import { makeLambdaZipFileInput } from "@kensio/yulin/lambda";
+
+const simAws = new SimAws();
+const logGroupName = "/aws/lambda/orders";
+const logStreamName = "2026/08/16/[$LATEST]0f7c1a";
+const received: string[] = [];
+
+await simAws.lambda().createFunction(
+  new CreateFunctionCommand({
+    FunctionName: "error-tracker",
+    Role: `arn:aws:iam::${simAws.defaultAccountId}:role/TrackerRole`,
+    Code: {
+      ZipFile: makeLambdaZipFileInput(
+        (event: { awslogs: { data: string } }) => {
+          const decoded = JSON.parse(
+            gunzipSync(Buffer.from(event.awslogs.data, "base64")).toString(),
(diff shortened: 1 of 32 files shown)


hughgrigg:
@coderabbitai full review

Issue #652, opened 2026-08-17
Serve simulated Cognito managed login pages for local user sign-in and sign-up


hughgrigg:
## Problem

A simulated `/oauth2/authorize` can sign in a federated user and has no way to sign in a local one. `SimCognitoAuthorizeInput` carries `identity_provider` and `idp_identifier` and no credentials. So the half of managed login that most applications reach for stays outside this simulation. That half is a local user creating an account, confirming it from the emailed code, and then signing in with an email address and a password.

The consumer need is chineseboost.com again. It is moving email and password sign-in off `AdminInitiateAuth` and onto managed login, because a browser form post cannot reach a Lambda function URL behind CloudFront origin access control at all. See #651. After the move, the site's own sign-in is a redirect to `auth.chineseboost.com` and a callback carrying an authorization code, which is the shape AWS documents for a confidential client. The federated path is covered by the work in #544. The local path is the one with no test, and it is the path every account on that site will take.

## Current behaviour

- `SimCognitoDomainEndpoints` serves `GET /oauth2/authorize`, `POST /oauth2/token` and `GET /logout`, and hands authorize straight to `hostedAuthorize` with the query parameters.
- `hostedAuthorize` signs in the external user held by the identity provider the request names, and answers a redirect carrying a code. A request naming no provider has nobody to sign in.
- `AWS::Cognito::UserPoolDomain` already accepts `ManagedLoginVersion`, so a stack asking for managed login deploys and the domain is served on its own hostname.
- `AWS::Cognito::ManagedLoginBranding` is an unsupported resource type, recorded and stepped over.
- `SignUp`, `ConfirmSignUp` and `ResendConfirmationCode` are already simulated, as is the credential check `InitiateAuth` makes.

So the serving, the hostname routing and every operation these pages need are built already. What is missing is the pages, and the wiring from a form post to the operation behind it.

## Desired behaviour

Everything starts at the authorize endpoint, as it does on real Cognito.

**In process first**, because most tests will use this and no HTML is involved. `hostedAuthorize` accepts a username and a password beside the parameters it already takes, checks them the way `InitiateAuth` checks them, and answers the same redirect carrying a code. A wrong password, an unknown user and an unconfirmed account answer with what they answer elsewhere in this simulation.

**Then three served pages**, each reached from the one before it.

[...]

PR #655, opened 2026-08-17
feat: sign a local user in at authorize

hughgrigg:
`/oauth2/authorize` signs one of a pool's own users in where the request names no identity provider. The username and password real managed login takes from its form arrive beside the parameters the request already carries, and are checked with the checks `InitiateAuth` makes, so a wrong password, an unknown user and an unconfirmed account are refused the way they are refused everywhere else in this simulation. The app client needs `COGNITO` among its `SupportedIdentityProviders`, as it does on real Cognito. A sign-in real managed login would answer with a further page, which is a user owing a second factor and a user holding a temporary password, is refused with a message naming the page and where the simulation does issue that challenge.

This is the first half of #652. The three served pages, and the form posts behind them, follow in a second pull request. Splitting it keeps this one to the sign-in itself, which is what most tests will reach for and what involves no HTML at all.

Part of #652

### DIFF
--- a/docs/services/cognito/README.md
+++ b/docs/services/cognito/README.md
@@ -1568,6 +1568,74 @@ An authorization code is single use and lasts five minutes on the simulated cloc
 endpoint also answers a `grant_type` of `refresh_token`, with the refresh token the grant handed
 out.
 
+### Signing a local user in
+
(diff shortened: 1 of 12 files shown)


hughgrigg:
@coderabbitai full review

PR #714, opened 2026-08-18
feat: serve IAM and ELBv2 over the Query protocol

### LINKED ISSUE #697: Serve IAM and ELBv2 over the Query endpoint, so an environment can set itself up
## Problem

`aws iam` and `aws elbv2` against a served simulated environment are refused with `501 Not Implemented`, and so is any SDK client pointed at the endpoint.

IAM is the valuable half. Every served request is authorized as the principal that signed it. Setting up the users, roles and policies a hybrid environment needs therefore has to happen in the Node process that built the simulation. A shell script or a container that creates its own identity has no way in, which makes the endpoint awkward to use for exactly the setup step it needs most.

## Current behaviour

Both speak the Query protocol. Simulated IAM routes fifteen SDK Commands, including `CreateUser`, `CreateRole`, `CreateAccessKey`, `PutUserPolicy` and `PutRolePolicy`. Simulated ELBv2 routes twenty-two, all of them load balancer, listener, rule and target group operations.

Neither is reachable over an endpoint URL. The Query reader lives in `src/service/sts/serve/` and serves STS alone.

## Desired behaviour

`aws iam create-user`, `aws iam create-access-key` and the rest of what simulated IAM implements answer over an endpoint URL, so an environment can be set up from outside the process that built it. ELBv2 answers the same way.

## Example

```bash
export AWS_ENDPOINT_URL=http://localhost:8787
aws iam create-user --user-name deployer
aws iam create-access-key --user-name deployer
```

## Acceptance criteria

- [ ] IAM and ELBv2 answer an endpoint URL for the operations each simulated service implements
- [ ] Each states its own XML namespace in the response envelope
- [ ] An operation a service has not implemented is refused in the Query error shape, so an SDK raises it by name
- [ ] A `.loc.test.ts` creates an IAM User over the endpoint and signs a later served request with the access key it answers with
- [ ] `docs/serve` lists both among the services that answer

## Blocked by

#685, which moves the Query reader out of `src/service/sts/serve/` to somewhere every Query service reaches it. This is the second consumer of that move and adds no protocol work of its own.

## Out of scope

The IAM operations simulated IAM has not implemented, such as `ListUsers` and `GetUser`. Those are worth their own issue, and this one covers reaching what exists.

## Implementation notes

The identity a served request is authorized as comes from its SigV4 signature. An access key created over the endpoint therefore has to sign the requests that follow. That path already works for the temporary credentials `AssumeRole` issues, which is worth confirming holds for a long-lived key created this way.

hughgrigg:
`aws iam` and `aws elbv2` were refused with 501 against a served simulated environment. IAM is the half that matters. Every served request is authorized as the principal that signed it, so setting up users, roles and policies had to happen in the Node process that built the simulation, and a container or a shell script could not create its own credentials. Simulated IAM now serves the fifteen operations it implements and simulated ELBv2 the twenty-two, both on the shared Query layer from #698, each with its own XML namespace and operations map and each registered under the signing name the CLI actually sends (`iam` and `elasticloadbalancing`, checked against captured SDK requests rather than assumed). An operation a service has not implemented is refused in the Query error shape, so an SDK raises it by name. The `.loc.test.ts` that closes the loop creates an IAM User over the endpoint, asks the same endpoint for an access key, and signs a later served request with what it was answered.

Resolves https://github.com/KensioSoftware/yulin/issues/697

- [x] Rebased off latest main

Two notes for review. The shared Query layer has no reader for a list of plain values and no writer for a nested structure, and both services need them, so each carries a small local helper (`elbV2QueryValues`, `elbV2QueryStructure`, `iamQueryEntity`). Lifting them into `src/serve/http/api/query/` is the obvious follow-up, and that directory was left alone here to stay out of the way of concurrent work. Separately, simulated IAM holds a policy's `AttachmentCount` at zero whatever is attached to it, which makes `ListPolicies --only-attached` answer with nothing. That is pre-existing and untouched.

## Summary by CodeRabbit

* **New Features**
  * Added simulated IAM support for users, roles, access keys, inline policies, and managed policies.
  * Added simulated Elastic Load Balancing v2 support for load balancers, target groups, listeners, certificates, rules, routing, and target health.
  * Registered both services with the served AWS endpoint.
* **Documentation**
  * Added CLI usage guidance and supported-operation details for IAM and ELBv2.
* **Tests**
  * Added comprehensive integration coverage for IAM and ELBv2 workflows.

### DIFF
--- a/docs/serve/README.md
+++ b/docs/serve/README.md
@@ -246,7 +246,7 @@ A request carrying no signature is anonymous and reaches nothing. In process an 
 
 ### Which services answer
 
-S3, STS, SNS, CloudFormation, Lambda, and the services that speak the AWS JSON protocol. Those are DynamoDB, DynamoDB Streams, SQS, Cognito Identity Provider, EventBridge, ECS, SSM, ACM, CloudWatch, CloudWatch Logs, KMS, Secrets Manager and Rekognition.
+S3, STS, IAM, ELBv2, SNS, CloudFormation, Lambda, and the services that speak the AWS JSON protocol. Those are DynamoDB, DynamoDB Streams, SQS, Cognito Identity Provider, EventBridge, ECS, SSM, ACM, CloudWatch, CloudWatch Logs, KMS, Secrets Manager and Rekognition.
 
 A request to any other service is refused with `501 Not Implemented` and a body saying why. Every service is reachable in process through `SimAws` and through [SDK interception](../sdk/README.md), whether or not it answers here.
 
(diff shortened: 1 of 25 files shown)


hughgrigg:
@coderabbitai full review

PR #715, opened 2026-08-18
feat: serve the REST-JSON protocol for Lambda

### LINKED ISSUE #696: Serve the REST-JSON protocol so aws lambda invoke reaches a simulated function
## Problem

`aws lambda invoke` against a served simulated environment is refused with `501 Not Implemented`, and so is any SDK client pointed at the endpoint. Invoking a simulated function from a container, from a Python application or from a shell script is the thing the served endpoint exists to make possible, and Lambda is the service people reach for first.

REST-JSON is the last protocol family with nothing built. Four simulated services speak it: Lambda, API Gateway v2, SES v2 and EventBridge Scheduler.

## Current behaviour

The endpoint reads three protocols. AWS JSON names its operation in the `x-amz-target` header. S3 REST-XML names its operation in the method, the path and a query-string sub-resource. STS Query names its operation in a form-encoded `Action` field.

REST-JSON names its operation in the method and the path, as REST-XML does, and carries JSON in the body where REST-XML carries a document. Nothing reads it, so a Lambda control-plane request falls through to the AWS JSON dispatcher and is refused for having no `x-amz-target`.

Simulated Lambda routes sixteen SDK Commands, `InvokeCommand` among them.

## Desired behaviour

`aws lambda invoke` runs a simulated function and writes its response, and the other Lambda control-plane operations the simulator implements answer over the same endpoint URL.

Reading a REST-JSON request works the way the S3 route table does. The other three services then become incremental once Lambda is in.

## Example

```bash
export AWS_ENDPOINT_URL=http://localhost:8787
aws lambda invoke --function-name orders --payload '{"id":1}' /tmp/out.json
cat /tmp/out.json
```

## Acceptance criteria

- [ ] A route table maps method and path to the Lambda Commands simulated Lambda implements, as `sim-s3-api-*routes.ts` does for S3
- [ ] `Invoke` runs the function and answers with its payload, its status and the `X-Amz-Function-Error` header a handler failure sets
- [ ] Operation output is written as JSON, with the members the SDK reads
- [ ] An error reaches the client under the name real Lambda gives it, so the SDK raises it rather than failing to parse
- [ ] A path this endpoint has no operation for is refused by name, and never falls through to one sharing its method
- [ ] A `.loc.test.ts` drives a real `LambdaClient` against a served endpoint URL through an invoke
- [ ] `docs/serve` lists Lambda among the services that answer

## Out of scope

API Gateway v2, SES v2 and EventBridge Scheduler, which each want their own route table on top of the reading this builds. A Lambda Function URL, which is already served on its own hostname and is a separate surface from the control plane.

## Implementation notes

The S3 work in #676 is the model, and the lesson from it is worth repeating here. A path this endpoint has no operation for has to be refused. Ignoring one answers the question with whatever operation shares its method and path. That bug shipped in S3, and driving the real CLI caught it where the unit tests passed.

`SimAwsApiEndpoint` picks a protocol from the SigV4 signing name with an `if` per service. #685 proposes replacing that with a registry, and this adds the fourth entry either way.

hughgrigg:
`aws lambda invoke` against a served simulated environment was refused with `501 Not Implemented`, and so was any SDK client pointed at the endpoint. REST-JSON names its operation in the method and the path, and nothing at the endpoint read that, so a Lambda control-plane request fell through to the AWS JSON dispatcher and was refused for having no `x-amz-target`. A shared REST-JSON layer under `src/serve/http/api/rest-json/` now matches a request against a table of path templates and answers in JSON, and simulated Lambda registers the sixteen operations it implements. `Invoke` answers with the function's payload, its status and the `X-Amz-Function-Error` header a handler failure sets. A path with no route is refused by name rather than matched against the nearest operation sharing its method, which was the bug the S3 work shipped. API Gateway v2, SES v2 and EventBridge Scheduler each want a route table of their own on top of this.

Resolves https://github.com/KensioSoftware/yulin/issues/696

The route table was checked against botocore's own Lambda model, and every method, path and response status here is the one that model states. The endpoint was then driven with the real `aws` CLI: an invoke, a handler failure, a function created from a zip file, the policy and Function URL operations, and `list-functions`, `list-aliases` and `get-function-concurrency`, each of which is refused by name at the path it arrived at.

- [x] Rebased off latest main

### DIFF
--- a/docs/serve/README.md
+++ b/docs/serve/README.md
@@ -246,7 +246,7 @@ A request carrying no signature is anonymous and reaches nothing. In process an 
 
 ### Which services answer
 
-S3, STS, SNS, CloudFormation, and the services that speak the AWS JSON protocol. Those are DynamoDB, DynamoDB Streams, SQS, Cognito Identity Provider, EventBridge, ECS, SSM, ACM, CloudWatch, CloudWatch Logs, KMS, Secrets Manager and Rekognition.
+S3, STS, SNS, CloudFormation, Lambda, and the services that speak the AWS JSON protocol. Those are DynamoDB, DynamoDB Streams, SQS, Cognito Identity Provider, EventBridge, ECS, SSM, ACM, CloudWatch, CloudWatch Logs, KMS, Secrets Manager and Rekognition.
 
 A request to any other service is refused with `501 Not Implemented` and a body saying why. Every service is reachable in process through `SimAws` and through [SDK interception](../sdk/README.md), whether or not it answers here.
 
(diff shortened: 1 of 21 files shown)


hughgrigg:
@coderabbitai full review

PR #764, opened 2026-08-19
feat: publish Lambda versions and aliases

### LINKED ISSUE #757: Publish simulated Lambda function versions and aliases, and invoke by qualifier
## Problem

A deployed application points its integrations at a published function version or at an alias, so the thing invoked stays fixed while `$LATEST` moves. Simulated Lambda holds one function per name at one version, so code that publishes a version or creates an alias has nowhere to run, and every simulated service that could name one refuses the qualifier.

## Current behaviour

Functions live in a `Map` keyed by name (`src/service/lambda/function/sim-lambda-function.ts`), and `configuration()` reports `Version: "$LATEST"` always. `PublishVersion`, `ListVersionsByFunction`, `CreateAlias`, `UpdateAlias`, `GetAlias`, `ListAliases` and `DeleteAlias` have no handlers. `InvokeCommand` declares `Qualifier` on its input and never reads it (`src/service/lambda/command/invoke/invoke.handler.ts`), so a caller asking for `PROD` runs `$LATEST` and is told `ExecutedVersion: "$LATEST"`.

## Desired behaviour

`PublishVersion` snapshots the function as it stands under the next version number, starting at `1`, and answers with that version's configuration and ARN. A published version keeps the code, handler, timeout, memory and environment it was published with, and later changes to the function leave it alone. `CreateAlias` and `UpdateAlias` point a name at a version, and `GetAlias`, `ListAliases` and `DeleteAlias` read and remove them.

`Invoke` resolves the version to run from its `Qualifier`, or from a qualifier on the `FunctionName` ARN, and reports the resolved version number as `ExecutedVersion`. A qualifier naming no version or alias fails with `ResourceNotFoundException`. `GetFunction` takes a `Qualifier` the same way.

## Example

```ts
const published = await lambda.send(
  new PublishVersionCommand({ FunctionName: "orders" }),
);
await lambda.send(
  new CreateAliasCommand({
    FunctionName: "orders",
    Name: "live",
    FunctionVersion: published.Version,
  }),
);
const invoked = await lambda.send(
  new InvokeCommand({ FunctionName: "orders", Qualifier: "live" }),
);
// invoked.ExecutedVersion === "1"
```

## Acceptance criteria

- [ ] `PublishVersion` answers with `1` for a function's first version, and numbers each later one upwards.
- [ ] A published version runs the code it was published with after the function's code and configuration change.
- [ ] `CreateAlias`, `UpdateAlias`, `GetAlias`, `ListAliases` and `DeleteAlias` work against a published version, and an alias for a version that does not exist is refused.
- [ ] `Invoke` with a `Qualifier` runs the version the qualifier names and reports it as `ExecutedVersion`.
- [ ] `Invoke` with a qualified `FunctionName` ARN resolves the same way, and an unknown qualifier fails with `ResourceNotFoundException`.
- [ ] `ListVersionsByFunction` lists `$LATEST` and each published version.

## Out of scope

- `Qualifier` on the permission commands. That is its own issue.
- Accepting a qualified ARN in the services that refuse one today. Those follow, service by service.
- `AWS::Lambda::Version` and `AWS::Lambda::Alias` in a CloudFormation template.
- Qualified Function URLs, alias `RoutingConfig` weights and provisioned concurrency.
- `RevisionId`, and the `CodeSha256` and `Description` checks `PublishVersion` makes.

## Implementation notes

- `parseSimLambdaFunctionArn` already reads a qualifier off an ARN and reports it. Four services read that field to refuse a qualified target (S3 notifications, SNS subscriptions, CloudWatch Logs subscriptions and Cognito triggers), and their refusal messages say simulated Lambda has no versions.
- `simLambdaFunctionNameOf` drops a qualifier on the way to a function name. Event source mappings and the CloudFormation target-function reader both depend on that, and EventBridge targets drop one too.
- The function map is keyed by name and read straight through by `SimLambdaFunctionLookup`. Versions need somewhere to live that keeps a name resolving to `$LATEST` for every caller that passes no qualifier.

hughgrigg:
PublishVersion copies a function as it stands and numbers the copy, and CreateAlias, UpdateAlias, GetAlias, ListAliases and DeleteAlias give those numbers names. Invoke and GetFunction resolve a Qualifier, or one appended to the function name, and an invocation reports the number it ran as ExecutedVersion. Versions and aliases are held beside the function map, so a bare function name still resolves to `$LATEST` for every caller that passes no qualifier, and deleting a function drops both.

A published version is a copy of the function rather than a second reference to it, so what was published keeps its code, handler, timeout, memory and environment. `UpdateFunctionCode` is still absent, so nothing can make `$LATEST` drift from a published version yet. The four services that refuse a qualified ARN, `Qualifier` on the permission commands, `AWS::Lambda::Version` and `AWS::Lambda::Alias`, and the API Gateway URI are the rest of the series and are left alone here.

Resolves #757

## Summary by CodeRabbit

- **New Features**
  - Added Lambda function version publishing and version listing.
  - Added alias creation, updating, retrieval, listing, and deletion.
  - Added invocation and configuration support using version or alias qualifiers.
  - Added SDK interception support for version and alias commands.
  - Added an example demonstrating published versions and aliases.

- **Documentation**
  - Expanded Lambda guidance to cover versions, aliases, qualified requests, and limitations.
  - Clarified that HTTP serving supports 16 Lambda operations; version and alias commands remain available in-process or through SDK interception.

- **Bug Fixes**
  - Function deletion now also removes associated versions and aliases.

### DIFF
--- a/docs/serve/README.md
+++ b/docs/serve/README.md
@@ -457,7 +457,7 @@ aws lambda create-function --function-name orders \
   --zip-file fileb://orders.zip
 ```
 
-The operations served are the sixteen simulated Lambda implements:
+The operations served are sixteen of the ones simulated Lambda implements:
 
 - **Functions** — `CreateFunction`, `GetFunction`, `DeleteFunction`, `Invoke`
 - **Function URLs** — `CreateFunctionUrlConfig`, `GetFunctionUrlConfig`,
@@ -465,6 +465,9 @@ The operations served are the sixteen simulated Lambda implements:
 - **Permissions** — `AddPermission`, `RemovePermission`, `GetPolicy`
 - **Event source mappings** — `CreateEventSourceMapping`, `GetEventSourceMapping`,
   `ListEventSourceMappings`, `DeleteEventSourceMapping`
+
+The version and alias operations have no route here yet, and reach the simulation through `SimAws`
+or SDK interception instead.
 
 Anything else is refused as `NotImplemented`, which an SDK raises under that name. The refusal names
 the path it arrived at. `aws lambda list-functions` reports that `GET /2015-03-31/functions` is
--- a/docs/services/lambda/README.md
+++ b/docs/services/lambda/README.md
@@ -702,6 +702,88 @@ console.log(dryRunOutput.StatusCode);
 `Event` invocation handler errors are dropped. Asynchronous retries and failure destinations are
 left out so far.
 
+## Versions and aliases
+
+`PublishVersionCommand` copies a function as it stands and numbers the copy. The first is version
+`1`, and each later one counts up. The copy keeps the code, handler, timeout, memory and
+environment it was published with. `CreateAliasCommand` gives one of those versions a name, and
+`UpdateAliasCommand` moves the name to another version, so a deployment can repoint what callers
+invoke.
+
+`InvokeCommand` and `GetFunctionCommand` take a `Qualifier` naming a version number or an alias.
+The same qualifier can travel on the `FunctionName` instead, appended to the name (`orders:live`)
+or to a function ARN. `ExecutedVersion` on the invocation output is the version number that ran,
+and an invocation through an alias reports the version behind it. Inside the handler, `context`
+carries that number as `functionVersion` and the qualified ARN as `invokedFunctionArn`.
+
+```typescript sim-lambda-versions-and-aliases
+/**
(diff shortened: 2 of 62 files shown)


hughgrigg:
@coderabbitai rate limit

hughgrigg:
@coderabbitai full review

hughgrigg:
@coderabbitai full review

Issue #920, opened 2026-08-22
Retry and Catch error handling in simulated Step Functions


hughgrigg:
## Problem

Retry and Catch are the reason a workflow is a state machine and not a function calling three other functions. They hold the behaviour a team most wants covered by a test. Does a throttled write get retried, does the fourth failure reach the compensating branch, does the error land in `$.error` where the next state expects it.

A state machine without them, run in simulated Step Functions, fails an execution on the first task error and there is nothing to assert about recovery.

## Desired behaviour

- A `Retry` array on a `Task` state retries a failing task on the simulated clock. `ErrorEquals`, `IntervalSeconds` (default 1), `MaxAttempts` (default 3), `BackoffRate` (default 2.0) and `MaxDelaySeconds` are all read.
- Each attempt is scheduled at its own due instant, and an execution sits `RUNNING` between them. One `advanceBy` covering the whole backoff walks every attempt, because work an advance schedules settles before the call returns where it is itself due by the new time (`docs/time/README.md`). A test advances once and then asserts.
- A retry that exhausts its attempts leaves `advanceBy` returning normally. The failure is read from `DescribeExecution`, following the interpreter issue.
- A `Catch` array routes a failure to another state. `ErrorEquals`, `Next` and `ResultPath` are read, and `ResultPath` writes `{ Error, Cause }` into the state input.
- Retriers are tried in order and the first matching `ErrorEquals` wins. Catchers work the same way.
- Error names match. A Lambda function that throws surfaces as `States.TaskFailed` through `lambda:invoke`, and as its own error type through a bare function ARN. `States.ALL` matches anything, and `States.TaskFailed`, `States.Timeout` and `States.Runtime` are raised where they belong.
- `TimeoutSeconds` and `HeartbeatSeconds` on a `Task` state fail it with `States.Timeout` on the simulated clock.
- A `Retry` that exhausts its attempts falls through to the state's `Catch`, and then fails the execution.
- The inspection accessor reports attempts, not only the states visited, so a test can assert that a task ran three times.

## Example

```typescript
Check: {
  Type: "Task",
  Resource: "arn:aws:states:::lambda:invoke",
  Parameters: { FunctionName: "check-enrolment" },
  Retry: [{ ErrorEquals: ["States.TaskFailed"], IntervalSeconds: 2, MaxAttempts: 3 }],
  Catch: [{ ErrorEquals: ["States.ALL"], Next: "Compensate", ResultPath: "$.error" }],
[...]

Issue #918, opened 2026-08-22
Run a simulated Step Functions state machine from Amazon States Language


hughgrigg:
## Problem

A state machine is among the hardest things in an AWS application to cover with a test. The workflow lives in a template as data, and no unit test runs it. A test can call each Lambda function in the workflow by hand, and everything between them goes uncovered (the Choice branches, the payload reshaping, the order the states run in). The alternative is Step Functions Local, a Java process in a container, which pulls a test suite out of the process it runs in.

Yulin simulates nothing under `states` today. A CloudFormation deploy skips `AWS::StepFunctions::StateMachine`, and an `SFNClient` handed to `SimSdk` is refused with `SimSdkUnknownServiceError` (`src/sdk/error/sim-sdk.error.ts`).

## Desired behaviour

Simulated Step Functions interprets Amazon States Language and runs a state machine against the rest of the simulation.

- `CreateStateMachine`, `DescribeStateMachine`, `UpdateStateMachine`, `ListStateMachines` and `DeleteStateMachine` work through `simAws.stepFunctions()` and through an intercepted `SFNClient`. The definition arrives as an ASL JSON string.
- `StartExecution` answers with an execution ARN and runs the execution on the background scheduler (`src/util/background/background.ts`). `DescribeExecution` reports `RUNNING`, `SUCCEEDED`, `FAILED` or `TIMED_OUT`, with the execution output.
- Six state types run. `Pass`, `Task`, `Choice`, `Wait`, `Succeed` and `Fail`.
- A `Task` state invokes a simulated Lambda function, through `arn:aws:states:::lambda:invoke` and through a bare function ARN. The first wraps the answer in `Payload` and `StatusCode`, and the second answers with the payload alone. CDK's `LambdaInvoke` emits both, depending on `payloadResponseOnly`.
- `InputPath`, `Parameters`, `ResultSelector`, `ResultPath` and `OutputPath` apply in that order, with `.$` paths read by a JSONPath subset (root, child by dot and by bracket, array index).
- Choice carries the string, numeric, boolean and timestamp comparators, their `Path` variants, `IsPresent`, `IsNull`, `And`, `Or`, `Not` and `Default`.
[...]

PR #924, opened 2026-08-22
feat: read Amazon States Language data flow

hughgrigg:
Amazon States Language moves data through five fields on every state, and this repository had no way to read a path with. EventBridge refuses `InputPath` and `InputTransformer` for that reason. This adds the layer the rest of #918 rests on, and nothing here reaches a simulated service yet.

## What is in it

- **Reference Paths.** The subset Amazon States Language itself uses. A document root, a child by dot or by bracketed name, and an array element by index. The wider JSONPath grammar (wildcards, filters, slices, recursive descent) is refused by name.
- **Intrinsic functions.** `States.Format`, `States.Array`, `States.ArrayLength`, `States.StringToJson` and `States.JsonToString`, including nested calls and quoted arguments carrying commas. `States.UUID` and `States.MathRandom` are left out, since both answer differently on every call.
- **Payload Templates.** `Parameters` and `ResultSelector`, with `.$` fields resolved from a path or an intrinsic, and objects and arrays walked.
- **The five fields in order.** `InputPath` then `Parameters` for the effective input, and `ResultSelector` then `ResultPath` then `OutputPath` for the effective output. `ResultPath` reads the raw state input, so a state can keep fields `InputPath` had already taken away.
- **Failures carry their Amazon States Language error name.** `States.ParameterPathFailure` and the rest are held on the error, which is what a `Retry` or a `Catch` will match on in #920.

## Two decisions worth a look

An escaped brace survives the string-literal reader and is resolved by `States.Format`. Resolving `\{` when the quotes come off would take away the only thing telling a literal brace apart from a placeholder. A brace escape reaching another intrinsic arrives with its backslash, and the docs will record that divergence.

The context object is refused by name. `$$.Execution.Name` needs an execution to exist, and that arrives with the interpreter.

## Size

This is the first of the pull requests against #918, split the way that issue's implementation notes describe. The interpreter, the state types and the service surface follow. The data-flow layer is where a subtle mistake is most expensive, and it is worth reviewing on its own.

## New Features
- Added Amazon States Language data-flow simulation, including input/output paths, parameters, result selection, and result insertion.
- Added reference-path parsing and traversal for nested objects and arrays.
- Added intrinsic function evaluation for formatting, arrays, length checks, and JSON conversion.
- Added payload-template evaluation with path and intrinsic expressions.
- Added clearer errors for invalid paths, expressions, and unsupported simulation features.

## Tests
- Added comprehensive coverage for data flow, reference paths, intrinsic functions, nesting, escaping, arrays, and error cases.

### DIFF
--- a/src/service/stepfunctions/data/sim-states-data-flow.iso.test.ts
+++ b/src/service/stepfunctions/data/sim-states-data-flow.iso.test.ts
@@ -1,7 +1,9 @@
 import {
+  assertIdentical,
   assertObjectEquals,
   assertStringIncludes,
   assertThrowsError,
+  assertTrue,
 } from "@kensio/smartass";
 import { describe, it } from "vitest";
 import {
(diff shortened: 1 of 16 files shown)


hughgrigg:
@coderabbitai full review

PR #925, opened 2026-08-22
feat: trigger a Lambda from a Kinesis stream

### LINKED ISSUE #909: Trigger a simulated Lambda from a Kinesis stream event source mapping
## Problem

A Lambda event source mapping can name an SQS queue or a DynamoDB stream. `simLambdaEventSourceArnOf` (`src/service/lambda/event-source/sim-lambda-event-source-arn.ts`) refuses everything else, and a mapping naming `arn:aws:kinesis:eu-west-2:111111111111:stream/orders` fails at creation with `InvalidParameterValueException`. Reading a Kinesis stream from a Lambda function is the commonest thing anyone does with one.

The polling machinery comes close and stops short. `SimLambdaEventSourceStreamService` is written around DynamoDB Streams. Its stream description carries `TableName`, its record body carries `Keys`, `NewImage` and `OldImage`, and `SimDynamoDbEventSourceStreamShard` reads the single shard a DynamoDB stream has.

## Desired behaviour

A mapping naming a Kinesis stream ARN polls that stream and invokes the function with a Kinesis event.

Each record arrives under `eventSource` of `aws:kinesis`, with the payload base64-encoded in `kinesis.data` alongside `partitionKey`, `sequenceNumber`, `kinesisSchemaVersion` and `approximateArrivalTimestamp` in epoch seconds. `eventID` is the shard identifier and the sequence number joined by a colon, and `eventSourceARN` is the stream.

The poller reads every shard of the stream and keeps a position for each one. Batching, `StartingPosition`, retry backoff and `ReportBatchItemFailures` behave as they do for a DynamoDB stream mapping, since they come from the shared poll turn.

Polling authorizes as the function's execution role through `DescribeStream`, `GetShardIterator` and `GetRecords`, the way the DynamoDB poller does. A role missing `kinesis:GetRecords` leaves the mapping unable to poll, reported the same way.

## Implementation notes

`SimLambdaEventSourceStreamService` and `SimLambdaEventSourceStreamRecord` need generalising before a second stream service can plug into them, since both are DynamoDB-shaped today.

The union in `sim-lambda-event-source-arn.ts` and the switch in `makeSimLambdaEventSourcePoller` each gain a `kinesis` case. The comment on that switch already says a new source kind belongs there.

Multi-shard reading is new. The DynamoDB poller reads one shard, because a simulated table is one partition.

## Acceptance criteria

- [ ] A mapping created against a Kinesis stream ARN polls it, and a record put onto the stream reaches the function.
- [ ] The event a function receives matches the Kinesis event shape, with `data` base64-encoded and `approximateArrivalTimestamp` in epoch seconds.
- [ ] A stream with two shards delivers every record, whichever shard it landed on.
- [ ] Records batch into one invocation up to `BatchSize`, and a function reporting `batchItemFailures` has the mapping resume from the sequence number it named.
- [ ] An execution role without `kinesis:GetRecords` fails to poll, and the failure is visible to a test.
- [ ] `AWS::Lambda::EventSourceMapping` naming a Kinesis stream deploys and starts polling.
- [ ] The Lambda docs cover Kinesis as an event source.

## Out of scope

- **Enhanced fan-out.** A mapping whose `EventSourceArn` names a consumer is refused.
- **`BisectBatchOnFunctionError`, `MaximumRetryAttempts` and `DestinationConfig`.** `create-event-source-mapping-refusals.ts` already refuses all three, for a DynamoDB stream as much as for Kinesis. Leave that refusal in place.
- **`MaximumBatchingWindowInSeconds` above zero.** Refused today for the same reason.
- **`AWS::Kinesis::Stream`.** Deploying the stream itself is a separate issue, and this one can be built against a stream created through the SDK.

## Depends on

https://github.com/KensioSoftware/yulin/issues/908 supplies the commands this poller reads through.

hughgrigg:
An event source mapping can now name a Kinesis stream, and the records put onto it reach the function as a Kinesis event with the shape real Lambda sends: the payload base64 encoded under `kinesis.data`, the arrival as epoch seconds, and an `eventID` joining the shard to the sequence number. A stream is read by one processor per shard, as real Lambda reads one, so each shard keeps its own place, delivers its own batches and backs off on its own when one fails, and a failing batch blocks its own shard and no other. Polling authorizes as the function's execution role through DescribeStream, GetShardIterator and GetRecords, so a role missing one of them is refused when the mapping is created rather than left subscribed to nothing. A handler that puts records back onto the stream that invoked it is refused, the way one writing back into its own source table already was.

Getting there generalised three pieces that were DynamoDB-shaped. A starting position is now a position and, for AT_TIMESTAMP, the instant it needs, which is what lets a Kinesis mapping take the third position real Lambda offers. The batch response reads a failure report against the sequence numbers a batch held rather than against DynamoDB records, and the stream progress takes where a read left the reader rather than the records it read. The cascade guard describes whichever source fed itself. `AWS::Lambda::EventSourceMapping` needed no change and deploys a Kinesis mapping as it stands.

Resolves #909

## Summary by CodeRabbit

* **New Features**
  * Added support for connecting Kinesis streams to Lambda event source mappings.
  * Added shard-based polling, batch delivery, ordering, retries, partial batch failures, and payload decoding.
  * Added support for `TRIM_HORIZON`, `LATEST`, and `AT_TIMESTAMP` starting positions.
  * Added validation for stream availability, permissions, configuration, and unsupported consumer ARNs.
  * Added safeguards against recursive stream-triggered invocations.

* **Documentation**
  * Expanded Kinesis and Lambda documentation with configuration, limitations, IAM, CloudFormation, and event payload details.

* **Bug Fixes**
  * Clarified DynamoDB starting-position behavior and standardized event-source mapping configuration.

### DIFF
--- a/docs/services/kinesis/README.md
+++ b/docs/services/kinesis/README.md
@@ -200,6 +200,13 @@ const { Records } = await kinesis.getRecords(
 console.log(Records.length);
 ```
 
+## Triggering a Lambda function
+
+A [Lambda event source mapping](../lambda/#triggering-a-function-from-a-kinesis-stream "Simulated Lambda Kinesis event source docs")
+polls a stream and invokes a function with the records it reads. Every shard is read by a processor
+of its own, as real Lambda reads one, and the function's execution role is what the polling is done
+as.
+
 ## Permissions
 
 Every operation goes through simulated IAM. The action is the `kinesis:` name of the operation, and
@@ -356,8 +363,6 @@ Anything else refuses on send with `SimSdkUnsupportedCommandError`.
   out, and no response carries an `EncryptionType`.
 - **Tags are kept and never listed.** A stream created with `Tags` holds them, readable through
   `findStream`. `AddTagsToStream`, `ListTagsForStream` and `RemoveTagsFromStream` are absent.
-- **Nothing triggers a Lambda function.** An event source mapping naming a Kinesis stream is refused
-  at creation.
 - **`AWS::Kinesis::Stream` does not deploy.** A CloudFormation template holding one records the
   resource as skipped.
 - **Kinesis Data Firehose is a separate service and is absent.** So is Kinesis Video Streams.
--- a/docs/services/lambda/README.md
+++ b/docs/services/lambda/README.md
@@ -1477,7 +1477,7 @@ the `Records` shape real Lambda uses.
 `StartingPosition` is required for a stream and is `TRIM_HORIZON` or `LATEST`. `TRIM_HORIZON` reads
 what the stream still holds, so changes made before the mapping existed are delivered too. `LATEST`
 reads only what the table changes from the moment the mapping starts reading. `AT_TIMESTAMP` is for
-a Kinesis stream and is refused by name.
+[a Kinesis stream](#triggering-a-function-from-a-kinesis-stream) and is refused by name here.
 
 `BatchSize` says how many records one invocation may be given, and defaults to 100, the number
 CDK's `DynamoEventSource` also asks for. The table and the function have to be in the same account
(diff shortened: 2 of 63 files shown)


hughgrigg:
@coderabbitai full review

Issue #922, opened 2026-08-22
Call DynamoDB, SNS and SQS from a Step Functions Task state


hughgrigg:
## Problem

Most Task states never touch Lambda. A workflow writes to DynamoDB, publishes to SNS, sends to SQS or puts an event on EventBridge, through the optimized integrations (`arn:aws:states:::dynamodb:putItem`) or the SDK integrations (`arn:aws:states:::aws-sdk:dynamodb:putItem`). CDK's `tasks` module emits them from `DynamoPutItem`, `SnsPublish`, `SqsSendMessage` and the rest.

With Lambda as the only Task target, a workflow doing its own writes cannot run, and the parts of it that do reach Lambda cannot be reached either.

## Desired behaviour

- An `arn:aws:states:::aws-sdk:<service>:<operation>` Resource calls that operation on the simulated service in the same simulation, with `Parameters` as the request and the response as the task result. The services already simulated are the ones it reaches.
- The optimized integrations for DynamoDB (`putItem`, `getItem`, `updateItem`, `deleteItem`), SNS (`publish`), SQS (`sendMessage`) and EventBridge (`putEvents`) work, with the request shapes those integrations define.
- The execution's role authorizes each call through simulated IAM, against the resource the Parameters name.
- A service error becomes a task error named the way ASL names it (`DynamoDb.ConditionalCheckFailedException`), so a `Retry` or `Catch` on it matches.
- A Resource naming an operation or a service this simulation has no answer for is refused when the state machine is created, naming the operation.
- `arn:aws:states:::states:startExecution` starts another simulated state machine, which is how a workflow calls a workflow.

## Example

```typescript
Record: {
  Type: "Task",
  Resource: "arn:aws:states:::dynamodb:putItem",
  Parameters: {
    TableName: "enrolments",
    Item: { id: { "S.$": "$.id" }, term: { S: "2026-autumn" } },
  },
  Next: "Announce",
}
```

## Acceptance criteria

- [ ] A `dynamodb:putItem` Task writes an item a `GetItem` afterwards reads back.
- [ ] An `aws-sdk:sqs:sendMessage` Task puts a message a receive afterwards takes off the queue.
- [ ] An `sns:publish` Task delivers to a subscribed simulated Lambda function.
- [ ] A conditional check failure becomes a task error a `Catch` matches by name.
- [ ] An execution whose role lacks the action fails the task with an access denied error.
- [ ] `states:startExecution` starts a second state machine and answers with its execution ARN.
[...]

Issue #921, opened 2026-08-22
Parallel and Map states in simulated Step Functions


hughgrigg:
## Problem

Fan-out is the other half of what a state machine is for. A `Map` state over an array of records, or a `Parallel` state running two independent branches, is where the workflow stops being a sequence and starts being worth drawing. Neither runs in simulated Step Functions, so a workflow built around one cannot be tested at all.

Both states also carry semantics worth getting wrong quietly. A failing branch cancels its siblings, `MaxConcurrency` bounds how many iterations are in flight, and the result is an array ordered by input position and not by completion.

## Desired behaviour

- A `Parallel` state runs each of its `Branches` from its own `StartAt`, and answers with an array of branch outputs in branch order.
- Every branch receives the same effective input, and each has its own states with no shared names across branches.
- A `Map` state runs its `ItemProcessor` once per element. `ItemsPath` selects the array, and `MaxConcurrency` bounds the iterations in flight (0 meaning unbounded).
- `ItemSelector` builds each iteration's input, with `$$.Map.Item.Value` and `$$.Map.Item.Index` available to it.
- A `Map` result is an array of iteration outputs in input order, whatever order they finished in.
- A failing branch or iteration fails the state with `States.BranchFailed`, and the siblings still in flight are abandoned.
- `Retry` and `Catch` on the `Parallel` or `Map` state itself work as they do on a `Task`.
- The context object carries `$$.Execution`, `$$.State` and `$$.Map`, which is what `ItemSelector` reads.
- The inspection accessor reports iterations and branches separately, so a test can assert that four records were processed.

## Example

```typescript
Enrol: {
  Type: "Map",
  ItemsPath: "$.students",
  MaxConcurrency: 2,
  ItemSelector: { "id.$": "$$.Map.Item.Value.id", "term.$": "$.term" },
  ItemProcessor: {
    StartAt: "Register",
    States: { Register: { Type: "Task", Resource: register.FunctionArn, End: true } },
  },
  Next: "Confirm",
}
```

## Acceptance criteria

- [ ] A `Parallel` state answers with one output per branch, in branch order.
- [ ] A `Map` state over four records runs its processor four times and answers with four outputs in input order.
- [ ] `ItemSelector` reading `$$.Map.Item.Index` reaches the iteration's task.
- [ ] `MaxConcurrency: 1` runs the iterations one at a time, provable by a task that records the clock.
- [ ] A failing iteration fails the `Map` state, and a `Catch` on that state routes the failure.
[...]

Issue #919, opened 2026-08-22
Deploy AWS::StepFunctions::StateMachine from CloudFormation and CDK


hughgrigg:
## Problem

Nobody writes a state machine by hand. It comes out of CDK, where `stepfunctions.StateMachine` and the `LambdaInvoke` tasks around it synthesize into one `AWS::StepFunctions::StateMachine` Resource. A test that has to call `CreateStateMachine` with a literal definition is testing a definition it wrote, and the one CDK produced goes untested.

Simulated CloudFormation skips the Resource type today and records it as unsupported (`src/service/cloudformation/resource/unsupported/sim-cfn-unsupported-resource.ts`). Everything else in the Stack still deploys, so a Stack with a workflow in it comes up with the workflow missing.

## Desired behaviour

- `AWS::StepFunctions::StateMachine` deploys into simulated Step Functions, from a template and from a synthesized CDK cloud assembly.
- `DefinitionString` is read after the template intrinsics resolve. CDK emits it as an `Fn::Join` over `Ref` and `Fn::GetAtt` values, and the joined string is the ASL the interpreter reads.
- `Definition` (the object form) is read as ASL directly.
- `StateMachineName`, `RoleArn`, `StateMachineType` and `Tags` are carried across. An unnamed state machine gets a generated name, following the other Resource factories.
- `Ref` answers with the state machine ARN, and `Fn::GetAtt` answers `Arn` and `Name`.
- Deleting the Stack deletes the state machine.
- `AWS::StepFunctions::StateMachineVersion` and `AWS::StepFunctions::StateMachineAlias` are refused as unsupported until something needs them.
- A definition the interpreter refuses drops that one state machine and records the reason on the Stack. The rest of the Stack deploys.

## Example

```typescript
const stack = new Stack(new App(), "Enrolment");
const workflow = new StateMachine(stack, "Workflow", {
  definitionBody: DefinitionBody.fromChainable(
    new LambdaInvoke(stack, "Check", { lambdaFunction: check, payloadResponseOnly: true }),
  ),
});

await simAws.cloudFormation().deployCdkStack(stack);
```

## Acceptance criteria

- [ ] A CDK Stack holding a `StateMachine` deploys, and the deployed state machine runs an execution to `SUCCEEDED`.
- [ ] `DefinitionString` written as an `Fn::Join` over a `Ref` to a Lambda function resolves to a Task the execution invokes.
[...]

PR #930, opened 2026-08-22
feat: run a simulated Step Functions state machine

hughgrigg:
A state machine is held in a template as data, and no unit test runs it. This adds `simAws.stepFunctions()`, the second pull request against #918, built on the data-flow layer that landed in #924.

## What runs

- **The service.** `CreateStateMachine`, `DescribeStateMachine`, `UpdateStateMachine`, `DeleteStateMachine`, `ListStateMachines`, `StartExecution` and `DescribeExecution`, through the accessor and through an intercepted `SFNClient`.
- **Three state types.** `Pass`, `Succeed` and `Fail`, with the five data-flow fields applying around each one. A definition using `Task`, `Choice`, `Wait`, `Parallel` or `Map` is refused when the state machine is created, naming the state and its type.
- **Definition checking at creation.** A `Next` naming a state that is not there, a state carrying neither `Next` nor `End`, a `Fail` carrying `InputPath`, a `Type` Amazon States Language does not define. Each is refused where real Step Functions refuses it, so an execution walks a definition it can rely on.
- **An inspection accessor** reporting the states an execution went through, and a state machine's executions.

## Three decisions worth a look

**`StartExecution` settles before it answers.** Real Step Functions answers before the execution has run, so a caller there sees `RUNNING` first. The interpreter here runs inline until it reaches something that has to wait on the clock, which for now is never. That spares every test a wait for work already done, and it is the shape `Wait` needs later: the walk stops at the `Wait`, leaves the execution `RUNNING`, and `advanceBy` moves it on. The alternative, scheduling the whole walk through `background.schedule`, puts it behind a `setTimeout` that no microtask yield reaches.

**A failing execution is recorded, never raised.** This follows the carve-out simulated time already makes for delivery, where EventBridge records a failure rather than throwing it into whichever caller happened to be advancing the clock. An execution failing is as often the thing under test as it is a fault.

**Every failure carries its Amazon States Language error name.** `States.ParameterPathFailure`, `States.Runtime` and the rest are read off the error and recorded on the execution. #920's `Retry` and `Catch` match on exactly that, so the wiring is in place before it is needed.

## Note on the file layout

There are more small files here than the shape of the code suggests. FTA scores density rather than length, so extracting code from a file can raise its score: `sim-state-machine-writes.ts` went **up** from 51.7 to 53.6 on one split before coming down when `create` moved out with its token mass. The command classes ended up split along read/write lines, which is the Kinesis layout anyway.

## Still to come on #918

`Choice` and `Wait` next, then `Task` with Lambda and IAM authorization. #919, #920, #921 and #922 all sit on top.

## Summary by CodeRabbit

* **New Features**
  * Added simulated AWS Step Functions support for creating, updating, listing, describing, and deleting state machines.
  * Added execution start and status inspection, including outputs, failures, visited states, and execution history.
  * Supports `Pass`, `Succeed`, and `Fail` states with data-flow handling, Standard and Express machine types, and AWS SDK command interception.
* **Bug Fixes**
  * Preserved falsy values when inserting data into arrays.
* **Documentation**
  * Added Step Functions documentation and runnable usage examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ npm i -D @kensio/yulin
 - [SNS](./docs/services/sns "Simulated SNS docs")
 - [SQS](./docs/services/sqs "Simulated SQS docs")
 - [SSM Parameter Store](./docs/services/ssm "Simulated SSM Parameter Store docs")
+- [Step Functions](./docs/services/stepfunctions "Simulated Step Functions docs")
 - [STS](./docs/services/sts "Simulated STS docs")
 - [WAFv2](./docs/services/wafv2 "Simulated WAFv2 docs")
 
--- a/docs/services/stepfunctions/README.md
+++ b/docs/services/stepfunctions/README.md
@@ -4,7 +4,7 @@ Simulated Step Functions interprets Amazon States Language and runs a state mach
 process as the code under test. A workflow held in a template as data becomes something a test can
 run and assert on.
 
-Step Functions specific types are imported from the `@kensio/yulin/stepfunctions` subpath.
+Types for simulated Step Functions are imported from the `@kensio/yulin/stepfunctions` subpath.
 
 ## What runs today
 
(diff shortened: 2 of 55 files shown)


hughgrigg:
@coderabbitai full review

PR #934, opened 2026-08-22
feat: deliver Firehose records into an S3 bucket

hughgrigg:
A delivery stream takes records through PutRecord and PutRecordBatch, buffers them until the buffer
passes its size or its interval, and writes them into a simulated S3 Bucket as its RoleARN.
Advancing the simulated clock past the interval is what delivers a buffer, and the object key
follows the real Firehose format of prefix, UTC date path, delivery stream name, version, timestamp
and a random string. S3 is the only destination and DirectPut the only source, and the rest are
refused by name at CreateDeliveryStream. A buffer the delivery Role could not write is kept on
simAws.firehose().getDeliveryFailures(), since real Firehose answered the put minutes earlier and
tells the producer nothing.


hughgrigg:
@coderabbitai rate limit

PR #935, opened 2026-08-22
feat: deploy AWS::StepFunctions::StateMachine

hughgrigg:
A template declaring AWS::StepFunctions::StateMachine now creates a state machine, and a CDK app synthesizing one deploys and runs an execution against it. The definition is read once the template intrinsics have resolved. The Fn::Join CDK writes DefinitionString as arrives with real ARNs in it, the object Definition form is read as Amazon States Language directly, and DefinitionSubstitutions are applied before either is read. StateMachineName, RoleArn and StateMachineType go through the ordinary CreateStateMachine command. A template and an SDK caller reach the same state machine, and a definition Step Functions will not take is refused in the words that command refuses it in. An unnamed state machine is named after the stack and the logical ID. Ref answers the ARN, Fn::GetAtt answers Arn and Name, and deleting the stack deletes the state machine. A definition holding a state type the interpreter does not run drops that one state machine and records the reason on the stack, and the rest of the stack still deploys. StateMachineVersion and StateMachineAlias are refused as unsupported, and Tags, LoggingConfiguration, TracingConfiguration and EncryptionConfiguration are recorded against the Resource as unsimulated.

Three parts of #919 are outstanding.

- **A Task the execution invokes.** The Fn::Join over a Ref is covered, with a queue ARN joined into a Pass state, and an execution reads the resolved ARN out of its output. Joining a Lambda ARN into a Task state needs the Task slice of #918, which has not landed. The rest of that acceptance criterion is tested.
- **DefinitionBody.fromFile.** It synthesizes to DefinitionS3Location, which #919 puts out of scope. That state machine is dropped and the reason recorded rather than deployed without the definition it was meant to run. The object Definition form is covered.
- **Tags.** Simulated Step Functions holds no tags on a state machine and has no ListTagsForResource, so there is nowhere to carry them to. The property is recorded on stack.ignoredProperties and the state machine is created without them.

One note on placement. The factory sits under src/service/cloudformation/resource/cfn/stepfunctions/ rather than src/service/stepfunctions/cfn/ where the other services keep theirs, to stay clear of the interpreter work still in flight on #918. Moving it once that lands is a rename and a factory-map edit.

## Summary by CodeRabbit

* **New Features**
  * Added support for deploying and simulating Step Functions state machines through CloudFormation.
  * Supports state machine definitions, substitutions, generated names, references, attributes, execution, and deletion.
  * Added an executable deployment and execution example.

* **Documentation**
  * Documented CloudFormation and CDK support, supported properties, limitations, versions, aliases, and S3 definitions.

* **Bug Fixes**
  * Added validation and clear handling for invalid or unsupported state machine configurations.

* **Tests**
  * Added comprehensive integration coverage for deployment, execution, references, substitutions, errors, skipped resources, and deletion.

### DIFF
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -3263,6 +3263,7 @@ The resource types it creates are:
 - `AWS::SecretsManager::Secret`
 - `AWS::SQS::Queue`
 - `AWS::SSM::Parameter`
+- `AWS::StepFunctions::StateMachine`
 - selected CDK custom resources: `Custom::CDKBucketDeployment` and `Custom::S3BucketNotifications`
 
 Each service's own docs describe what its resource types support.
--- a/docs/services/stepfunctions/README.md
+++ b/docs/services/stepfunctions/README.md
@@ -305,8 +305,10 @@ console.log(started.executionArn);
 ## Deploying one from CloudFormation
 
 A template declaring `AWS::StepFunctions::StateMachine` creates a state machine. It goes through the
-ordinary `CreateStateMachine` command. A template and an SDK caller reach the same state machine, and
-a definition this simulator will not run is refused in the same words either way.
+ordinary `CreateStateMachine` command, and a template and an SDK caller reach the same state machine.
+A definition Amazon States Language itself refuses fails the Resource, in the words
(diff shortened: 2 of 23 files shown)


hughgrigg:
@coderabbitai full review

PR #980, opened 2026-08-24
feat: set up CloudWatch Logs delivery

### LINKED ISSUE #979: Deploy AWS::Logs::DeliverySource, DeliveryDestination and Delivery from a template
## Problem

CloudFront standard logging v2 is configured through the CloudWatch Logs API. The distribution itself carries no property for it. A stack that turns it on is made of three resource types and nothing else. `AWS::Logs::DeliverySource` names the distribution, `AWS::Logs::DeliveryDestination` names the bucket, and `AWS::Logs::Delivery` joins the two.

Simulated CloudFormation handles none of the three. `SimLogsCfnResourceFactory` covers `AWS::Logs::LogGroup` and says so in its own comment. Each of the three delivery resources is skipped, marked `CREATE_COMPLETE` and recorded on `SimCfnStack.skippedResources`. That skipping is deliberate and is argued in #273.

So a test of a CloudFront logging construct deploys a stack where every resource was skipped, and the assertions have nothing to stand on. KensioSoftware/rainlytics#8 is the downstream consumer, and its construct is exactly these three resources.

## Example

```typescript
new CfnResource(stack, "Source", {
  type: "AWS::Logs::DeliverySource",
  properties: { Name: "site-access-logs", ResourceArn: distribution.distributionArn, LogType: "ACCESS_LOGS" },
});

const deployed = await simAws.cloudFormation().deployCdkOut({ path: "cdk.out" });
// deployed.stacks[0].skippedResources includes Source
```

## Desired behaviour

The three resource types deploy like any other supported type, and the same three are reachable through the SDK. A construct test asserts on the wiring through `DescribeDeliverySources`, `DescribeDeliveryDestinations` and `DescribeDeliveries`.

## Acceptance criteria

- [ ] A stack declaring all three creates them, `Ref` and `Fn::GetAtt Arn` resolve on each, none appears in `skippedResources`, and deleting the stack deletes all three.
- [ ] The three are created, read and deleted through the SDK with `PutDeliverySource`, `PutDeliveryDestination` and `CreateDelivery`, the three `Describe` commands and the three `Delete` commands.
- [ ] A CloudFront delivery source is refused outside us-east-1, whatever region the destination bucket sits in, matching AWS.
- [ ] A second delivery source for a `ResourceArn` that already has one fails with the AWS message about the resource id already being used.
- [ ] `OutputFormat` is fixed once the delivery destination exists. A template change to it replaces the resource, and an SDK update that changes it is refused.
- [ ] An S3 delivery destination accepts Hive-compatible paths and a suffix path with the `{DistributionId}`, `{yyyy}`, `{MM}`, `{dd}`, `{HH}` and `{accountid}` variables, and both read back.

## Out of scope

Writing actual CloudFront access log files into the simulated bucket. Delivery sources for services other than CloudFront.

## Implementation notes

Output format is one of JSON, Plain, w3c, Raw or Parquet, and Parquet is available only for an S3 destination.

The exact error name AWS returns for a delivery source created outside us-east-1 needs checking against the API before the refusal is written.

hughgrigg:
Simulated CloudWatch Logs now holds delivery sources, delivery destinations and deliveries, reachable through the nine SDK commands that manage them and through `AWS::Logs::DeliverySource`, `AWS::Logs::DeliveryDestination` and `AWS::Logs::Delivery` in a template. A distribution carries no logging property of its own, and a construct that turns CloudFront standard logging v2 on is made of those three resource types. A test of one used to deploy a stack where every resource was recorded as a skip, which proved nothing about the wiring. Four AWS rules are modelled alongside the resources, because each of them is a deploy that looks fine until it runs: one delivery source per distribution, CloudFront delivery only from us-east-1, an output format that is fixed once the destination exists, and Parquet on S3 only. A suffix path naming a partition variable delivery would never substitute is refused too.

Resolves https://github.com/KensioSoftware/yulin/issues/979

## Notes for the reviewer

This is a large one, around 4,400 lines. Folding the SDK commands in alongside the CloudFormation work roughly doubled it. The seam if you would rather take it in two passes is the `cfn/` directory plus the value adapters and the two CloudFormation test files, split off from the model and SDK work underneath, which stands alone.

Two things I could not settle against real AWS. The error name for a CloudFront delivery source created outside us-east-1 is `ValidationException` here, which is what the delivery API's model carries, with a message of Yulin's own wording. The one source per distribution message is the verbatim AWS text. And deleting a delivery source while a delivery still uses it is permitted here. Real CloudWatch Logs has its own rule about the order, guessing either way would be a divergence, and it is recorded under Limitations instead.

The delivery operations report a bad input as `ValidationException` and `ConflictException` rather than the `InvalidParameterException` the log group operations use. That is real: they were added to CloudWatch Logs long after the rest, and carry the error shape AWS gives newer APIs. `SimLogsPage` is shared with the older listings, so a bad `limit` on a `Describe` still raises `InvalidParameterException`, which is the one place the split leaks.

The CloudFormation creators go through the ordinary commands rather than writing to the stores, unlike `SimCfnLogGroupCreator`. There is no equivalent here of a log group a Lambda function already made for itself, so a template hits every refusal an SDK caller would, down to the region rule.

Three files went over the FTA threshold on the first pass. `simLogsCreatedDelivery`, `simCfnDeliveryInput` and `SimLogsCfnResourceDeleter` exist to carry tokens out of the files that were breaching, not because they were the natural shape to start with.

`SimLogs` gained `SimLogsDeliveryOperations` and `SimLogsDeliveryInspection` as bases, following `SimLambda`. The facade was at the line limit with nine delegating methods still to add. That is what turned `#commands` into a `protected commands`.

The branch is named `claude/confident-jang-379625` rather than `feat/hg/...`. Say the word and I will move it.

## Summary by CodeRabbit

* **New Features**
  * Added simulated CloudWatch Logs delivery sources, destinations, and deliveries.
  * Added support for creating, listing, describing, and deleting delivery resources through SDK and CloudFormation workflows.
  * Added CloudFront-to-S3 delivery configuration, including output formats and Hive-compatible partition paths.
  * Added validation, authorization, ARN generation, pagination, and lifecycle handling.
* **Documentation**
  * Added API, CloudFormation, limitations, and configuration examples.
* **Tests**
  * Added comprehensive coverage for delivery workflows, validation, authorization, routing, and resource lifecycle behavior.

### DIFF
--- a/docs/services/logs/README.md
+++ b/docs/services/logs/README.md
@@ -376,6 +376,11 @@ rules from real AWS are modelled here, each of them a deploy that looks fine unt
 - **One delivery source per resource.** A second source over a distribution that already has one
   fails with `ConflictException`, carrying the message an account gives, "This ResourceId has
   already been used in another Delivery Source in this account".
+- **A delivery holds both its ends.** Deleting the source or the destination while a delivery joins
+  them fails with `ConflictException`. The delivery goes first, and CloudFormation orders that
+  itself when a stack is deleted.
+- **CloudFront delivers `ACCESS_LOGS` and nothing else.** Any other `logType` over a distribution is
+  refused.
 - **CloudFront delivery is set up from `us-east-1`**, whatever region the destination bucket is in.
   A CloudFront delivery source put from anywhere else is refused.
 - **Output format is fixed once the destination exists.** A `PutDeliveryDestination` that would
@@ -385,7 +390,7 @@ rules from real AWS are modelled here, each of them a deploy that looks fine unt
 The suffix path decides the key each log file lands under. `{DistributionId}`, `{distributionid}`,
 `{yyyy}`, `{MM}`, `{dd}`, `{HH}` and `{accountid}` are substituted, and a variable outside that set
 is refused. Delivery would write the text out literally, and the bucket would look partitioned when
-it was not.
+it was not. A path over the 256 characters CloudWatch Logs takes is refused too.
 
 ### Declaring delivery in a template
 
@@ -422,9 +427,10 @@ The template carries the S3 layout as two flat properties, and the API takes the
 `Ref` resolves to the name on the source and the destination, and to the delivery ID on the
 delivery. CloudWatch Logs issues that ID. A template cannot predict it.
 
-`Fn::GetAtt` gives `Arn` on all three. The source also publishes `Service` and `ResourceArns`, the
-destination `DeliveryDestinationType`, and the delivery `DeliveryId`, `DeliverySourceName`,
-`DeliveryDestinationArn` and `DeliveryDestinationType`.
+`Fn::GetAtt` gives `Arn` on all three. The source also publishes `Service` and `ResourceArns`, and
+the delivery publishes `DeliveryId` and `DeliveryDestinationType`. The destination publishes nothing
+else, and `DeliveryDestinationType` is a property of it rather than an attribute, so read that one
+off the delivery. Anything outside this set is refused here, as CloudFormation refuses it.
 
 `Tags` and `DeliveryDestinationPolicy` are recorded as ignored properties, and the stack still
 deploys.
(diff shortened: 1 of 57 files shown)


hughgrigg:
Replaced by #981, which is the same commits from a branch following the naming convention.

PR #1065, opened 2026-08-27
feat: deny an action with a service control policy

### LINKED ISSUE #1059: Deny an action with a service control policy on a simulated Account
## Problem

An organization's service control policies decide what a member account may do. A stack asking for something the SCPs block fails at deployment time, in a real account, once the code is already written. Working out which SCP is in the way costs a deploy and a wait, once for every permission the account has drifted away from granting.

Yulin evaluates IAM policy documents in some depth. That evaluation stops at the account boundary, leaving the surrounding organization's rules outside anything a test can express.

## Current behaviour

`SimIamPolicyDecision` evaluates identity policies and resource policies. Its own doc comment lists service control policies among what it leaves out. `SimIamAuthZPolicySourceType` carries `identity-inline`, `identity-managed`, `resource`, `trust`, `permissions-boundary` and `session`, and an organization source would be new.

A resource created through sim CloudFormation reaches the owning service's command handler, and that handler authorizes the same way an SDK call does. `SimCfnS3BucketCreator` calls `simS3.createBucket`, which runs `CreateBucketCommandHandler` and its `s3:CreateBucket` check. The CloudFormation path carries no caller. Sim IAM falls back to the Account root, and `SimIamAuthZCallerContextBuilder` hands root a synthetic `Allow *` inline policy called `SimAwsAccountRootAccess`, which is why every CloudFormation-created resource is allowed. In AWS an SCP applies to a member account's root principal. That makes this request exactly the one an SCP would catch.

## Desired behaviour

A test attaches an SCP document to a simulated Account, and every authorization decision for a principal in that Account passes through it first.

An SCP filters and never grants:

- A matching `Deny` in an attached SCP denies the request.
- An attached SCP set with no matching `Allow` denies the request, whatever an identity policy says.
- An Account with no SCP attached decides as it does today.

The filter covers the Account root along with the Account's Roles and Users. A CloudFormation deployment and an intercepted SDK call are both subject to it. A denial fails the resource, which `SimCfnResourceCreateOperation` already marks `CREATE_FAILED` before rejecting the deployment.

## Example

A sketch of the shape, with the accessor name still open:

```typescript
simAws.organizations().attachServiceControlPolicy("123456789012", {
  Version: "2012-10-17",
  Statement: [{ Effect: "Deny", Action: "s3:CreateBucket", Resource: "*" }],
});

await expect(
  simAws.cloudFormation().deployTemplateFile({
    templatePath: "cdk.out/SiteStack.template.json",
  }),
).rejects.toThrow(/service control policy/u);
```

## Acceptance criteria

- [ ] A simulated Account can be given one or more SCP documents through a public accessor.
- [ ] `SimIamPolicyDecision` reports the SCP verdict apart from the identity and resource verdicts, and exposes the matching SCP statements for a test to assert on.
- [ ] An action an SCP denies raises `SimIamAccessDenied` with a message naming the service control policy, as the AWS message does.
- [ ] An SCP set with no matching `Allow` denies an action an identity policy allows.
- [ ] Deploying a template whose resource the SCP denies leaves that resource `CREATE_FAILED` and rejects the deployment.
- [ ] An intercepted SDK call under `SimAws.runAs` is denied by the SCP attached to the caller's Account.
- [ ] Documentation covers the realism ceiling, including which condition keys are matched.

## Out of scope

Organizational units and policy inheritance, covered by #1060. Resource control policies, declarative policies, backup policies and tag policies. Creating an organization from `AWS::Organizations::*` CloudFormation resources, covered by #1061. The Organizations SDK client (`CreatePolicy`, `AttachPolicy`, `ListAccounts` and the rest). The service-linked role exemption and the management account exemption, both of which need an organization structure behind them.

## Implementation notes

An SCP `Allow` grants no permission by itself. The evaluation belongs ahead of the two existing policy sides as a gate, with `SimIamAllowRequirement` going on combining identity and resource Allows as it does now.

AWS attaches `FullAWSAccess` (`Allow *` on `*`) to every node by default, which is why turning SCPs on leaves an account's permissions as they were until somebody writes a policy. A simulated Account with SCPs needs the same default, or a test attaching one `Deny` statement locks the whole Account out.

hughgrigg:
Simulated Organizations holds the service control policies attached to each simulated Account, reached as `simAws.organizations()`. Sim IAM evaluates them as a gate in front of the identity and resource sides, so an SCP filters what an Account may do and grants nothing.

Resolves https://github.com/KensioSoftware/yulin/issues/1059

The gate covers the Account root along with the Account's Roles and Users, which is what makes it catch a deployment. Sim CloudFormation creates each resource through the owning service's command handler, and that handler already authorizes, so no new CloudFormation hook was needed. A denial leaves the resource `CREATE_FAILED` and rejects the deployment, and the `AccessDenied` message names the service control policy the way AWS words it. An intercepted SDK client is filtered the same way. `FullAWSAccess` is attached by default as it is in AWS, and `detachFullAwsAccess` turns an Account's policies into an allow list.

Two refactors ride along, both needed to get `sim-iam-decision.ts` back under the FTA threshold once the gate was added. The identity and resource evaluation moved to `SimIamPolicyEvaluation`, mirroring `SimIamScpGate`, and `SimIamPolicyDecisionValue` moved to its own module and is re-exported.

The wide part of the diff is one line at each of the 60 `new SimIamAccessDenied` sites, passing `decision.denialReason` through so the SCP wording reaches the error wherever a service throws one.

- [x] Rebased off latest main

## Summary by CodeRabbit

* **New Features**
  * Added simulated AWS Organizations service control policies (SCPs), including policy attachment, removal, default full access, allow-list behavior, and account-level evaluation.
  * SCP decisions now support explicit and implicit denials, policy details, and CloudFormation enforcement.
* **Bug Fixes**
  * Access-denied errors across supported services now include clearer IAM denial reasons.
* **Documentation**
  * Added Organizations service documentation, IAM limitations, usage guidance, and practical SCP examples.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,6 +26,7 @@ behaviour and includes example code that can be copied into tests or local devel
 - [KMS](https://yulinsim.dev/services/kms/ "Simulated KMS usage docs")
 - [Lambda](https://yulinsim.dev/services/lambda/ "Simulated Lambda usage docs")
 - [CloudWatch Logs](https://yulinsim.dev/services/logs/ "Simulated CloudWatch Logs usage docs")
+- [Organizations](https://yulinsim.dev/services/organizations/ "Simulated Organizations service control policies usage docs")
 - [Personalize](https://yulinsim.dev/services/personalize/ "Simulated Amazon Personalize usage docs")
 - [Rekognition](https://yulinsim.dev/services/rekognition/ "Simulated Rekognition usage docs")
 - [Route53](https://yulinsim.dev/services/route53/ "Simulated Route53 usage docs")
--- a/docs/services/iam/README.md
+++ b/docs/services/iam/README.md
@@ -1090,7 +1090,8 @@ Sim IAM models the policy behaviour that multi-service tests most commonly need.
 
 - Groups are absent. An `AWS::IAM::User` naming one fails, and an `AWS::IAM::Policy` naming one
   fails too
-- Permissions boundaries, session policies, and service control policies are not evaluated
+- Permissions boundaries and session policies are not evaluated. Service control policies are, and
+  are attached through [simulated Organizations](https://yulinsim.dev/services/organizations/ "Simulated Organizations service control policies usage docs")
 - Managed Policies have a single version, and the policy version commands are absent
 - `DeleteUserPolicy` and `DetachUserPolicy` are absent, along with `DeleteAccessKey` and
   `DeleteLoginProfile`. `DeleteUserCommand` refuses a User that still holds a policy, and
--- a/docs/services/organizations/README.md
+++ b/docs/services/organizations/README.md
@@ -1,0 +1,186 @@
+# Simulated Organizations
+
+Yulin simulates AWS Organizations service control policies. A test can find out that the
+organization around an account forbids something before a deployment does.
+
+A service control policy filters what an account's principals may do and grants nothing. Sim IAM
+evaluates the ones attached to an account ahead of that account's identity and resource policies,
+which means an SCP applies to a CloudFormation deployment, an intercepted SDK client, and a direct
+service call alike.
+
+## Attach a service control policy to an Account
+
+An organization spans accounts. It belongs to the whole simulated environment and is reached as
+`simAws.organizations()`, not from an account scope.
+
+```typescript sim-organizations-attach-scp
+/**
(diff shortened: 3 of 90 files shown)


hughgrigg:
@coderabbitai full review

PR #1066, opened 2026-08-27
feat: inherit a service control policy from an OU

### LINKED ISSUE #1060: Apply an organizational unit's service control policy to the Accounts below it
## Problem

Teams rarely attach an SCP to one account. They attach it to an organizational unit, and every account under that OU inherits it. A simulated Account that can only hold its own SCPs models the wrong shape, and a test written against it has to flatten the org chart by hand and keep that copy in step with the real one.

The effective set for an account is also what makes a denial hard to read in AWS. A policy three levels up is the one that blocks the deploy, and the account itself carries nothing.

## Desired behaviour

A simulated organization has a root, organizational units nested under it, and Accounts placed in those units. An SCP attached to any node applies to every Account beneath it. The effective set for an Account is every SCP on the path from the root down to that Account, and each one has to allow the action for the action to go ahead.

The management account is exempt from SCP evaluation, as it is in AWS.

A test can read the effective set for an Account. A denial then traces back to the node its policy hangs on.

## Example

A sketch of the shape, with the accessor names still open:

```typescript
const organizations = simAws.organizations();
const workloads = organizations.createOrganizationalUnit({ name: "Workloads" });

organizations.moveAccount("123456789012", workloads);
organizations.attachServiceControlPolicy(workloads, denyBucketCreation);

console.log(organizations.effectivePoliciesFor("123456789012"));
```

## Acceptance criteria

- [ ] Organizational units can be created, nested, and given Accounts.
- [ ] An SCP attached to the root or to an OU denies an action for every Account beneath it.
- [ ] An Account inherits the SCPs of each node on its root-to-Account path, and each of those has to allow an action for the Account to be allowed it.
- [ ] An `Allow` in one node's SCP leaves a `Deny` at another node standing.
- [ ] The Account designated as the management account skips SCP evaluation.
- [ ] The effective SCP set for an Account is readable for diagnostics.
- [ ] Documentation covers the organization shape, inheritance, and the management account exemption.

## Out of scope

Everything #1059 leaves out. Moving an Account between OUs mid-test, unless it falls out for free. Organization-level features other than SCPs. Delegated administrator accounts. The `aws:PrincipalOrgID` and `aws:PrincipalOrgPaths` condition keys, which are worth their own issue once an org structure exists to populate them.

## Implementation notes

Depends on #1059, which decides where an SCP is stored and how it is evaluated. This one puts a tree in front of that store and changes what the evaluation reads.

The organization spans Accounts. It belongs on `SimAws` alongside `SimIamRegistry`, which already indexes IAM state per Account for cross-Account decisions.

hughgrigg:
A simulated organization now has a root and organizational units under it, and an Account inherits every service control policy on the path down to it.

Resolves https://github.com/KensioSoftware/yulin/issues/1060

The behaviour change worth reviewing is that each level is asked separately. #1059 read an Account's policies as one flat list, so a single matching `Allow` anywhere opened the gate. AWS requires an `Allow` at every level between the root and the Account, so a root allowing `s3:*` and a unit allowing `dynamodb:*` leave an Account under both able to do neither. `SimIamScpGate` now evaluates level by level, and a denial reports `unallowedLevels`, naming the nodes that allowed nothing matching.

`setManagementAccount` names the Account AWS exempts from every SCP. An Account that only ever had a policy attached to it directly is still filtered by that policy alone, so everything #1059 shipped keeps working unchanged.

`SimOrganizations` is split across `SimOrganizationsStructure` (root, units, Account placement) and the policy half, following the `SimS3 extends SimS3Operations` pattern. That split was forced by the FTA threshold, and extracting to a collaborator made the score worse rather than better, which is the usual density trap.

- [x] Rebased off latest main

## Summary by CodeRabbit

- **New Features**
  - Added hierarchical Organizations support with roots, organizational units, account placement, and policy inheritance.
  - Service control policies can now be attached and evaluated at each organizational level.
  - Added management-account exemptions and support for accounts outside an organization.
  - Exposed organization structure, node identifiers, effective policies, and policy evaluation results.
  - Added examples demonstrating inheritance, multi-level authorization, and management-account behavior.

- **Documentation**
  - Expanded Organizations guidance with evaluation rules, examples, terminology, and updated limitations.

### DIFF
--- a/docs/services/organizations/README.md
+++ b/docs/services/organizations/README.md
@@ -276,9 +276,10 @@ account, in the order they were evaluated, including `FullAWSAccess` where it is
 `serviceControlPolicySetFor(accountId).levels` keeps the policies grouped by the node they hang on,
 root first. That grouping is what sim IAM evaluates.
 
-The flattened list is empty in two cases that behave differently. An account that was never named
-sits outside the organization and stays unrestricted. An account left holding no policy is denied
-everything.
+The flattened list is empty in three cases that behave differently. An account that was never named
+sits outside the organization and stays unrestricted. The management account is exempt and equally
+unrestricted. An account left holding no policy is denied everything. `applies` separates the last
+of those from the other two.
 `serviceControlPolicySetFor(accountId).applies` tells the two apart, and so does
 `decision.serviceControlPolicy.isApplied`.
 
@@ -292,7 +293,8 @@ Simulated Organizations supports:
 - `setManagementAccount`, exempting the Account AWS exempts
 - `attachServiceControlPolicy`, attaching a policy document to the root, a unit, or an Account
 - `detachFullAwsAccess`, turning a node's policies into an allow list
-- `detachServiceControlPolicies`, taking every policy back off a node
+- `detachServiceControlPolicies`, taking every policy back off one node and leaving the rest alone
+- `removeAccount`, taking an Account out of the organization
 - `serviceControlPoliciesFor`, reading the policies in force for an Account
 - `serviceControlPolicySetFor`, reading those policies along with whether any apply
 - The AWS-managed `FullAWSAccess` policy, attached by default as it is in AWS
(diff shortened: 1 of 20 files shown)


hughgrigg:
@coderabbitai full review

PR #1067, opened 2026-08-27
feat: deploy an organization from CloudFormation

### LINKED ISSUE #1061: Deploy an organization and its service control policies from CloudFormation
## Problem

An organization managed in CDK or CloudFormation is declared as `AWS::Organizations::*` resources. A test that re-declares the same SCPs through a Yulin accessor keeps a second copy of the org chart, and the two drift apart at the moment the real one changes.

## Current behaviour

`simCfnServiceFactories` maps each `AWS::<Service>::*` name to the simulated service that creates it, and it has no `Organizations` entry. Every Organizations resource in a template is therefore reported as unsupported, recorded against the Stack by `SimCfnResourceCreateOperation.recordUncreated`, and skipped while the rest of the deployment carries on.

## Desired behaviour

Deploying a template carrying `AWS::Organizations::Organization`, `AWS::Organizations::OrganizationalUnit`, `AWS::Organizations::Account` and `AWS::Organizations::Policy` builds the simulated organization those resources describe. A `Policy` of type `SERVICE_CONTROL_POLICY` attaches to each node its `TargetIds` names, and the evaluation built in #1059 and #1060 then applies to the Accounts underneath.

`Ref` and `GetAtt` resolve as CloudFormation does, so a `Policy` can target an `OrganizationalUnit` declared in the same template.

A policy type Yulin models no evaluation for, such as `TAG_POLICY` or `BACKUP_POLICY`, is skipped with an unsupported-Resource message and leaves the Stack deploying.

## Example

```yaml
Resources:
  Workloads:
    Type: AWS::Organizations::OrganizationalUnit
    Properties:
      Name: Workloads
      ParentId: !GetAtt Organization.RootId

  DenyBucketCreation:
    Type: AWS::Organizations::Policy
    Properties:
      Name: DenyBucketCreation
      Type: SERVICE_CONTROL_POLICY
      TargetIds:
        - !Ref Workloads
      Content:
        Version: "2012-10-17"
        Statement:
          - Effect: Deny
            Action: s3:CreateBucket
            Resource: "*"
```

## Acceptance criteria

- [ ] `simCfnServiceFactories` routes `AWS::Organizations::*` to a simulated Organizations Resource factory.
- [ ] `AWS::Organizations::Organization`, `::OrganizationalUnit`, `::Account` and `::Policy` each create the simulated equivalent.
- [ ] A `SERVICE_CONTROL_POLICY` attaches to every node in `TargetIds`, and an Account beneath one of those nodes is denied what the policy denies.
- [ ] `Ref` and `GetAtt` return the values the AWS resource documentation gives for each of the four types.
- [ ] A policy type other than `SERVICE_CONTROL_POLICY` is recorded as unsupported and the Stack deploys on.
- [ ] Stack teardown removes what the template created.
- [ ] Documentation lists the four Resource types and the properties read from each.

## Out of scope

`AWS::Organizations::ResourcePolicy`. Evaluating tag, backup and AI services opt-out policies. Account closure and account invitation. Anything #1059 and #1060 left out.

## Implementation notes

Depends on #1059 and #1060, which build the store and the tree this reads into.

`SimCfnScopedServiceFactory` hands a factory the Account and Region scope a Stack deploys into, and an organization sits above that scope. The factory needs a route from the scope up to `SimAws`, the way `SimIamRegistry` is reached for cross-Account decisions.

hughgrigg:
`AWS::Organizations::Organization`, `::OrganizationalUnit`, `::Account` and `::Policy` now build the simulated organization a template describes, so a stack already managing the org chart is the one a test deploys.

Resolves https://github.com/KensioSoftware/yulin/issues/1061

The issue's implementation note assumed the factory would need a route from the Account and Region scope up to `SimAws`. It does not: `SimCloudFormationResourceCreateContext` already carries `simAws`, so the factory is stateless and registers like the CloudFormation one does, with no scope accessor added.

`Content` is read inline or as JSON text, and `TargetIds` takes a list or a single value, since real templates use both. `Ref` and `Fn::GetAtt` answer what the AWS resource reference gives for each type, and an attribute it does not list is refused by name rather than resolving to nothing. A `Policy` of any type but `SERVICE_CONTROL_POLICY` is recorded as a skipped Resource and the stack deploys on, because nothing here evaluates a tag or backup policy and storing one would look like it did something.

Teardown takes the policies back off their nodes, removes the units, and takes any Account the stack created back out of the organization. A unit removed while it still holds something hands what it holds to its parent, so every Account keeps a path to the root.

Two supporting changes reach earlier work. A root or unit id is now a valid policy target alongside a handle, which is what a template carries after a `Ref` resolves, and `removeOrganizationalUnit` is new. Both are additive.

- [x] Rebased off latest main

## Summary by CodeRabbit

* **New Features**
  * Added CloudFormation support for simulated Organizations resources, including organizations, organizational units, accounts, and service control policies.
  * Added support for policy attachment, enforcement, `Ref`, and `Fn::GetAtt` outputs.
  * Added organizational unit removal with account and child-unit reparenting.
  * Added end-to-end deployment documentation and examples.

* **Bug Fixes**
  * CloudFormation teardown now removes accounts, organizational units, and policy enforcement cleanly.

* **Tests**
  * Added comprehensive coverage for validation, deployment, policy enforcement, outputs, and cleanup.

### DIFF
--- a/docs/services/organizations/README.md
+++ b/docs/services/organizations/README.md
@@ -311,7 +311,9 @@ console.log(decision.value); // "ExplicitDeny"
 ```
 
 `Content` takes the policy document inline or as JSON text. `TargetIds` takes a list or a single
-value, and each entry names the root, a unit, or an Account.
+value, and each entry names the root, a unit, or an Account. A policy reaches every target it names
+or none of them. A target this organization has never heard of fails the resource before anything is
+attached.
 
 A simulated environment has one organization from the start, so
 `AWS::Organizations::Organization` records the one already there rather than making another. It is
@@ -329,9 +331,10 @@ These are the properties read from each resource:
 | `AWS::Organizations::Account`            | `AccountName`, `Email`, `ParentIds`    | `Tags`, `RoleName`    |
 | `AWS::Organizations::Policy`             | `Name`, `Type`, `Content`, `TargetIds` | `Tags`, `Description` |
 
-Tearing the stack down takes the policies off the nodes they were attached to, removes the units,
-and takes any Account the stack created back out of the organization. A unit that still holds
-something when it goes hands what it holds to its parent, so every Account keeps a path to the root.
+Tearing the stack down takes its own policies off the nodes they were attached to, removes the
+units, and takes any Account the stack created back out of the organization. A node holding
+policies from more than one stack keeps the others. A unit that still holds something when it goes
+hands what it holds to its parent, so every Account keeps a path to the root.
 
 ## Reading a denial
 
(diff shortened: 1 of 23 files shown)


hughgrigg:
@coderabbitai full review

PR #1171, opened 2026-08-30
feat: read metrics out of an Embedded Metric Format log line

### LINKED ISSUE #1168: Publish a metric from an Embedded Metric Format log line
## Problem

AWS Lambda Powertools' `Metrics` publishes by writing an Embedded Metric Format document to stdout. CloudWatch reads the metric out of the log group, with no metric filter and no `PutMetricData` call anywhere. That is how a Powertools handler counts anything, and it is what a CDK `NodejsFunction` bundling Powertools does by default.

Yulin captures the document already. `src/service/lambda/function/code/vm/sim-lambda-vm-powertools.loc.test.ts` runs a real Powertools handler and reads the `_aws` envelope back out of the captured stdout, and #733 puts that output into `/aws/lambda/<name>`. The envelope then goes unread. The metric never exists, and an alarm over it cannot be driven to a state change.

## Desired behaviour

A log event whose message parses as an EMF document publishes its metrics. Each entry in `_aws.CloudWatchMetrics` names a namespace, a list of dimension key sets and a list of metric names. The values come from the top-level properties of the same document, and the timestamp from `_aws.Timestamp`.

A log event that fails to parse as EMF is stored and left alone.

## Example

```typescript
// The handler bundles Powertools and counts on the failure path.
await simAws.region(region).account().lambda().invoke({
  input: { FunctionName: "user", Payload: badRequest },
});

const stats = await simAws.region(region).account().cloudWatch()
  .getMetricStatistics({
    input: {
      Namespace: "ChineseBoost",
      MetricName: "UserRequestFailed",
      Dimensions: [{ Name: "service", Value: "user" }],
      StartTime: from,
      EndTime: to,
      Period: 300,
      Statistics: ["Sum"],
    },
  });
```

## Proposed scope

Parsing on `PutLogEvents`, one datapoint per metric name per dimension set, with `_aws.Timestamp` and the per-metric `Unit`. Publication writes into the metric store the way `PutMetricData` would, because an EMF namespace is a custom one.

## Acceptance criteria

- [ ] A log event carrying an EMF document publishes one datapoint per metric name per dimension set in `_aws.CloudWatchMetrics`.
- [ ] Metric values and dimension values are read from the document's top-level properties. A metric naming a property the document lacks is skipped.
- [ ] `_aws.Timestamp` sets the datapoint's timestamp, read as milliseconds.
- [ ] A metric's `Unit` reaches the datapoint, and `GetMetricStatistics` filters on it.
- [ ] A plain text log event, or a JSON one lacking `_aws`, is stored and publishes no metric.
- [ ] A Powertools handler bound to a simulated function drives an alarm over its own metric to `ALARM`.
- [ ] Usage docs under `docs/services/logs/` saying EMF is read.

## Implementation notes

`SimLogsSubscriptionFanOut.written()` runs at `src/service/logs/command/event/sim-logs-put-log-events.ts:76`, and EMF parsing hangs off the same call.

`_aws.CloudWatchMetrics[].Dimensions` is a list of dimension key lists, and each inner list is one whole dimension set. Powertools emits a single set, and a document carrying several publishes the same value once per set.

A `SimLogs` built on its own has no CloudWatch to publish into, the way it has no subscription destination to deliver to.

## Out of scope

`StorageResolution: 1`, refused everywhere else in simulated CloudWatch. EMF documents arriving by any route other than a log event. Metric filters over a log group are #1167.

## Depends on

#606, #622 and #733, all merged.

hughgrigg:
A log event that is itself an EMF document publishes the metrics it declares, which is how AWS Lambda Powertools counts without calling any CloudWatch API. Each entry in `_aws.CloudWatchMetrics` names a namespace, its dimension sets and its metrics, and the values come from the top-level properties of the same document. A handler bundling Powertools therefore drives an alarm over its own metric through the log group its output already reaches.

A line that is not a readable document is stored and left alone, because a log group is full of lines that were never metrics. A metric the metadata declares and the body carries no number under, and one asking for a resolution shorter than a minute, go on the publication ledger instead. That ledger now covers both metric filters and embedded documents, so its entries name their source and it is read through `metricPublicationFailures`.

Resolves #1168

## Summary by CodeRabbit

* **New Features**
  * Added support for extracting and publishing Embedded Metric Format (EMF) metrics from log events.
  * Added handling for metric values, dimensions, timestamps, units, and publication failures.
  * Added an executable EMF simulation example.
  * Added access to metric publication failures and improved metric operation support.

* **Documentation**
  * Clarified metric-filter defaults, failure reporting, and EMF processing behavior.
  * Added guidance and examples for EMF log events.

* **Tests**
  * Added comprehensive EMF, CloudWatch alarm, validation, and failure-handling coverage.

### DIFF
--- a/docs/services/logs/README.md
+++ b/docs/services/logs/README.md
@@ -806,7 +806,7 @@ console.log(counted.Datapoints?.[0]?.Sum);
 
 Datapoints go in through the same `PutMetricData` an ordinary caller uses, into the CloudWatch of the log group's own Account and Region. A datapoint is stamped from the simulation's clock at the instant the event was written.
 
-A transformation carrying a `defaultValue` publishes that value once for a minute that took log events and matched none of them, which is how real CloudWatch Logs keeps a metric reporting through a quiet stretch. A minute a match landed in publishes the match alone. A transformation with no default value publishes nothing for a minute it matched nothing in, and the metric is then left with no datapoint at all over that stretch.
+A transformation carrying a `defaultValue` publishes that value once for a minute that took log events and matched none of them. That is how real CloudWatch Logs keeps a metric reporting through a quiet stretch. A minute a match landed in publishes the match alone. A transformation with no default value publishes nothing for a minute it matched nothing in, and the metric is then left with no datapoint at all over that stretch.
 
 A transformation carrying dimensions cannot also carry a `defaultValue`, and one carrying both is refused. Real CloudWatch Logs allows one or the other, because a default would have to be reported against every dimension value the filter has ever seen.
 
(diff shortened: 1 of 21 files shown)


hughgrigg:
@coderabbitai full review

PR #1177, opened 2026-08-30
feat: run PreAuthentication at managed login

### LINKED ISSUE #1173: Run PreAuthentication when a local user signs in at managed login
## Problem

A local user signing in at managed login runs no authentication trigger. The same user signing in through `InitiateAuth` runs both. An application that logs a sign-in or blocks one from a `PreAuthentication` handler passes its tests against the API sign-in and lets every managed login sign-in through untouched.

## Current behaviour

`SimCognitoPasswordSignIn` fires `PreAuthentication` and `SimCognitoSignInCompletion` fires `PostAuthentication` (`src/service/cognito/command/auth/sim-cognito-password-sign-in.ts:81`, `sim-cognito-sign-in-completion.ts:110`). Both belong to the API auth flows.

The hosted domain has a sign-in path of its own. `SimCognitoHostedPasswordSignIn.signIn` calls `requireSimCognitoSignInUser` and `requireSimCognitoSignIn` directly (`src/service/cognito/command/hosted/sim-cognito-hosted-password-sign-in.ts:33`), reaching no trigger. The passkey path and the managed login session path go the same way.

`PreTokenGeneration` does fire, from the token endpoint.

## Desired behaviour

AWS documents the sources a local user's managed login sign-in reports, in [Lambda triggers for Amazon Cognito local users in managed login](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-working-with-lambda-triggers.html). Against `/login` it names `PreAuthentication_Authentication` and `TokenGeneration_Authentication`, which is the source the token endpoint already reports.

A password sign-in and a passkey sign-in at the authorize endpoint both fire `PreAuthentication`. A handler that throws from it refuses the sign-in, and the serving layer draws that refusal on the sign-in form the way it draws a wrong password.

A browser signing in from the managed login session cookie it is already holding leaves the trigger unfired. The `PreAuthentication` docs say the trigger "doesn't activate" on renewal of an existing authentication session.

## Acceptance criteria

- [ ] A password sign-in at `/oauth2/authorize` fires `PreAuthentication` with `PreAuthentication_Authentication`
- [ ] A passkey sign-in at the same endpoint fires it too
- [ ] A handler that throws refuses the sign-in, and a served request comes back as the sign-in form carrying the message
- [ ] A browser signed in from its managed login session cookie leaves the trigger unfired
- [ ] The Cognito docs say which triggers a managed login sign-in runs, beside the API table already there

## Out of scope

`PostAuthentication`. The documented table names it for a federated sign-in and leaves it out for a local managed login one. Whether real Cognito fires it there was not checked against a live account, so this leaves it unfired and says so in the docs.

The federated sign-in path, which #1172 covers.

`TokenGeneration_AuthenticateDevice`, `UserMigration_Authentication` and the custom sender sources the same table names. Device tracking, user migration and the custom senders are all outside the simulation and refused when a pool names them.

hughgrigg:
A local user signing in at the hosted domain ran no authentication trigger, where the same user signing in through `InitiateAuth` ran both, so a handler that logs or blocks a sign-in passed its API tests and waved every managed login sign-in through. `PreAuthentication` now runs there under the `PreAuthentication_Authentication` source real Cognito reports from `/login`, for a password and for a passkey alike. It runs once the user is known and before the password is checked, which is where the API sign-ins run it, so a sign-in the pool is about to refuse reaches the handler too. A passkey runs it on the request that asks for the credential rather than on the one presenting it, matching the `USER_AUTH` sign-in, and only an asked request holds a session to present against, so it runs once either way. A browser coming back on the managed login session it already holds runs nothing, which is what the trigger's own docs say of a renewed session. A handler that throws is drawn on the sign-in form the way a wrong password is. `PostAuthentication` stays unfired and the docs say why.

One thing worth knowing about the test fixture. `simCognitoWithHostedPasskey` sets the pool's `SignInPolicy` with `UpdateUserPool`, and an update replaces a pool's settings whole, so the request was taking the pool's `LambdaConfig` back off it. That is real Cognito behaviour rather than a simulator bug, and the fixture now carries the triggers through the update with a comment saying why.

Resolves #1173

## Summary by CodeRabbit

* **New Features**
  * Hosted password and passkey sign-ins now invoke the configured **PreAuthentication** trigger during initial sign-in.
  * Trigger responses and errors are surfaced while preserving relevant form data.
  * Trigger context includes the sign-in source and username.
* **Bug Fixes**
  * Session renewals no longer re-run **PreAuthentication** triggers.
  * Documented managed-login behavior, including cases where sign-in triggers do not run.

### DIFF
--- a/docs/services/cognito/README.md
+++ b/docs/services/cognito/README.md
@@ -2752,6 +2752,18 @@ try {
 Cognito. Neither fires for `REFRESH_TOKEN_AUTH`, as neither does on real Cognito, where
 `PreTokenGeneration` does.
 
+A sign-in at the hosted domain runs `PreAuthentication` too, under the same
+`PreAuthentication_Authentication` source real Cognito reports it from `/login`. It runs for a
+password and for a passkey, once per sign-in, before the password is checked. A handler that throws
+refuses the sign-in, and a served domain draws that refusal on the sign-in form the way it draws a
+wrong password. A browser coming back on the managed login session it already holds runs nothing,
+because the `PreAuthentication` docs say the trigger does not activate on the renewal of a session
+that already exists.
+
+`PostAuthentication` stays unfired at the hosted domain. AWS names it for a federated sign-in and
+leaves it out of the table for a local user at managed login, and what real Cognito does there was
+not checked against a live account.
+
 ### Federated sign-in triggers
 
 A user arriving from an identity provider runs the pool's triggers too, and which ones it runs
@@ -4672,6 +4684,8 @@ Sim Cognito currently supports:
 - The triggers a federated sign-in runs, being `PreSignUp` and `PostConfirmation` on a first
   sign-in, `PreAuthentication` and `PostAuthentication` on every one after it, and
   `PreTokenGeneration` under `TokenGeneration_HostedAuth` when the code is exchanged
+- The `PreAuthentication` trigger a local user's managed login sign-in runs, for a password and for
+  a passkey alike, with a handler's refusal drawn on the sign-in form
 - A `REGIONAL` web ACL in front of the pool, attached by `AssociateWebACL` on simulated WAFv2 and
   evaluated against every request the hosted domain and the two `.well-known` documents answer
 - App client OAuth settings: `AllowedOAuthFlowsUserPoolClient`, `AllowedOAuthFlows`,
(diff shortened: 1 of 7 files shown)


hughgrigg:
@coderabbitai full review

PR #1238, opened 2026-09-02
feat: fill three CFN template language gaps

### LINKED ISSUE #1225: Fn::If in Conditions, Condition on an Output and FindInMap DefaultValue
## Problem

Three gaps in the template language, each small and each in the same layer of simulated CloudFormation.

`Fn::If` is rejected inside the `Conditions` section, where CloudFormation allows it. The `Condition` attribute is read on a resource but not on an output, so a conditional output is resolved and present in `stack.outputs` whichever way its condition falls, where CloudFormation leaves it out. And `Fn::FindInMap` accepts only the three-argument form, so a lookup carrying `{ "DefaultValue": ... }` as a fourth argument is rejected before the map is read.

## Current behaviour

`SimCfnConditionEvaluator` evaluates each named condition on demand through `SimCfnConditionExpression`, which handles the condition functions and refuses everything else.

`SimCfnStackOutputResolver` builds an entry for every output the template declares. It never sees `SimCfnConditions`, which `SimCfnResourceConditions` shows how to read.

`SimCfnFnFindInMapParser` requires an array of exactly three, and `SimCfnFnFindInMap.resolve` throws when the map, the top level key or the second level key is missing.

## Desired behaviour

`Fn::If` resolves inside the `Conditions` section, picking its branch from a condition already evaluated. A cycle it creates is caught by the chain the evaluator already carries.

An output whose `Condition` is false is left out of the stack's outputs and publishes no export. An output naming a condition the template does not define fails the deployment, matching how a resource naming an undefined condition is handled.

`Fn::FindInMap` accepts a fourth argument of `{ "DefaultValue": <value> }` and returns that value where the map, the top level key or the second level key is absent. Without the fourth argument a missing key still fails the resource.

## Acceptance criteria

- A condition defined with `Fn::If` evaluates, and the resources it gates deploy accordingly.
- A condition whose `Fn::If` closes a cycle fails with the existing self-reference error.
- An output whose condition is false is absent from `stack.outputs` and from `DescribeStacks`.
- An output whose condition is true resolves and exports as it does today.
- An output naming an undefined condition fails the deployment.
- `Fn::FindInMap` with a `DefaultValue` returns it for a missing map, top level key and second level key, and returns the mapped value where one exists.

## Out of scope

- `Fn::FindInMap` arguments that depend on a created resource, which CloudFormation rejects too.
- Restricting `Fn::Split` and `Fn::Select` to the functions CloudFormation allows inside them.
- `Fn::ForEach` and the language extensions transform.

## AWS references

- [Fn::FindInMap](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-findinmap.html)
- [Conditions section](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html)

hughgrigg:
`Fn::If` evaluates inside the `Conditions` section, picking a branch from a condition the evaluator has already answered. A cycle it closes fails with the self-reference error the evaluator already carried. An output whose `Condition` is false is left out of the stack, absent from `stack.outputs` and `DescribeStacks` and publishing no export, while an output naming a condition the template never defines fails the deployment the way a resource does. `Fn::FindInMap` takes a fourth `{ "DefaultValue": ... }` argument and answers a missing map, top level key or second level key with it. A lookup without one still fails.

Resolves #1225

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -328,3 +328,4 @@ test/terraform/**/.terraform.tfstate.lock.info
 *.tfplan
 
 .worktrees/
+.pnpm-store
--- a/.gitignore
+++ b/.gitignore
@@ -326,3 +326,5 @@ test/terraform/**/builds/
 test/terraform/**/terraform.tfstate*
 test/terraform/**/.terraform.tfstate.lock.info
 *.tfplan
+
+.worktrees/
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -771,6 +771,23 @@ The value a lookup returns can be any type. A list value is returned as a list.
 used in resource properties and in `Outputs`. A map name or key missing from `Mappings` fails the
 deployment with an error naming the path that could not be found.
 
+A fourth argument of `{ "DefaultValue": ... }` gives the lookup an answer for that case.
+
+```json
+{
+  "Fn::FindInMap": [
+    "EnvironmentMap",
+    { "Ref": "Environment" },
+    "BucketName",
+    { "DefaultValue": "fallback-site-bucket" }
+  ]
+}
+```
+
+The default answers a map name, a top-level key or a second-level key the `Mappings` section does
+not carry. It is a template value like any other, so it can itself be a `Ref` or a nested
+expression. A lookup that finds its value in the map ignores the default.
+
 ### `Fn::Split` and `Fn::Select`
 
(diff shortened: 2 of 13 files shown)


hughgrigg:
@coderabbitai rate limit

PR #1241, opened 2026-09-02
feat: apply Function URL CORS configuration

### LINKED ISSUE #1217: Apply Function URL CORS configuration to served responses
## Problem

Yulin accepts Function URL resources but ignores their CORS configuration. The simulated Lambda API does not retain or return `Cors`, the `AWS::Lambda::Url` creator reads only `AuthType` and `InvokeMode`, and the HTTP controller invokes the function for every request. A template can therefore deploy successfully while its browser preflight and response headers differ from AWS.

Issue #174 added Function URLs and named CORS handling as follow-up work. The current limitation now creates a false pass because valid `Cors` properties are silently discarded.

## Desired behaviour

Store the complete Function URL CORS configuration and expose it through the create, get, update, and list commands. `AWS::Lambda::Url.Cors` should use the same path. When CORS is configured, the Function URL HTTP controller should add the configured headers to normal responses and answer preflight requests without invoking the function.

AWS gives configured headers precedence for preflight requests. For other methods it returns both configured headers and CORS headers supplied by the function, including duplicates where they conflict.

## Acceptance criteria

- Create, get, update, and list Function URL configuration commands round-trip every supported `Cors` field.
- `AWS::Lambda::Url.Cors` deploys the same configuration as the Lambda API.
- A valid preflight request returns only the configured CORS headers and does not invoke the handler.
- A non-preflight response contains the configured CORS headers as well as any CORS headers returned by the handler.
- Invalid CORS values are rejected with Lambda-compatible validation errors before the URL configuration changes.
- Isolated API, CloudFormation, and HTTP tests cover allowed origins, methods, headers, credentials, exposed headers, and maximum age.

## Out of scope

- `RESPONSE_STREAM` response streaming.
- Function URLs for aliases or published versions.
- API Gateway CORS configuration.

## AWS references

- [Creating and managing Lambda function URLs](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html)
- [Lambda CORS API type](https://docs.aws.amazon.com/lambda/latest/api/API_Cors.html)

hughgrigg:
Yulin accepted a Function URL's `Cors` block and discarded it. A template could deploy clean while its browser preflight and response headers differed from AWS. The configuration is now held on the Function URL and round-trips through the create, get, update and list commands. `AWS::Lambda::Url.Cors` and SAM's `FunctionUrlConfig` deploy it by that same path. The HTTP controller answers a valid preflight from the configuration without invoking the handler, and adds the configured headers to every other response alongside the ones the handler sent, duplicates included. Values outside the bounds the Lambda API publishes are refused before the URL configuration changes.

Resolves #1217

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -328,3 +328,4 @@ test/terraform/**/.terraform.tfstate.lock.info
 *.tfplan
 
 .worktrees/
+.pnpm-store
--- a/.gitignore
+++ b/.gitignore
@@ -326,3 +326,5 @@ test/terraform/**/builds/
 test/terraform/**/terraform.tfstate*
 test/terraform/**/.terraform.tfstate.lock.info
 *.tfplan
+
+.worktrees/
--- a/docs/services/cloudformation/README.md
+++ b/docs/services/cloudformation/README.md
@@ -2507,8 +2507,7 @@ time window gets `OFF`.
 once, and a failed one is recorded by the rule or the schedule that made it.
 
 A `FunctionUrlConfig` on the function expands into an `AWS::Lambda::Url` named after it, `RatesUrl`
-for a function called `Rates`. `AuthType` and `InvokeMode` carry over. `Cors` is left out (the
-simulated Function URL answers no preflight request).
+for a function called `Rates`. `AuthType`, `InvokeMode` and `Cors` carry over.
 
 ### Queue, stream, topic and bucket events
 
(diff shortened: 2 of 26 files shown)


hughgrigg:
@coderabbitai rate limit

PR #1243, opened 2026-09-02
feat: honor stream mapping retry limits

### LINKED ISSUE #1219: Honor finite retry and record-age limits on stream event source mappings
## Problem

Yulin refuses `MaximumRetryAttempts` and `MaximumRecordAgeInSeconds` on Lambda event source mappings. DynamoDB Streams and Kinesis mappings instead retry a failed batch five times after the first attempt, with fixed delays of 1, 2, 4, 8, and 16 simulated seconds. Applications that set a finite retry quota or discard old records cannot test either boundary.

Issues #341 and #909 introduced the stream mappings and deliberately deferred these controls. Both settings govern the same failed-batch lifecycle and can share the existing stream retry state without taking on the other deferred event source features.

## Desired behaviour

Accept finite values for DynamoDB Streams and Kinesis mappings, retain them on the mapping, and expose them through create, get, and list responses. A failed record should stop retrying when it reaches either its retry quota or its maximum age. Discarding expired or exhausted records must advance the shard checkpoint so later records can be processed.

Keep Yulin's positive retry delay so tests can advance through attempts explicitly. When both settings are omitted or `-1`, continue to use the documented five-retry safety cap.

## Acceptance criteria

- Create and CloudFormation deployment accept valid finite values and reject values outside Lambda's ranges.
- Get and list responses report the configured values for DynamoDB Streams and Kinesis mappings.
- `MaximumRetryAttempts: 0` makes one invocation attempt, and a positive value allows exactly that many retries.
- Before each retry, records older than `MaximumRecordAgeInSeconds` are discarded while unexpired records remain eligible.
- Discarding all exhausted or expired records advances the checkpoint and allows later shard records to run.
- Isolated tests cover both stream services, partial batch failures, the retry limit, the age limit, and the documented default safety cap.

## Out of scope

- `DestinationConfig`, `BisectBatchOnFunctionError`, non-zero batching windows, and parallelization.
- `UpdateEventSourceMapping` support.
- Replacing Yulin's deterministic retry delays with AWS's internal polling cadence.

## AWS references

- [Lambda parameters for Kinesis event source mappings](https://docs.aws.amazon.com/lambda/latest/dg/services-kinesis-parameters.html)
- [CreateEventSourceMapping API](https://docs.aws.amazon.com/lambda/latest/api/API_CreateEventSourceMapping.html)

hughgrigg:
DynamoDB Streams and Kinesis event source mappings now take `MaximumRetryAttempts` and `MaximumRecordAgeInSeconds`, hold them, and report them through create, get and list. A failed batch stops being delivered once it has had its retries or once the next attempt would fall past the record age, and what is left is discarded, moving the checkpoint on so later records run. Records age out of the front of a batch, and the ones behind an expired record go over again on their own. Values outside Lambda's documented ranges are refused as validation errors, a queue mapping is refused for naming either, and a mapping naming neither keeps the documented five-retry cap.

Resolves #1219

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring maximum retry attempts and record age for DynamoDB and Kinesis stream event-source mappings.
  * Stream mappings now apply retry limits, exponential backoff, and record-age expiration during failed-batch processing.
  * Retry settings are preserved when mappings are retrieved or listed.
  * Queue event-source mappings reject stream-only retry settings.

* **Documentation**
  * Updated Lambda retry behavior documentation and added a runnable example.

* **Bug Fixes**
  * Improved stream record timing and partial-batch retry handling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -328,3 +328,4 @@ test/terraform/**/.terraform.tfstate.lock.info
 *.tfplan
 
 .worktrees/
+.pnpm-store
--- a/.gitignore
+++ b/.gitignore
@@ -326,3 +326,5 @@ test/terraform/**/builds/
 test/terraform/**/terraform.tfstate*
 test/terraform/**/.terraform.tfstate.lock.info
 *.tfplan
+
+.worktrees/
--- a/docs/services/lambda/README.md
+++ b/docs/services/lambda/README.md
@@ -1755,15 +1755,152 @@ Advancing the simulation's clock is what hands the batch over again:
 await simAws.clock().advanceBy({ seconds: 30 });
 ```
 
-The batch is delivered again five times, after 1, 2, 4, 8 and 16 seconds, so six deliveries in all.
-Then it is discarded and the mapping carries on with the stream, as AWS does once a stream mapping's
-error handling has run out.
+A mapping that asks for no limit of its own delivers the batch again five times, after 1, 2, 4, 8
+and 16 seconds, so six deliveries in all. Then it is discarded and the mapping carries on with the
+stream, as AWS does once a stream mapping's error handling has run out.
 
-Both the cadence and the number of attempts are simulator constraints rather than AWS behaviour. AWS
-documents no delay between attempts and retries until the records age out of the stream, a day
-later. A delay of zero here would fall due at the instant the clock already reads. A handler that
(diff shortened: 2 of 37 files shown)


hughgrigg:
@coderabbitai full review

